PR #5, opened 2017-05-26
Feature/tbl 040 ensure logging writes to stdout

liamtoozer:
Basic change to ensure existing logs (previously 'print' statements) are sent to stdout for Cloud Foundry (CF):

- Replaced print statements with logger calls in app.py (all set to 'debug' level logging).
- Added simple configuration for logger in config.py.

### How to review 
_Note: This test simply ensures log information is output to stdout and has no dependencies such as an OAuth2 server running etc._

Push the changes to Cloud Foundry (this will create a new app):
`cf push <create-app-name> -b https://github.com/cloudfoundry/python-buildpack.git`

Tail the logs from CF:
`cf logs frontstage-lt-2605 | grep -E 'APP.*OUT'`

Go to an end point in a browser (e.g. `http://<app-name>.cfapps.io/sign-in/`) and ensure stdout log information is output from CF

### DIFF
--- a/rasfrontstage/app.py
+++ b/rasfrontstage/app.py
@@ -4,6 +4,8 @@ Main file that is ran
 #from __future__ import print_function
 from functools import wraps, update_wrapper
 from datetime import datetime
+import logging
+import sys
 from flask import Flask, make_response, render_template, request
 import os
 import requests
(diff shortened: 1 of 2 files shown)


ras-frontstage/app.py
```diff
@@ -67,13 +69,12 @@ def logged_in():
...
-                print " {} is: {}".format(key, decodedJWT[key])
+                app.logger.debug(" {} is: {}".format(key, decodedJWT[key]))
                 #userID = decodedJWT['user_id']
             return render_template('signed-in.html', _theme='default', data={"error": {"type": "success"}})
 
         except JWTError:
-            #TODO Provide proper logging
-            print "This is not a valid JWT Token"
+            app.logger.debug("This is not a valid JWT Token")
             #app.logger.warning('JWT scope could not be validated.')
```
rjweeks: Delete unwanted code?

PR #9, opened 2017-05-31
Code issue fixes

gemmai95:


app/application.py
```diff
@@ -421,30 +421,23 @@ def register_enter_your_details():
...
+                                       "username": email_address}
 
             # We need to take our token from teh OAuth2 server and encode in a JWT token and send in the authorization
             # header to the party service microservice
             encoded_jwt_token = encode(data_dict_for_jwt_token)
 
         except JWTError:
-            #TODO Provide proper logging
+            # TODO Provide proper logging
             print("This is not a valid JWT Token")
```
nherriot: Can you replace this with a debug.warning please ? :-)

app/application.py
```diff
@@ -363,15 +373,6 @@ def register_enter_your_details():
             # TODO A redirect to a page that helps the user
 
         except requests.exceptions.Timeout:
```
nherriot: can you change the logger.debug to a logger.warning or logger.error please. :-)

app/application.py
```diff
@@ -363,15 +373,6 @@ def register_enter_your_details():
...
-        except requests.exceptions.RequestException as e:
-            # TODO catastrophic error. bail. A page that tells the user something horrid has happeded and who to inform
-            print(e)
-            app.logger.debug("There seems to be no server listening on this connection?")
-            # TODO A redirect to a page that helps the user
-
-        except requests.exceptions.Timeout:
             app.logger.debug("Timeout error. Is the OAuth Server overloaded?")
             # TODO A redirect to a page that helps the user
         except requests.exceptions.RequestException as e:
```
nherriot: After the except is caught we don't do anything. Not even a 'print' ! Can you put at least a #TODO or/and logger.warning/logger.error message. :-) Thank you.

app/application.py
```diff
@@ -547,7 +539,7 @@ def get_id(_id):
                 return res
 
             # Anything else but a token match means we reject the call
-                app.logger.debug("tokens don't match")
+            app.logger.debug("tokens don't match")
```
nherriot: might was well raise this as an info or warning as we might want to trace when a token does not match. this is due to us rejecting a call - so that means someone might be 'miffed off'! hence let's raise this up a little (i think debug will get ignored on live systems)

test-scripts/resourceOwnerPasswordGrant.py
```diff
@@ -53,16 +52,15 @@ class OAuthConfig(Config):
...
 
-
-client_id='onc@onc.gov'
+client_id = 'onc@onc.gov'
 client_secret = 'password'
 username = 'testuser@email.com'
 password = 'password'
 
-print " *** Variables setup ***"
+print(" *** Variables setup ***")
```
nherriot: Can we make this a log? I only mention this because there was an update to make it python3 specific. :-)

test-scripts/resourceOwnerPasswordGrant.py
```diff
@@ -71,11 +69,11 @@ class OAuthConfig(Config):
 client.prepare_request_body(username=username, password=password, scope=['ci.write', 'ci.read'])
 oauth = OAuth2Session(client=client)
 
-print " OAuth2 Session has been created for client: {}".format(client_id)
+print(" OAuth2 Session has been created for client: {}".format(client_id))
```
nherriot: Can we make this a logger.debug statement ? :-) Thank you.

test-scripts/resourceOwnerPasswordGrant.py
```diff
@@ -71,11 +69,11 @@ class OAuthConfig(Config):
...
 
 token = oauth.fetch_token(token_url='http://localhost:8000/api/v1/tokens/', username=username, password=password, client_id=client_id, client_secret=client_secret)
 
-print " *** Access Token Granted *** "
-print " Values are: "
+print(" *** Access Token Granted *** ")
```
nherriot: Same for all the 'print' lines here to. can we make them all logger statements?
thank you in advance.

PR #14, opened 2017-06-07
US026 My Surveys

duncannixon:


app/templates/partials/header-non-sim.html
```diff
@@ -3,12 +3,20 @@
         <div class="logo header__logo">
             <img src="{{ url_for('static', filename='images/logo.svg') }}" alt="Office for National Statistics" class="logo__img" />
         </div>
+        {% if logged_in %}
+        <p>Hello Nick</p>
```
nherriot: This has to be removed! As not everyone is me!!!!! :-)

app/templates/signed-in.html
```diff
@@ -9,6 +9,9 @@
         <h1 class="saturn">Congratulations. You are signed in.</h1>
         <p>You can now access protected endpoints.</p>
         <p>To sign out please <a href="{{ url_for('logout') }}" > select here.</a> </p>
+        {% if data.logged_in == "True" %}
```
nherriot: Should we not use the other way of doing this by looking at the session variable? i.e. it has the jwt token ? :-) this might be a better way - as in the future we can make this a template tag bla bla bla bla..... You know what I mean.
Anyway let me know what you think on that.

app/utils.py
```diff
@@ -1,7 +1,7 @@
 """Utils module."""
 
 from __future__ import print_function
-from app.models import *
+# from models import *
```
nherriot: You can delete this if you like. :-)

nherriot:
Can you check out the jinja2 login parameter ? Is that what you want? might be better to use the session.jwttoken ? :-) :-) 

cool work!

app/utils.py
duncannixon: Removed.

app/templates/signed-in.html
duncannixon: I removed this clause completely and renamed the template to not-signed-in.html as we only now visit this template if we try to access a page before we have signed in.

app/templates/partials/header-non-sim.html
duncannixon: Removed!

PR #18, opened 2017-06-09
Fixed lint errors. Fixed some enrolment urls. Refactored My Surveys p…

duncannixon:
…ages to use new Api gateway endpoint instead of individual microservices.


app/application.py
```diff
@@ -41,7 +40,7 @@
 else:
     logger.info(" *** APP.Info Testing server settings are being used. ***")
     app.config.from_object(TestingConfig)
-    print ("testing server started...")
+    print("testing server started...")
```
nherriot: Can we make this a log call now? Maybe logging.info("testing serve started...")
:-)

app/application.py
```diff
@@ -52,6 +51,24 @@ def hello_world():
     return render_template('_temp.html', _theme='default')
 
 
+def build_survey_data(statusFilter):
+
+    # TODO - Derive the Party Id
```
nherriot: A very picky point. A link to the confluence page would be nice to let the user know where the current details are for this.

app/application.py
```diff
@@ -52,6 +51,24 @@ def hello_world():
...
+    party_id = "3b136c4b-7a14-4904-9e01-13364dd7b972"
+
+    # TODO - Add security headers
+    # headers = {'authorization': jwttoken}
+    headers = {}
+
+    # Call the API Gateway Service to get the To Do survey list
```
nherriot: Put a #TODO for the 'todo' in the URL. That has to come out, but we might forget about it. If you put a #TODO it will always be picked up by the lint checker and/or pycharm! :-) ....
Or you could stick this as a variable in the config file, which will then get changed when that piece of work gets complete.... Yesh-of-coursh!

app/application.py
```diff
@@ -52,6 +51,24 @@ def hello_world():
...
+    url = Config.API_GATEWAY_SURVEYS_URL + 'todo/' + party_id
+    req = requests.get(url, headers=headers, params=statusFilter, verify=False)
```
nherriot: You are making a call to requests.get which depends on an external application. So we should wrap this up in an except clause. It sh*t can happen it will! So we can protect against that by catching the exception and presenting a page with maybe "Currently have issues getting your cases please contact Bill Gates!" or something like that.

Check out the method 'fetch.token' which does a request.get under the hood. We catch all exceptions and display some sense to the user. It's not perfect, but better than nothing! :-)

app/application.py
```diff
@@ -72,19 +89,12 @@ def logged_in():
...
-            # Build the survey data (To Do survey type)
-            dataArray = build_survey_data()
-
-            # Filter the data array to remove surveys that shouldn't appear on the To Do page
-            allowedStatuses = ['Not started', 'In progress']
-
-            # TODO - the line below can be commented out to demonstrate sorting. This comment can be removed eventually.
-            dataArray = filter_surveys(dataArray, allowedStatuses)
+            # Filters the data array to remove surveys that shouldn't appear on the To Do page
+            statusFilter = {'status_filter': '["Not started", "In progress"]'}
```
nherriot: You have a list inside a string????? I'm not sure why you are doing that? Try this, in your terminal type python then paste this code into your >>>.
>>> mydict = {'status':['not started','started','maybe']}
>>> mydict['status']
['not started', 'started', 'maybe']
>>> for key in mydict['status']:
...     print key
... 
not started
started
maybe
>>>

You pass the string in the header, so maybe the other side de-seriealises it and jobs a good'n but from a perfect python sense it's a bit strange! Are you happy with this implementation ?

app/application.py
```diff
@@ -72,19 +89,12 @@ def logged_in():
...
 
-            # Sort the data array so that the closed Submit by dates appear at the top of the list
-            dataArray = sort_survey_data(dataArray)
+            # Get the survey data (To Do survey type)
+            dataArray = build_survey_data(statusFilter)
```
nherriot: Can the build_survey return any type of error? if it can then it needs to be captured and delt with.

app/application.py
```diff
@@ -671,9 +561,15 @@ def register_enter_your_details():
...
+        registrationData = {'emailAddress': email_address,
+                            'firstName': first_name,
+                            'lastName': last_name,
+                            'telephone': phone_number,
+                            'status': 'CREATED'}
+
         headers = {'authorization': encoded_jwt_token, 'content-type': 'application/json'}
-        partyServiceURL = PartyService.PARTYSERVICE_PROTOCOL + PartyService.PARTYSERVICE_SERVER + PartyService.PARTYSERVICE_RESPONDENTS_ENDPOINT
+
+        partyServiceURL = Config.API_GATEWAY_PARTY_URL + 'respondents'
```
nherriot: This is much more clear which is great! Good job. ;-) ...
But if you make the creation of the URL with things like protocol, server and endpoints it means you can change things more granularly in the future. Which might be good when you have multiple endpoints e.g. the OAuth2 server now has 4 endpoints and can support http or https if we put an TLS layer in front of the server.
With great power comes great responsibility !!! If there are lots of changes required to party service URL you can maintain this! :-)

app/application.py
duncannixon: Agreed - I have changed this.

PR #28, opened 2017-06-21
Feature/sm029 save draft updates

MIslam92:
Fixed bugs below:

- Characters  after space in subject field is persisted but not displayed
- Remove bullet point displayed after saving draft
- Make link work after saving draft


MIslam92:
- fix to PUT requests on drafts in HTML forms

PR #45, opened 2017-07-04
Feature/sm support message to and from details

gemmai95:


app/templates/secure-messages.html
```diff
@@ -85,9 +85,9 @@
             </dt>
             <dd class="grid__col col-9@m pull-1@m">
                 {% if 'UNREAD' in message['labels'] %}
-                    <b>{{ message['urn_from'] }}</b>
+                    <b>{{ message['@msg_from']['email'] }}</b>
```
rjweeks: do we need to check for None?

app/templates/secure-messages.html
```diff
@@ -85,9 +85,9 @@
...
                 {% else %}
-                    {{ message['urn_from'] }}
+                    {{ message['@msg_from']['email'] }}
```
rjweeks: Do we need checks for this? Check what happens if a key doesn't exist.

gemmai95: If a message key doesn't exist or is none it will leave this blank

PR #48, opened 2017-07-06
Sm067 save draft reply

dcdarrell9:
### Context of this PR
- [Trello card](https://trello.com/c/d6gHlGxG)
- Adding save draft functionality to draft page

### How to review
- Check all functionality works correct
- Some duplicate code(maybe suggest way to cut code down/re-use)

### DIFF
--- a/app/views/secure_messaging.py
+++ b/app/views/secure_messaging.py
@@ -81,7 +81,7 @@ def reply_message():
 
     # if session.get('jwt_token'):
     if request.method == 'POST':
-        if request.form['submit'] == 'Reply to message':
+        if request.form['submit'] == 'Send message':
             logger.info("Reply to Message")
             data = {'msg_to': 'BRES',
                     'msg_from': '0a7ad740-10d5-4ecb-b7ca-3c0384afb882',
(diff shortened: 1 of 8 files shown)


app/views/secure_messaging.py
```diff
@@ -73,18 +81,57 @@ def reply_message():
...
+                response = requests.post(SecureMessaging.CREATE_MESSAGE_API_URL, data=json.dumps(data), headers=headers)
+                if response.status_code != 201:
+                    # TODO replace with custom error page when available
+                    return redirect(url_for('error_page'))
+                resp_data = json.loads(response.text)
+                logger.debug(resp_data.get('msg_id', 'No response data.'))
+                return render_template('message-success-temp.html', _theme='default')
+
+            if request.form['submit'] == 'Save draft':
+                data = {'msg_to': 'BRES',
```
rjweeks: data assignment can be done before if, so there is no duplicate code.

dcdarrell9: @rjweeks The other data (in reply) has thread_id, is that used for anything?

dcdarrell9:
[Save draft reply ( See Ras Beta) |5|](https://trello.com/c/bv5CMbD6/370-save-draft-reply-see-ras-beta-5)

PR #50, opened 2017-07-11
Feature/us003 implement password complexity

liamtoozer:
To add a password 'minimum requirement' check. This change is for back-end validation as there is already front-end validation in place.

### How to review 
Start the server and navigate to: http://localhost:5001/create-account/enter-account-details/, enter an invalid password (one that does not meet the minimum requirements) and click the **Continue** button. This will kick off the back-end validation and report back the error to the user.


duncannixon:
Pretty horrendous merge issues, managed to get it merged and working OK I think...

app/models.py
```diff
@@ -105,6 +111,15 @@ def validate_phone_number(form, field):
             logger.debug(" There is a number parse exception in the phonenumber field")
             raise ValidationError('This should be a valid UK number e.g. 01632 496 0018. ')
 
+    def validate_password(form, field):
+        if len(field.data) > Config.PASSWORD_MAX_LENGTH:
```
liamtoozer: Thanks for the feedback. This is similar to what's in place in the **validate_phone_number** function as well. I think this in place as a 'catch-all' just in case someone makes a request to the service via cURL, as the field validation would be bypassed. I do need to implement a check for length < PASSWORD_MAX_LENGTH now though as I seemed to have forgotten to do this here!

app/models.py
```diff
@@ -81,9 +82,14 @@ class RegistrationForm(FlaskForm):
...
-    password = PasswordField('Create a password', validators=[DataRequired(), Length(min=8, max=254, message='Your password must be more than 8 characters')])
-    password_confirm = PasswordField('Re-type your password', validators=[DataRequired(), EqualTo('password', message='Passwords must match'),
-                                     Length(min=8, max=254, message='Your password must be more than 8 characters')])
+    password = PasswordField('Create a password', validators=[DataRequired(message=Config.PASSWORD_CRITERIA_ERROR_TEXT),
+                                                              EqualTo('password_confirm', message=Config.PASSWORD_MATCH_ERROR_TEXT),
+                                                              Length(min=Config.PASSWORD_MIN_LENGTH,
+                                                                     max=Config.PASSWORD_MAX_LENGTH,
+                                                                     message=Config.PASSWORD_CRITERIA_ERROR_TEXT)])
+    password_confirm = PasswordField('Re-type your password', validators=[Length(min=Config.PASSWORD_MIN_LENGTH,
```
liamtoozer: Agreed, this should be removed. Thanks

PR #51, opened 2017-07-11
Feature/us022 enter enrolment code unit tests

duncannixon:


oddjobz:
Passed local tests and runs.

PR #52, opened 2017-07-12
Feature/us008 verify new account

duncannixon:
Frontstage pages corresponding to /activate-account?t={token} and associated pages.


app/application.py
```diff
@@ -590,13 +590,54 @@ def register_almost_done():
...
+
+    if result.status_code == 200:
+        json_response = json.loads(result.text)
+
+        if json_response['active']:
+            # Successful account activation therefore redirect off to the login screen
+            return redirect(url_for('login', account_activated=True))
+        else:
+            # Try to get the user id
+            user_id = json_response['userId']
```
andrewmil: json_response.get('userId') is better here to avoid KeyErrrors

app/application.py
```diff
@@ -590,13 +590,54 @@ def register_almost_done():
...
+
+    # Call the Party service to try to activate the account corresponding to the token that was supplied
+    url = Config.API_GATEWAY_PARTY_URL + 'emailverification/' + token
+    result = requests.post(url)
+    logger.debug('Activate account - response from party service is: {}'.format(result.content))
+
+    if result.status_code == 200:
+        json_response = json.loads(result.text)
+
+        if json_response['active']:
```
andrewmil: same as below

app/templates/sign-in.html
```diff
@@ -56,11 +59,12 @@ <h1 class="saturn">Sign in</h1>
...
 
-                    <a href="/forgot-password">Forgot password</a>
+                    <br />
+                    <a href="/forgot-password">Forgot password?</a>
                 </div>
                 <br />
 
-                <button class="btn btn--primary venus" type="submit">Sign in</button>
+                <button class="btn btn--primary venus" type="submit" id="SIGN_IN_BUTTON">Sign in</button>
             {% if errorType == "failed" %}
```
andrewmil: separate issue from this pull request but should sign-in trigger this error stuff when it's submitted empty as well?

duncannixon: The RAD team have confirmed that the sign in screen should indeed have empty field validation - this will be followed up on a separate ticket.

PR #53, opened 2017-07-14
Fixed unit tests. Changed the run_linting script so that you don’t need…

duncannixon:
… a virtual env to run it. Fixed frontstage app so that it runs when you override the oauth settings in the config file. Fixed duplicate page title bug in html templates.


oddjobz:
Ok, I'm seeing the one test failure as expected.

PR #55, opened 2017-07-17
Feature/us002 respondent email address validation

andrewmil:


app/models.py
```diff
@@ -111,14 +116,13 @@ def validate_phone_number(form, field):
             logger.debug(" There is a number parse exception in the phonenumber field")
             raise ValidationError('This should be a valid UK number e.g. 01632 496 0018. ')
 
-    def validate_password(form, field):
-        if len(field.data) > Config.PASSWORD_MAX_LENGTH:
-            raise ValidationError(Config.PASSWORD_CRITERIA_ERROR_TEXT)
+    @staticmethod
+    def validate_password(field):
```
liamtoozer: Thanks for putting the changes in here!

app/templates/register.enter-your-details.html
```diff
@@ -5,21 +5,27 @@
...
+        <div class="panel panel--error">
+            <div class="panel__header">
+                {% if errors|length == 1 %}
+                <h1 class="panel__title venus">{{ errors[error][0] }}</h1>
+                {% elif errors|length > 1 %}
+                <h1 class="panel__title venus">There are {{ errors|length }} errors on this page</h1>
+                {% endif %}
+            </div>
+            <div class="panel__body" data-qa="error-body">
+                {% for error in errors %}
```
liamtoozer: This will mean `password_confirm` will be shown as a field name if an error occurs. Is this okay with for UX?

andrewmil: I actually don't have any validation for password confirm since it'll all be done by the equalTo in the password field so we shouldn't ever get password_confirm here.

app/models.py
andrewmil: 👍

app/templates/register.enter-your-details.html
liamtoozer: All okay for me now I've pulled from the merge in a more recent commit. Thanks

andrewmil:
The travis build was failing because of some postgreSQL functionality, is this actually being used anywhere in the frontstage?

PR #58, opened 2017-07-20
Feature/secure messaging jwt

andrewmil:
Requires changes in ras-common jwt_decorator which are currently in branch`integration_baseline_1`

Updated secure_messaging.py to be in line with the rest of the code and make use of ras-common for jwt validation


andrewmil:
The lack of testing on secure_message.py has been noted. New story covering creation of these tests will be made

PR #60, opened 2017-07-21
Feature/us012 request forgotten password

duncannixon:
Fixes a sign-in problem on master. Also includes some initial screen work on forgot / reset password  - doesn't break anything so safe to merge to master. Also includes changes to the dev-only home page to aid handover and demos.

### DIFF
--- a/app/templates/passwords/reset-password.html
+++ b/app/templates/passwords/reset-password.html
@@ -7,7 +7,7 @@
 
 {% block main %}
 
-{% if errorType != None %}
+{% if errorType %}
     {% set errorNoMatchTitle = "Your passwords do not match" %}
     {% set errorNoMatchDescription = "Please check the passwords and try again" %}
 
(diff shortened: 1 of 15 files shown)


app/templates/passwords/reset-password.html
```diff
@@ -50,24 +55,26 @@ <h1 class="saturn">Reset your password</h1>
     </ul>
 
     <fieldset id="reset-details" class="js-new-password-group">
-        {% if errorType != None %}
+
+        {% if errorType.password %}
```
andrewmil: The password field is displaying an error with javascript code when the password complexity is not met. There is an identical problem on the register-account-details page so could fix there too

app/views/passwords.py
```diff
@@ -1,45 +1,72 @@
...
+        # TODO do some kind of back end processing to validate the email address, error handling, logging
+        # print('Email address=' + email_address)
+
+        return redirect(url_for('passwords_bp.forgot_password_check_email'))
+
     template_data = {
         "error": {
-            "type": request.args.get("error")
+            "type": form.errors,
+            "logged_in": "False"
```
andrewmil: is logged_in used anywhere for this?

duncannixon: Thanks - logged_in is not used so I have taken it out.

app/templates/passwords/reset-password.html
duncannixon: Agreed, suggest confirming the requirements with the RAD team and either removing the javascript as a defect or meet their defined requirement as an iteration.

PR #69, opened 2017-07-27
Cut-down case service code only

oddjobz:
Ok, this is a cut-down PR designed to integrate case-service event logging only. However (!) in order for this to work, it *needs* to have consistent data, to this end the patch includes inserting a party_id in the JWT on login.

This code contains the 'beginnings' of the static routing PR, but is limited to changes in **config.py** and the new **post_event.py** file.

#### TODO

- The party_id inserted into the JWT is a hard-coded value from the agreed cross-site test data and said id designed to facilitate end-to-end testing of login/upload/download. This party_id *should* come from a party service lookup by email address - when the party service provides that facility.


rjweeks:
You need to merge master

PR #76, opened 2017-07-28
Replacing all gateway calls with P2P microservice calls

oddjobz:
- This includes the remaining changes to replace calls via the gateway with direct point to point calls.
- Also included is logging for failed downloads using the new category supplied by JT.
- All endpoints are defined in config.py with environment overrides
- All URI's are now also defined in config.py so changes resulting from API changes only need to happen in one place.
- see **scripts/run.py** for a sample set of environment variables to run locally

#### Background
The standard format for calling an endpoint is;
- Config.<endpoint_name>.format(<service>, [{parameters}])

For example;
- url = Config.RM_IAC_GET.format(Config.RM_IAC_SERVICE, enrolment_code)

Where;
- RM_IAC_GET is '{}iacs/{}'
- RM_IAC_SERVICE is '{}://{}:{}/'.format(...)

The paths of some if not many of the endpoints currently in use are likely to change, this pattern means all potentially changeable configuration is in one place and to re-jig the configuration you only need to edit one file.


app/views/surveys.py
```diff
@@ -25,7 +25,9 @@ def build_survey_data(session, status_filter):
     headers = {}
 
     # Call the API Gateway Service to get the To Do survey list
-    url = Config.API_GATEWAY_AGGREGATED_SURVEYS_URL + 'todo/' + party_id
+    # url = Config.API_GATEWAY_AGGREGATED_SURVEYS_URL + 'todo/' + party_id
+
+    url = Config.RAS_AGGREGATOR_TODO.format(Config.RAS_API_GATEWAY_SERVICE, session.get('party_id'))
```
andrewmil: There's still a call to the gateway service here, should that be changed too?

oddjobz: No, this is calling "/todo", which currently lives in the gateway ...

oddjobz: Running through the mocks now ...

oddjobz: (unit tests should now be passing ...)

PR #87, opened 2017-08-02
removing hardcoded ru_id in create_message

andrewmil:
- Added call to party service to get ru_id for create_message
- changed user_uuid in sign_in to match current testuser@email.com in oauth
- Removed unused selenium scripts

Note that currently testuser@email.com does not have any associations so creating a message will fail with this user as frontstage will not be able to find a ru_id for it. 

For MVP we assume that all users will have an association and only one survey "BRES" for each


andrewmil:
I'm realising now that this change will make it tricky for devs to create a message for without a user who is associated with some organisation. Should talk to Mark potentially to see if this exists/could be created

frontstage/views/sign_in.py
```diff
@@ -102,7 +102,7 @@ def login():
             "scope": token['scope'],
             "expires_at": token['expires_at'],
             "username": username,
-            "user_uuid": "ce12b958-2a5f-44f4-a6da-861e59070a32",
+            "user_uuid": "db036fd7-ce17-40c2-a8fc-932e7c228397",
```
dcdarrell9: What's happening with this id going forward? Wasn't it going to be removed?

frontstage/views/secure_messaging.py
```diff
@@ -42,13 +44,25 @@
...
+    party_response = requests.get(url)
+    if party_response.status_code != 200:
+        return redirect(url_for('error_bp.error_page'))
+    party_response_json = party_response.json()
+    associations = party_response_json.get('associations')
+    if associations:
+        ru_id = associations[0].get('partyId')
+    else:
+        ru_id = None
+
```
dcdarrell9: Can we do this outside, or does it need to be in create_message()?

frontstage/views/sign_in.py
andrewmil: Torrance is working on this changeover now

frontstage/views/secure_messaging.py
andrewmil: Moved it outside and now doing it in drafts as well as its an extra 2 lines of code

PR #88, opened 2017-08-03
Replaced use of user_uuid with party_id

AndrewTorrance:
Replaced user_uuid with party_id and tested locally

### DIFF
--- a/frontstage/views/sign_in.py
+++ b/frontstage/views/sign_in.py
@@ -102,7 +102,6 @@ def login():
             "scope": token['scope'],
             "expires_at": token['expires_at'],
             "username": username,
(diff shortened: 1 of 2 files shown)


frontstage/views/sign_in.py
```diff
@@ -102,7 +102,6 @@ def login():
             "scope": token['scope'],
             "expires_at": token['expires_at'],
             "username": username,
-            "user_uuid": "ce12b958-2a5f-44f4-a6da-861e59070a32",
```
dcdarrell9: I think this id needs to be in, but maybe as a different one (and different name)? At the moment I use "db036fd7-ce17-40c2-a8fc-932e7c228397"

PR #89, opened 2017-08-03
D 149 bug save draft

rhiGriff:
Draft can now be modified then sent, and the draft being deleted on send.


frontstage/views/secure_messaging.py
```diff
@@ -98,6 +98,9 @@ def reply_message(session):
                     'ru_id': 'f1a5e99c-8edf-489a-9c72-6cabe6c387fc',
                     'survey': 'BRES'}
 
+            if "msg_id" in request.form:
```
rjweeks: Can you add a comment saying 'message already saved as draft'. Thanks.

PR #90, opened 2017-08-03
Ts 221 modify secure message collection case hardcoded

dcdarrell9:
[Trello Card](https://trello.com/c/zKGpcaUs)

removed hardcoding for collection case in create message

### DIFF
--- a/frontstage/views/secure_messaging.py
+++ b/frontstage/views/secure_messaging.py
@@ -42,7 +42,7 @@ chrome_driver = "{}/tests/selenium_scripts/drivers/chromedriver".format(os.envir
 def create_message(session):
     """Handles sending of new message"""
 
-    url = app.config['RM_CASE_GET_BY_PARTY'].format(app.config['RM_CASE_SERVICE'], session['partyId'])
+    url = app.config['RM_CASE_GET_BY_PARTY'].format(app.config['RM_CASE_SERVICE'], session['party_id'])
     # url = "http://localhost:8050/api/party-api/respondents/id/" + session['user_uuid']
     collection_response = requests.get(url)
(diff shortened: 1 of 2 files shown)


frontstage/views/secure_messaging.py
```diff
@@ -42,12 +42,24 @@
 def create_message(session):
     """Handles sending of new message"""
 
+    url = app.config['RM_CASE_GET_BY_PARTY'].format(app.config['RM_CASE_SERVICE'], session['user_uuid'])
```
andrewmil: still using `user_uuid` should be `party_id` as you've got @AndrewTorrance changes

frontstage/views/secure_messaging.py
```diff
@@ -42,12 +42,24 @@
...
+    # url = "http://localhost:8050/api/party-api/respondents/id/" + session['user_uuid']
+    collection_response = requests.get(url)
+    if collection_response.status_code != 200:
+        return redirect(url_for('error_bp.error_page'))
+    collection_response_json = collection_response.json()
+    collection_id = collection_response_json[0].get('id')
+    if collection_id:
+        collection_case = collection_id
+    else:
+        collection_case = None
```
andrewmil: Could also have this as a separate function as we're using with reply as well but that could be done in that card.

frontstage/views/secure_messaging.py
dcdarrell9: 👍  Apparently the merge didn't work on my terminal.

frontstage/views/secure_messaging.py
dcdarrell9: Chuck it into the other card, just in case there is anything different.

PR #97, opened 2017-08-07
Ts 227 improve error handling

andrewmil:
- Added more specific error pages for 404 and 500 errors
- Added general error handlers for 404, 500 and ConnectionErrors which work across the whole of frontstage
- Added custom exception ExternalServiceError for handling general error responses for calls to external services
- Added error handler for ExternalServiceError which logs out the url and status code returned from the external service
- Applied new error handlers across all of the frontstage views files


frontstage/views/register.py
```diff
@@ -34,6 +35,8 @@ def validate_enrolment_code(enrolment_code):
 
     if result.status_code == 200 and json.loads(result.text)['active']:
         case_id = json.loads(result.text)['caseId']
+    elif result.status_code != 200:
```
dcdarrell9: Should it be everything other than 200 or should it be more specific? (I know it's probably difficult to say)

andrewmil: So it depends on the service, Some services have more specific errors we know they spit out, some have 204's when you look for something that's not found for example.
For any response other than the service specific ones, e.g. 400, 400, 500 we're always just going to be going to a general error page so I feed them all into the ExternalServiceError exception. That will log out the url it was hitting and the status code it returns

dcdarrell9: 👍 
Would it be worth maybe having a list of error codes so that when we get one we could pass it through that list then link it to that page, and if not it would go to the default?

dcdarrell9:
screen shot 2017-08-07 at 11 36 42

So, I hit the pages in the right places, so all good to me 👍

frontstage/views/register.py
andrewmil: I think that could be a good idea for the more standard errors that we could expect from other services, 400, 404. 500 etc. For the service specific one's we'll have to handle them individually

PR #98, opened 2017-08-07
adding 202 response for failed iac call

andrewmil:
Small change to return a 202 status code on a failed iac code request

### DIFF
--- a/frontstage/views/register.py
+++ b/frontstage/views/register.py
@@ -79,6 +79,9 @@ def register():
                 }
             }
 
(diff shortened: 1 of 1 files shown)


dcdarrell9:
I can't comment on the line, but line 66 has a 'TODO'... Does this need to be done?

andrewmil:
Potentially yes though I don't really understand what it means. It has party id but it also posts a created_by? Surely it's created by whoever made the request, i.e. the user?

dcdarrell9:
I'm not sure, maybe worth speaking to Mark about it.

PR #100, opened 2017-08-08
making variable name changes asked for by mark

andrewmil:
Made small variable name changes to register and tests to integrate with party service registration.


andrewmil:
id is what is coming from the party service so we dont have a choice at the moment!

PR #103, opened 2017-08-08
removing commented code from sign in and register

andrewmil:
^ What title says

### DIFF
--- a/frontstage/views/register.py
+++ b/frontstage/views/register.py
@@ -116,10 +116,7 @@ def register_confirm_organisation_survey():
 
     # TODO More error handling e.g. cater for case not coming back from the case service, etc.
 
-    # TODO Use ras-common for this lookup
     # Look up the case by case_id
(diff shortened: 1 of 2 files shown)


frontstage/views/register.py
```diff
@@ -230,91 +220,13 @@ def register_enter_your_details():
...
 
-        # Lets try and create this user on the OAuth2 server
-        # OAuth_payload = {
-        #     "username": email_address,
-        #     "password": password,
-        #     "client_id": app.config['RAS_FRONTSTAGE_CLIENT_ID'],
-        #     "client_secret": app.config['RAS_FRONTSTAGE_CLIENT_SECRET']
-        # }
-
-        headers = {'content-type': 'application/x-www-form-urlencoded'}
```
emilioward: Does this need to be removed?

andrewmil: The headers are defined at line 280 of the new version. It was being defined twice in a row so this one wasn't used

PR #104, opened 2017-08-08
removing unused sign-in endpoints

andrewmil:
I've kept the templates in case these end up being useful at some point


dcdarrell9:
Are these completely unused?

andrewmil:
Yep this isn't used anywhere at the moment as far as I could find

PR #107, opened 2017-08-09
Fixed 500 error on draft save

emilioward:
This PR fixes the 500 error draft save (Trello cards - https://trello.com/c/6DBcxJRi & https://trello.com/c/ShZt86uH) 

To test create a new draft with an error in subject and body, save, then resolve the error and when you save test that it doesn't go to 500 error page.


dcdarrell9:
👍

PR #108, opened 2017-08-09
Bring JWT authentication into frontstage.

rjweeks:


frontstage/common/authorisation.py
```diff
@@ -0,0 +1,26 @@
+from functools import wraps
+from flask import render_template
+from jose.jwt import encode, decode
+
+def jwt_authorisation(request):
```
dcdarrell9: Should it be 'authorization'?

andrewmil: I prefer 'authorization' as well, just because that's what we're calling the cookie

rjweeks: ok changed name. Change is fully working now

PR #111, opened 2017-08-10
Feature/remove common

rjweeks:
Remove ras-common code dependency.


frontstage/common/cryptographer.py
```diff
@@ -0,0 +1,29 @@
+from Crypto import Random
+from Crypto.Cipher import AES
+from base64 import b64encode, b64decode
+
+
+def encrypt(self, raw_text):
```
andrewmil: These methods are causing issues in the tests for some reason. Things appear to run alright on a standard run though

PR #112, opened 2017-08-10
Tbl084 jwt session needs to validate jwt

andrewmil:
- Added a validate function called by the jwt_authorization decorator
- Scope is not currently in the token provided by the oauth service
- Added JWTValidationError Exception which triggers new error handler for this error
- This error handler redirects to the new 'errors/403' endpoint
- Error endpoints now return appropriate status codes
- Fixed unit tests to work with these changes
- Updated encoded jwt's used in testing to have much later expires at dates


frontstage/common/authorisation.py
```diff
@@ -1,26 +1,68 @@
+import logging
+
+from datetime import datetime
 from functools import wraps
 from flask import render_template
```
dcdarrell9: Think this can be removed.

frontstage/common/authorisation.py
```diff
@@ -1,26 +1,68 @@
...
+                    logger.debug('event="Token decoded successfully"')
+                except JWTError:
+                    logger.warning('event="Unable to decode token", encrypted_jwt_token="{}"'
+                                   .format(encoded_jwt_token))
+                    raise JWTValidationError
+            else:
+                logger.warning('event="No authorization token provided"')
+                raise JWTValidationError
+
+            if app.config['VALIDATE_JWT']:
```
dcdarrell9: I'd argue if 'VALIDATE_JWT' was set to true then wouldn't this always be true so we wouldn't need an if?

frontstage/views/errors.py
```diff
@@ -18,11 +18,14 @@ def default_error_page():
...
 @error_bp.route('/500', methods=['GET', 'POST'])
 def server_error_page():
-    response = make_response(render_template('errors/500-error.html', _theme='default', data={"error": {"type": "failed"}}))
-    return response
+    return render_template('errors/500-error.html', _theme='default', data={"error": {"type": "failed"}}), 500
+
+
+@error_bp.route('/403', methods=['GET', 'POST'])
+def not_logged_in_error_page():
+    return render_template('errors/not-signed-in.html', _theme='default', data={"error": {"type": "failed"}}), 403
```
dcdarrell9: Could be in order? How many different error blueprints will we need?

tests/test_secure-message.py
```diff
@@ -26,11 +26,11 @@
...
-encoded_jwt_token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJyZWZyZXNoX3Rva2VuIjoiNDljZmI5YjYtMDM1Ny00OGFiLWE1ZTQtZDE" \
-                    "0ZGFmOWRhODRiIiwiYWNjZXNzX3Rva2VuIjoiNDg4MTc0NmMtY2UwNS00MzY2LWJlYzgtNmJiMGFkNjAyMzI2Iiwic2NvcGU" \
-                    "iOlsiIl0sImV4cGlyZXNfYXQiOjE1MDIyODc1NTEuNTM2NjU2OSwidXNlcm5hbWUiOiJ0ZXN0dXNlckBlbWFpbC5jb20iLCJ" \
-                    "yb2xlIjoicmVzcG9uZGVudCIsInBhcnR5X2lkIjoiZGIwMzZmZDctY2UxNy00MGMyLWE4ZmMtOTMyZTdjMjI4Mzk3In0.YCI" \
-                    "840GM7ihdSPKRtHI8ab33sQbZFkJfpD3mX82LFrg"
+encoded_jwt_token = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJyZWZyZXNoX3Rva2VuIjoiNmY5NjM0ZGEtYTI3ZS00ZDk3LWJhZjktNjN" \
+                    "jOGRjY2IyN2M2IiwiYWNjZXNzX3Rva2VuIjoiMjUwMDM4YzUtM2QxOS00OGVkLThlZWMtODFmNTQyMDRjNDE1Iiwic2NvcGU" \
+                    "iOlsiIl0sImV4cGlyZXNfYXQiOjE4OTM0NTk2NjEuMCwidXNlcm5hbWUiOiJ0ZXN0dXNlckBlbWFpbC5jb20iLCJyb2xlIjo" \
+                    "icmVzcG9uZGVudCIsInBhcnR5X2lkIjoiZGIwMzZmZDctY2UxNy00MGMyLWE4ZmMtOTMyZTdjMjI4Mzk3In0.hh9sFpiPA-O" \
+                    "8kugpDi3_GSDnxWh5rz2e5GQuBx7kmLM"
```
dcdarrell9: Why did the token change?

frontstage/common/authorisation.py
andrewmil: You're right, `VALIDATE_JWT` should either be set by an environment or just implied true

tests/test_secure-message.py
andrewmil: The old one had expired so tests were failing

andrewmil: Thinking about it there may be security concerns about keeping this environment variable here?

frontstage/views/errors.py
andrewmil: Will put them in order. 

Not sure how many more we'll need, these are all the cases I'd think we'd want for MVP

tests/test_logging.py
```diff
@@ -1,17 +0,0 @@
...
-    """Test case for logging"""
-
-    def test_basic_logger_config(self):
-        """Test logger configuration"""
-        with mock.patch('logging.basicConfig') as loggingConfig:
-            logger_initial_config(service_name='ras-secure-message', log_level='INFO', logger_format="message", logger_date_format='2017-06-13')
-            loggingConfig.assert_called_with(level='INFO', format="message", datefmt='2017-06-13')
-
-if __name__ == '__main__':
-    unittest.main()
```
andrewmil: I didn't mean to delete this but what is it actually doing? why is there a .main() here?

tests/test_secure-message.py
dcdarrell9: True, not sure how some of these tokens are supposed to work though, and I know there's a "SECRET_KEY" in config.py that's set to something like "needs to be changed".

tests/test_logging.py
dcdarrell9: Probably worth leaving it there and having a look at some refactoring, I'm not sure what it's doing though.

tests/test_secure-message.py
MIslam92: GJ Drew 👍

andrewmil: Cheers Sam 🕴

PR #114, opened 2017-08-11
Improve unittests

rhiGriff:


tests/test_account_activation.py
```diff
@@ -88,9 +77,6 @@ def test_email_resent_page(self, mock_object):
         # TODO check that the email was actually re-sent once the backend functionality has been developed
```
dcdarrell9: Is this 'todo' going to be done?

tests/test_app.py
```diff
@@ -89,7 +84,6 @@ def test_logged_in(self):
         response = self.app.get('/surveys/', headers=self.headers)
 
         self.assertEqual(response.status_code, 200)
-        self.assertTrue(bytes('Error! You are not logged in!', encoding='UTF-8') in response.data)
```
dcdarrell9: Should we actually test them logging in?

tests/test_account_activation.py
```diff
@@ -65,9 +59,6 @@ def test_activate_account_valid_token_specified(self, mock_object):
         response = self.app.get('/register/activate-account/' + token, headers=self.headers)
 
         self.assertEqual(response.status_code, 302)
-        self.assertTrue(
-            bytes('You should be redirected automatically to target URL', encoding='UTF-8') in response.data)
-        self.assertTrue('/sign-in/?account_activated=True'.encode() in response.data)
 
     # ============== YOUR LINK HAS EXPIRED PAGE ===============
```
andrewmil: Do we want to keep these? we don't have them in the other test files I don't think

tests/test_account_activation.py
```diff
@@ -39,9 +39,6 @@ def test_activate_account_invalid_token_specified(self, mock_object):
         response = self.app.get('/register/activate-account/' + token, headers=self.headers)
 
         self.assertEqual(response.status_code, 302)
```
andrewmil: We should be testing more than just the status code response, We know it's redirecting yeah but we should know it's going to the right place.

If you want to go to the actual final route rather than stopping at the redirect you can add `follow_redirects=True` to the get request

dcdarrell9:
We should probably decide how we want to layout the tests, like do we need a new line inbetween, and so on... There's a few 'todo's' there that we should probably look at.

dcdarrell9:
I think having the `assert(bytes)` tests are useful, but we could maybe look at doing something else?

PR #124, opened 2017-08-15
Ts 222 modify hardcoded survey id in secure message

rhiGriff:


frontstage/views/secure_messaging.py
```diff
@@ -74,23 +74,40 @@ def get_collection_case(party_id):
     return collection_case
 
 
+def get_survey_id(party_id):
```
andrewmil: I'm actually thinking that we should have a little more logging on these functions. Maybe log that you're trying to retrieve this data then also log out the survey name at the end. Both probably debug

dcdarrell9:
The only thing I'd mention is the part where we do 'if this: that.. else: None'....
Should we have tests to see whether we hit the None and whether or not we need it?

andrewmil:
I'm not really sure about that part. In theory for MVP we should never hit the problem of a user existing at this point without those values coming from party service. Potentially we should actually be sending them to an error page + logging it out

PR #125, opened 2017-08-15
updating handling of email verification response+tests

andrewmil:
https://trello.com/c/sSMpZqHh/488-email-verification-handle-expired-token-correctly-13

Handling response of email verification call using status codes rather than content of response
Improved error handling
Improved tests

### DIFF
--- a/frontstage/views/register.py
+++ b/frontstage/views/register.py
@@ -331,41 +331,47 @@ def register_activate_account(token):
 
     # Call the Party service to try to activate the account corresponding to the token that was supplied
     url = app.config['RAS_PARTY_VERIFY_EMAIL'].format(app.config['RAS_PARTY_SERVICE'], token)
+    logger.info('trying to verify email', URL=url)
     result = requests.put(url)
-    logger.debug('"event" : "call party service", "URL" : "{}"'.format(result.content))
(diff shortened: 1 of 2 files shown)


frontstage/views/register.py
```diff
@@ -331,41 +331,47 @@ def register_activate_account(token):
...
     elif result.status_code != 200:
-        # If the token was not recognised, we don't know who the user is so redirect them off to the error page
-        logger.warning('"event" : "unrecognised email token", "Token" : "{}", "Response code" : "{}"'.format(str(token), (str(result.status_code))))
+        raise ExternalServiceError(result)
+
+    json_response = json.loads(result.text)
+
+    if json_response.get('status') == RespondentStatus.ACTIVE.name:
+        # Successful account activation therefore redirect off to the login screen
+        logger.info('user account validated', party_id=json_response.get('id'))
```
emilioward: Do you need to add "event" : "value", "key" : "value" for these logs you created?

andrewmil: So this is the structlog logging format that I mentioned in slack earlier today. We need to have another discussion as a team to decide how we're going to format everything in the next logging review task

PR #126, opened 2017-08-16
Changes to handle failure in support services.

rjweeks:


frontstage/views/secure_messaging.py
```diff
@@ -96,11 +91,17 @@ def get_survey_id(party_id):
...
+        party_id = session['party_id']
+        collection_case = get_collection_case(party_id)
+        if collection_case is None:
+            return redirect(url_for('error_bp.default_error_page'))
+        ru_id = get_party_ru_id(party_id)
+        if ru_id is None:
+            return redirect(url_for('error_bp.default_error_page'))
+        survey_name = get_survey_id(party_id)
+        if survey_name is None:
+            return redirect(url_for('error_bp.default_error_page'))
```
andrewmil: This could be shortened but this can be left for refactoring

PR #132, opened 2017-08-18
Feature/updating logging

andrewmil:
The logger_config I use here gives out json formatted logs.

@rhiGriff has a different config for secure-message which gives slightly differently formatted logs, we should agree which we want to use and make them consistent with each other

### DIFF
--- a/frontstage/common/post_event.py
+++ b/frontstage/common/post_event.py
@@ -5,13 +5,16 @@
    Copyright (c) 2017 Crown Copyright (Office for National Statistics)
 
 """
-from json import loads, dumps
+from json import dumps, loads
+import logging
 
 import requests
-from ons_ras_common import ons_env
+from structlog import wrap_logger
 
 from frontstage import app
 
+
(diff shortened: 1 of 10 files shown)


frontstage/common/post_event.py
```diff
@@ -27,22 +30,20 @@ def post_event(case_id, description=None, category=None, party_id=None, created_
...
-        msg = 'description={} category={} party_id={} created_by={}'.format(
-            description, category, party_id, created_by
-        )
-        ons_env.logger.error('Insufficient arguments', arguments=msg)
+        logger.error('Insufficient arguments',
+                     description=description,
+                     category=category,
+                     party_id=party_id,
+                     created_by=created_by)
         return 500, {'code': 500, 'text': 'insufficient arguments'}
```
dcdarrell9: This probably isn't part of this card but this logic looks a bit off to me?

andrewmil: yeah I'm not sure about the post event thing in general, we should probably refactor as part of a later card

PR #134, opened 2017-08-21
added content-type to secure messaging post requests

andrewmil:
Adding `"Content-Type": "application/json"` to the headers for requests to secure-messaging


andrewmil:
We're using it for initialising the app.
I'm just changing the branch we're using since it looks like the one we were pulling from is gone

PR #135, opened 2017-08-21
moved secure-messaging config to config file

andrewmil:
Moved secure-messaging config from own file to config file

May need to review the environment variables used in the deployment environments for this


andrewmil:
We probably need `RAS_SECURE_MESSAGE_SERVICE_HOST`, `RAS_SECURE_MESSAGE_SERVICE_PORT`, and `RAS_SECURE_MESSAGE_SERVICE_PROTOCOL` environment variables to be set in the environments

PR #140, opened 2017-08-22
improved tests for surveys blueprint

andrewmil:
Redid the tests for the surveys.py file

### DIFF
--- a/tests/test_surveys.py
+++ b/tests/test_surveys.py
@@ -1,3 +1,4 @@
+import io
 import json
 import unittest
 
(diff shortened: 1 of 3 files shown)


tests/test_surveys.py
```diff
@@ -12,12 +13,16 @@
 with open('tests/test_data/collection_instrument.json') as json_data:
     collection_instrument_data = json.load(json_data)
 
-with open('tests/test_data/cases.json') as json_data:
+with open("tests/test_data/cases.json") as json_data:
```
dcdarrell9: Keep the quotes consistent?

PR #147, opened 2017-08-23
 Adding changes to show errors when a users account is locked or a us…

nherriot:
This covers the UI changes for when a user:
1) Has taken more than 10 times to login - and locked their account.
2) A user has not clicked on the verification link to mark their account as verified.

This also covers detecting specific 401's from the OAuth2 server and rendering the correct page - as opposed to raising an exception or continually asking them to sign in again.

One important thing to note is the removal of the OAuth2 library and calling the server directly with a post. This does mean we loose things like checking HTTPS and most importantly the methods to handle other grant_types. We only use grant_type password at the moment, and disregard grant_type client_type. But thoughts on this change would be great.

### DIFF
--- a/frontstage/views/sign_in.py
+++ b/frontstage/views/sign_in.py
@@ -68,7 +68,7 @@ def login():
                                            data={"error": {"type": "failed"}})
             if oauth2_response.status_code != 201:
                 logger.error('Unknown error from the OAuth2 server', ouath2_response=oauth2_response.txt,  status_code=oauth2_response.status_code)
-                raise ExternalServiceError(oauth2_response.text)
+                raise ExternalServiceError(oauth2_response)
             logger.debug('Access Token Granted')
         except (requests.ConnectTimeoutConnectionError, requests.ConnectionError) as e:
             logger.warning('Connection error between the server and the OAuth2 service of: {}'.format(exception=str(e)))
--- a/frontstage/views/sign_in.py
+++ b/frontstage/views/sign_in.py
@@ -6,7 +6,7 @@ from oauthlib.oauth2 import LegacyApplicationClient, MissingTokenError
 import requests
 from requests_oauthlib import OAuth2Session
 from structlog import wrap_logger
-
+import json
 from frontstage import app
 from frontstage.exceptions.exceptions import ExternalServiceError
 from frontstage.jwt import encode
@@ -29,27 +29,49 @@ def login():
         username = request.form.get('username')
         password = request.form.get('password')
 
-        # Creates a 'session client' to interact with OAuth2. This provides a client ID to our client that is used to
-        # interact with the server.
-        client = LegacyApplicationClient(client_id=app.config['RAS_FRONTSTAGE_CLIENT_ID'])
-
-        # Populates the request body with username and password from the user
-        client.prepare_request_body(username=username, password=password, scope=['ci.write', 'ci.read'])
-
-        # passes our 'client' to the session management object. this deals with
-        # the transactions between the OAuth2 server
-        oauth = OAuth2Session(client=client)
-        token_url = app.config['ONS_OAUTH_PROTOCOL'] + app.config['ONS_OAUTH_SERVER'] + app.config['ONS_TOKEN_ENDPOINT']
-
+        # TODO Consider moving this to a helper function.
+        # Lets get a token from the OAuth2 server
         try:
-            token = oauth.fetch_token(token_url=token_url, username=username, password=password, client_id=app.config['RAS_FRONTSTAGE_CLIENT_ID'],
-                                      client_secret=app.config['RAS_FRONTSTAGE_CLIENT_SECRET'])
+            token_url = app.config['ONS_OAUTH_PROTOCOL'] + app.config['ONS_OAUTH_SERVER'] + app.config['ONS_TOKEN_ENDPOINT']
+            data = {
+                'grant_type': 'password',
+                'client_id': app.config['RAS_FRONTSTAGE_CLIENT_ID'],
+                'client_secret': app.config['RAS_FRONTSTAGE_CLIENT_SECRET'],
+                'username': username,
+                'password': password,
+            }
+            headers ={
+                'Accept': 'application/json',
+                'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+            }
 
+            oauth2_response = requests.post(url=token_url, data=data , headers=headers, auth=(app.config['RAS_FRONTSTAGE_CLIENT_ID'], app.config['RAS_FRONTSTAGE_CLIENT_SECRET']))
+            oauth2_token = json.loads(oauth2_response.text)
+            # Check to see that this user has not attempted to login too many times or that they have not forgot to
+            # click on the activate account URL in their email by checking the error message back from the OAuth2 server
+            if oauth2_response.status_code == 401:
+                oauth2Error = json.loads(oauth2_response.text)
+                if oauth2Error['detail']== 'Unauthorized user credentials':
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form, data={"error": {"type": "failed"}})
+                elif 'User account locked' in oauth2Error['detail']:
+                    logger.warning('User account is locked on the OAuth2 server')
+                    return render_template('sign-in/sign-in.trouble.html', _theme='default', form=form,
+                                           data={"error": {"type": "account locked"}})
+                elif 'User account not verified' in oauth2Error['detail']:
+                    logger.warning('User account is not verified on the OAuth2 server')
+                    return render_template('sign-in/sign-in.account-not-verified.html', _theme='default', form=form,
+                                           data={"error": {"type": "account not verified"}})
+                else:
+                    logger.error('OAuth 2 server generated 401 which is not understood. Details are: {}'.format(oauth2Error['detail']))
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form,
+                                           data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server which was: {}'.format(oauth2_response.text))
+                raise ExternalServiceError(oauth2_response.text)
             logger.debug('Access Token Granted')
-
-        except MissingTokenError as e:
-            logger.warning('Missing token', exception=str(e))
-            return render_template('sign-in/sign-in.html', _theme='default', form=form, data={"error": {"type": "failed"}})
+        except (requests.ConnectTimeoutConnectionError, requests.ConnectionError) as e:
+            logger.warning('Connection error between the server and the OAuth2 service of: {}'.format(exception=str(e)))
+            raise ExternalServiceError(e)
 
         url = app.config['RAS_PARTY_GET_BY_EMAIL'].format(app.config['RAS_PARTY_SERVICE'], username)
         req = requests.get(url, verify=False)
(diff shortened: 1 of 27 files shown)


run.py
```diff
@@ -16,6 +16,6 @@
         logger.info('* Cloud Foundry port "{}"'.format(port))
         logger.info('* Cloud Foundry protocol "{}"'.format(protocol))
     else:
-        port = os.getenv('FS_DEV_PORT', 5001)
+        port = os.getenv('FS_DEV_PORT', 8080)
```
andrewmil: We should probably not change this as 5001 is what most devs will be used to for frontstage at the moment

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+            }
+
+            oauth2_response = requests.post(url=token_url, data=data , headers=headers, auth=(app.config['RAS_FRONTSTAGE_CLIENT_ID'], app.config['RAS_FRONTSTAGE_CLIENT_SECRET']))
+            oauth2_token = json.loads(oauth2_response.text)
+            # Check to see that this user has not attempted to login too many times or that they have not forgot to
+            # click on the activate account URL in their email by checking the error message back from the OAuth2 server
+            if oauth2_response.status_code == 401:
+                oauth2Error = json.loads(oauth2_response.text)
+                if oauth2Error['detail']== 'Unauthorized user credentials':
```
andrewmil: space missing before `==`

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                                           data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server which was: {}'.format(oauth2_response.text))
+                raise ExternalServiceError(oauth2_response.text)
             logger.debug('Access Token Granted')
-
-        except MissingTokenError as e:
-            logger.warning('Missing token', exception=str(e))
-            return render_template('sign-in/sign-in.html', _theme='default', form=form, data={"error": {"type": "failed"}})
+        except (requests.ConnectTimeoutConnectionError, requests.ConnectionError) as e:
```
andrewmil: This try/except doesn't have to wrap around this much, We could just wrap it around line 48 where you make the request

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                'client_secret': app.config['RAS_FRONTSTAGE_CLIENT_SECRET'],
+                'username': username,
+                'password': password,
+            }
+            headers ={
+                'Accept': 'application/json',
+                'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+            }
+
+            oauth2_response = requests.post(url=token_url, data=data , headers=headers, auth=(app.config['RAS_FRONTSTAGE_CLIENT_ID'], app.config['RAS_FRONTSTAGE_CLIENT_SECRET']))
```
andrewmil: a couple of these lines are a bit too long, could space them over multiple lines

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                elif 'User account locked' in oauth2Error['detail']:
+                    logger.warning('User account is locked on the OAuth2 server')
+                    return render_template('sign-in/sign-in.trouble.html', _theme='default', form=form,
+                                           data={"error": {"type": "account locked"}})
+                elif 'User account not verified' in oauth2Error['detail']:
+                    logger.warning('User account is not verified on the OAuth2 server')
+                    return render_template('sign-in/sign-in.account-not-verified.html', _theme='default', form=form,
+                                           data={"error": {"type": "account not verified"}})
+                else:
+                    logger.error('OAuth 2 server generated 401 which is not understood. Details are: {}'.format(oauth2Error['detail']))
```
andrewmil: To match how we do our other logging this would be
`logger.error('OAuth 2 server generated 401 which is not understood', details=oauth2Error['detail']))`

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form,
+                                           data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server which was: {}'.format(oauth2_response.text))
```
andrewmil: Similar to above
`logger.error('Unknown error from the OAuth2 server', details=oauth2_response.text`

andrewmil: Should maybe log out the status_code here aswell

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
-
+            token_url = app.config['ONS_OAUTH_PROTOCOL'] + app.config['ONS_OAUTH_SERVER'] + app.config['ONS_TOKEN_ENDPOINT']
+            data = {
+                'grant_type': 'password',
+                'client_id': app.config['RAS_FRONTSTAGE_CLIENT_ID'],
+                'client_secret': app.config['RAS_FRONTSTAGE_CLIENT_SECRET'],
+                'username': username,
+                'password': password,
+            }
+            headers ={
```
andrewmil: space before `{`

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+                    logger.warning('User account is not verified on the OAuth2 server')
+                    return render_template('sign-in/sign-in.account-not-verified.html', _theme='default', form=form,
+                                           data={"error": {"type": "account not verified"}})
+                else:
+                    logger.error('OAuth 2 server generated 401 which is not understood. Details are: {}'.format(oauth2Error['detail']))
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form,
+                                           data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server which was: {}'.format(oauth2_response.text))
+                raise ExternalServiceError(oauth2_response.text)
```
andrewmil: ExternalServiceError takes a response object as a parameter, so here `ExternalServiceError(oauth2_response`

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+        # TODO Consider moving this to a helper function.
+        # Lets get a token from the OAuth2 server
         try:
-            token = oauth.fetch_token(token_url=token_url, username=username, password=password, client_id=app.config['RAS_FRONTSTAGE_CLIENT_ID'],
-                                      client_secret=app.config['RAS_FRONTSTAGE_CLIENT_SECRET'])
-
+            token_url = app.config['ONS_OAUTH_PROTOCOL'] + app.config['ONS_OAUTH_SERVER'] + app.config['ONS_TOKEN_ENDPOINT']
+            data = {
+                'grant_type': 'password',
+                'client_id': app.config['RAS_FRONTSTAGE_CLIENT_ID'],
```
jonmorphic: client_id / secret not needed here, apparently

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,49 @@ def login():
...
+            oauth2_token = json.loads(oauth2_response.text)
+            # Check to see that this user has not attempted to login too many times or that they have not forgot to
+            # click on the activate account URL in their email by checking the error message back from the OAuth2 server
+            if oauth2_response.status_code == 401:
+                oauth2Error = json.loads(oauth2_response.text)
+                if oauth2Error['detail']== 'Unauthorized user credentials':
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form, data={"error": {"type": "failed"}})
+                elif 'User account locked' in oauth2Error['detail']:
+                    logger.warning('User account is locked on the OAuth2 server')
+                    return render_template('sign-in/sign-in.trouble.html', _theme='default', form=form,
```
jonmorphic: Why `sign-in.trouble.html`? rather than `sign-in.locked.html` or similar?

frontstage/views/sign_in.py
jonmorphic: Don't the client_id/secret here, apparently.

frontstage/views/sign_in.py
```diff
@@ -65,10 +87,10 @@ def login():
...
 
         data_dict_for_jwt_token = {
-            "refresh_token": token['refresh_token'],
-            "access_token": token['access_token'],
-            "scope": token['scope'],
-            "expires_at": token['expires_at'],
+            "refresh_token": oauth2_token['refresh_token'],
+            "access_token": oauth2_token['access_token'],
+            "scope": oauth2_token['scope'],
+            "expires_at": oauth2_token['expires_in'],
```
jonmorphic: Why's this renamed to `expires_at`, I think it's a time-delta rather than a specific time?

andrewmil:
Currently only happy path is tested in `tests/test_sign_in.py'
We can either add the couple of other tests required in this request or I could add a new task that I/one of other QA guys could pick up to add them

jonmorphic:
In that case let’s stick with current approach, as long as it gets tested somehow before preprod.

run.py
nherriot: I think we should change this as frontstage should be the one service that does use 80 or 8080. I sent out a proposal probably 1 - 2 months ago and included RM guys too. The docker-compose files have now been updated to this. So unless it clashes with other services I think we should update to a standard port for web UI.

frontstage/views/sign_in.py
nherriot: done.

frontstage/views/sign_in.py
nherriot: done.

frontstage/views/sign_in.py
nherriot: This can produce undesirable results. Using the {}.format method works with unicode and 8bit. It's the other way round we should be changing to using .format i think... :-) :-) :-) 
https://docs.python.org/3/whatsnew/2.6.html#pep-3101

nherriot: OK. It's done.

frontstage/views/sign_in.py
nherriot: good point.

frontstage/views/sign_in.py
nherriot: i think it reads better no? it also allows us to place multiple excepts for other errors that may turn up in testing.

andrewmil: Fair it's not changing anything functional so should be fine
We also have a general error handler which deals with connection errors but that's probably getting changed some time soon

frontstage/views/sign_in.py
nherriot: No they are not. However I'm not sure how other implementations of an OAuth2 server will interpret the spec's. It is also being explicit on the grant_type flow by showing the 4 parameters. Any objections in keeping them in until we test against UAA?

jonmorphic: No objections, it was a minor point.

frontstage/views/sign_in.py
nherriot: It's the one specified by Nick UI. He presented a specific type of text bla bla bla...

frontstage/views/sign_in.py
nherriot: The Oauth2 lib we used I believe was implementing some functionality to provide an expires at. Our system has since been built to decode this. In actual fact - the specs say expires_in.
https://tools.ietf.org/html/rfc6749#appendix-A.14

frontstage/views/sign_in.py
```diff
@@ -29,27 +29,50 @@ def login():
...
+                    logger.warning('User account is not verified on the OAuth2 server')
+                    return render_template('sign-in/sign-in.account-not-verified.html', _theme='default', form=form,
+                                           data={"error": {"type": "account not verified"}})
+                else:
+                    logger.error('OAuth 2 server generated 401 which is not understood', oauth2error=oauth2Error['detail'])
+                    return render_template('sign-in/sign-in.html', _theme='default', form=form,
+                                           data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server', ouath2_response=oauth2_response.txt,  status_code=oauth2_response.status_code)
+                raise ExternalServiceError(oauth2_response.text)
```
andrewmil: ExternalServiceError takes a response object so this should be `ExternalServiceError(oauth2_response)`

PR #150, opened 2017-08-24
Name of message from not displayed 

rhiGriff:


rjweeks:
Agree with Andrew's suggestion. Please add that change to this PR. Thanks.

PR #151, opened 2017-08-24
Adding google analytics to the footer

andrewmil:
To enable the google analytics set environment variable `GOOGLE_ANALYTICS` to the correct code.

If no code is given the script will not be included


dcdarrell9:
Is footer used everywhere? Would it be better to have the analytics in the _base template?

andrewmil:
Base includes footer so there's no real difference I don't think

rjweeks:
Can we do some checks with the developer tools to see if the javascript is delivered on different pages?

andrewmil:
I had a quick look through all the pages I could get to and it was still there. We only have codes for pre-prod and prod so I think the idea is once it gets released we ask Chris Au if he's getting out info through

PR #156, opened 2017-08-25
Feature/improve access surveys authentication

andrewmil:
Factored out reused code in separate functions and improved logging


andrewmil:
Checked this on the test environment, everything appears to be working fine

PR #157, opened 2017-08-29
Defect 16 - 'Submit by' column must not be present on History page

andrewmil:
`surveys.html` now checks for a `history` variable which if `True` removes the submit-by field

### DIFF
--- a/frontstage/templates/surveys/surveys.html
+++ b/frontstage/templates/surveys/surveys.html
@@ -11,7 +11,9 @@
         </div>
 
         <div id="SUBMIT_BY_COLUMN_HEADING" class="grid__col col-2@m">
+            {% if not history %}
             Submit by
+            {% endif %}
         </div>
 
         <div id="STATUS_COLUMN_HEADING" class="grid__col col-2@m">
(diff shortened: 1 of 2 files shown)


frontstage/templates/surveys/surveys.html
```diff
@@ -11,7 +11,9 @@
         </div>
 
         <div id="SUBMIT_BY_COLUMN_HEADING" class="grid__col col-2@m">
+            {% if not history %}
             Submit by
```
dcdarrell9: I'd probably say to indent this, but that's preference.

frontstage/templates/surveys/surveys.html
```diff
@@ -40,8 +42,10 @@
         </div>
 
         <div class="grid__col col-2@m">
+            {% if not history %}
             <span class="survey-list-item__label mercury">SUBMIT BY: </span>
             <span>{{ data.collectionExerciseData.scheduledReturnDateTimeFormatted}}</span>
```
dcdarrell9: same here with indent, again preference.

frontstage/templates/surveys/surveys.html
andrewmil: Not sure what the standard is here, I usually don't indent on the if/for statements so the html stays properly indented

PR #164, opened 2017-08-30
Unread messages count

rhiGriff:
Adding unread messages count to show on all message tabs (draft and sent). Also updated urls in test_secure-message.py to use urls from config not hardcoded urls


frontstage/views/secure_messaging.py
```diff
@@ -306,14 +306,16 @@ def messages_get(session, label="INBOX"):
...
 
     response_data = json.loads(resp.text)
-    total_msgs = 0
 
-    for x in range(0, len(response_data['messages'])):
-        if "UNREAD" in response_data['messages'][x]["labels"]:
-            total_msgs += 1
+    labels = app.config['LABELS_GET_API_URL']
+
+    totals = requests.get(labels, headers=headers)
```
rjweeks: Change totals to response

frontstage/views/secure_messaging.py
```diff
@@ -306,14 +306,16 @@ def messages_get(session, label="INBOX"):
...
+
+    total = json.loads(totals.text)
```
rjweeks: change to unread_msg_total

frontstage/views/secure_messaging.py
```diff
@@ -306,14 +306,16 @@ def messages_get(session, label="INBOX"):
...
+
     logger.info('Retrieved messages successfully')
     return render_template('secure-messages/secure-messages.html', _theme='default', messages=response_data['messages'],
-                           links=response_data['_links'], label=label, total=total_msgs)
+                           links=response_data['_links'], label=label, total=total['total'])
```
rjweeks: I guess this will need update because of my previous comment.

frontstage/views/secure_messaging.py
andrewmil: We should have error handling on this request as we do for our other responses.
Check the response status code to check that it's right and handle whatever error status codes can be returned

andrewmil: We should also add a test which covers what happens when we dont get the response we want from this request

frontstage/views/secure_messaging.py
```diff
@@ -305,14 +305,19 @@ def messages_get(session, label="INBOX"):
...
-    total_msgs = 0
 
-    for x in range(0, len(response_data['messages'])):
-        if "UNREAD" in response_data['messages'][x]["labels"]:
-            total_msgs += 1
+    labels = app.config['LABELS_GET_API_URL']
+
+    unread_label_data = requests.get(labels, headers=headers)
+    if unread_label_data.status_code != 200:
+        logger.error('Failed to retrieve unread label data')
```
andrewmil: I'd add party_id to this log line so we know what user this has failed for

PR #175, opened 2017-08-31
Accessibility improvements

nicprice:
Changed the value of 'aria-controls' to match corresponding ID of the guidance information it refers to.

The "aria-controls" attribute value should now point to the corresponding guidance information block in the html to which it refers.


andrewmil:
I'm not sure exactly what visual changes we're trying to make here. Are there any screens to compare with? I'll upload pictures here where I do see visual changes

andrewmil:
screen shot 2017-08-31 at 16 54 34

andrewmil:
Happy to approve if we're happy with the design of that screen

nicprice:
@andrewmil - the button text should still show as "Access survey"

I added in what should be visually hidden survey title, so a screenreader will read out (in this case) "Access business register and employment survey", but the button will show as "Access survey"

I wonder if we're missing the u-vh class from our CSS?

andrewmil:
screen shot 2017-09-01 at 14 02 37

This is how the messages tab now looks with css

andrewmil:
screen shot 2017-09-01 at 14 03 56

This is how messages tab looks with no css

PR #177, opened 2017-09-01
Feature/improve quality of bdd selenium tests

dcdarrell9:
- Added more tests in.
- 'run' 'run_behave_tests' or 'run_tests' to be able to get the tests running.
- Suggest any additional tests we may need, and any improvements.


tests/behavioural/features/steps/test_send_message.py
```diff
@@ -30,6 +34,42 @@ def step_impl_visit_create_message_page(context):
 def step_impl_add_text_to_subject_and_body(context):
     context.browser.find_by_id('secure-message-subject').send_keys('Test Subject')
     context.browser.find_by_id('secure-message-body').send_keys('Test Body')
+
+
+@when('I have a message with non alpha characters')
+def step_impl_add_non_alpha_text_to_subject_and_body(context):
+    context.browser.find_by_id('secure-message-subject').send_keys('[]!@£$%^&*()')
+    context.browser.find_by_id('secure-message-body').send_keys('"?><:|}{+_±')
```
andrewmil: This test wont pass without extra environment variables.
We should either specify these somewhere or reject the characters that aren't acceptable without them

dcdarrell9: So with this I think that we need to know what characters we'd actually need to include.

tests/behavioural/features/steps/common.py
```diff
@@ -2,11 +2,12 @@
 from selenium.common.exceptions import NoSuchElementException
 
 
+# Need an authenticated user
 @given('I am already logged in')
 def step_impl_go_to_sign_in_page(context):
     context.browser.visit('/sign-in/')
-    context.browser.find_by_id('username').send_keys('madpenguinxxx@linux.co.uk')
-    context.browser.find_by_id('password').send_keys('Gizmo007!')
+    context.browser.find_by_id('username').send_keys('edward.stevens@qa.com')
```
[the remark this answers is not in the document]
dcdarrell9: Yeah, it's not ideal, I'll look into it.

tests/behavioural/features/steps/test_send_message.py
andrewmil: good point, I'm not sure that we'd be required to support accepting `±` for example

PR #179, opened 2017-09-05
removing unread label after get

andrewmil:
The unread label is now not removed until after the message is retrieved.

I've also removed the error page being generated when we fail this as seemed a bit unneccessary


frontstage/views/secure_messaging.py
```diff
@@ -389,4 +381,12 @@ def message_get(session, msg_id):
...
     message = json.loads(get_message.text)
     loggerb.debug('Retrieved message successfully')
+
+    loggerb.debug('Attempting to remove unread label')
+    data = {"label": 'UNREAD', "action": 'remove'}
+    url = app.config['MESSAGE_MODIFY_URL'].format(msg_id)
+    response = requests.put(url, data=json.dumps(data), headers=headers)
+    if response.status_code != 200:
+        loggerb.error("Failed to remove unread label")
+
```
tej99: Do you not want to raise ExternalSerivceError if you can't modify the message status?

andrewmil: I've decided not to as I don't think we want to kick the user out over such a small thing

PR #183, opened 2017-09-06
Added cookies and contact information pages

saedwards:


andrewmil:
The text for the cookies looks quite squashed on my laptop not sure if it's supposed to be like that?
image

nicprice:
Looks like the CSS isn't pulling through, or needs updating.

1. There should be padding in the table cells
2. Column headings should be left-aligned
3. All cell content should be vertically aligned top

Here's a screenshot from the research prototype:

screenshot 2017-09-06 14 56 05

PR #184, opened 2017-09-06
Fix/include headers

andrewmil:
Removed `headers` from top level of the file as it was persisting between requests.

Headers is now set at the start of each request. (glad we caught this one)

### DIFF
--- a/frontstage/views/secure_messaging.py
+++ b/frontstage/views/secure_messaging.py
@@ -77,10 +77,7 @@ def create_message(session):
     """Handles sending of new message"""
     if request.method == 'POST':
         party_id = session['party_id']
-        headers = {
-            "Authorization": request.cookies['authorization'],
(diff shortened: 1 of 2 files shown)


frontstage/views/secure_messaging.py
```diff
@@ -79,6 +77,10 @@ def create_message(session):
     """Handles sending of new message"""
     if request.method == 'POST':
         party_id = session['party_id']
+        headers = {
+            "Authorization": request.cookies['authorization'],
+            "Content-Type" : "application/json"
```
jonmorphic: If you use the json keyword (instead of data) on requests.put, then it will do the json conversion and add the application/json header automatically.

andrewmil: Made the change definitely a better way of doing it

PR #185, opened 2017-09-06
removing file name and line number from logger

andrewmil:
Removing file name and line number from logger so splunk can parse the json format


liamtoozer:
Thanks for changing this 👍

PR #190, opened 2017-09-07
removing secure variable defaults

andrewmil:
Removed secure variables (secrets etc) from config.

If you want to run frontstage locally from your machine either set the environment variables on your machine or set `APP_SETTINGS=DevelopmentConfig` to use dummy ones.

The JWT secret must match that in secure-message for this to work


andrewmil:
If any of the settings in DevelopmentConfig are going to be used in deployed environments we will have to remove these as well.

Looking at `SECURITY_USER_NAME`, `SECURITY_USER_PASSWORD`, `RAS_FRONTSTAGE_CLIENT_ID` and `RAS_FRONTSTAGE_PASSWORD` in particular

andrewmil:
I've also got questions about `SM_USER_AUTHENTICATION_(PUBLIC/PRIVATE)_KEY` variables.

These are currently not used in our deployments but when we start encrypting our JWT's we'll have to separate the codes we use for testing from the production code

andrewmil:
https://trello.com/c/eIxe8JJt/564-defect-25-information-disclosed-via-github

^ relevant trello card

PR #194, opened 2017-09-08
Defect-25 Information Disclosed via GitHub Continued

andrewmil:
- Removed default value for JWT_ALGORITHM
- Removed leftover jwt-secret/algorithm from authorisation file
- Added Exception to fail application when a non-defaulted environment variable is not defined

### DIFF
--- a/frontstage/__init__.py
+++ b/frontstage/__init__.py
@@ -5,6 +5,7 @@ from flask import Flask
 from flask_wtf.csrf import CSRFProtect
 from structlog import wrap_logger
 
+from frontstage.exceptions.exceptions import MissingEnvironmentVariable
 from frontstage.filters.case_status_filter import case_status_filter
 from frontstage.filters.file_size_filter import file_size_filter
 from frontstage.logger_config import logger_initial_config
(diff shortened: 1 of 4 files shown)


andrewmil:
Came across a problem, we want `JWT_ALGORITHM` to be protected but if we cant include it in our repo how can we run the tests in travis? It's not like the other keys etc where you can just make up a new one

frontstage/__init__.py
```diff
@@ -15,12 +16,14 @@
 app_config = 'config.{}'.format(os.environ.get('APP_SETTINGS', 'Config'))
 app.config.from_object(app_config)
 
+log_level = 'DEBUG' if app.config['DEBUG'] else None
+logger_initial_config(service_name='ras-frontstage', log_level=log_level)
 logger = wrap_logger(logging.getLogger(__name__))
+logger.debug('{} used'.format(str(app_config)))
 
-logger.error('{} used'.format(str(app_config)))
 for var in app.config['NON_DEFAULT_VARIABLES']:
```
jonmorphic: Suggestion:

```
missing = [app.config.get(var) for var in app.config['NON_DEFAULT_VARIABLES'] if app.config.get(var)]
if missing:
    raise MissingEnvironmentVariables(missing, ...)
```
Reports all missing vars in the event there's more than one.

jonmorphic: Longer-term, we might want to avoid doing this check in the `__init__.py`, as a) it can prevent import of the package which might have unintended consequences and b) it's implicit.

frontstage/exceptions/exceptions.py
```diff
@@ -14,3 +9,19 @@ def __init__(self, response):
...
     pass
+
+
+class MissingEnvironmentVariable(Exception):
+    def __init__(self, app, logger):
+        self.app = app
+        self.logger = logger
+        self.log_missing_env_variables()
+
+    def log_missing_env_variables(self):
```
jonmorphic: Ok, this answers the other comment about logging all the missing variables. It would be better not to implement this logic in an exception (or to tie the exception together with logging), but will approve this as we need a wider discussion around exception handling & logging.

PR #196, opened 2017-09-12
Removing mention of locked/disabled account

nicprice:
Removed mention of locked/disabled account, including in the page <title>

### DIFF
--- a/frontstage/templates/sign-in/sign-in.locked-account.html
+++ b/frontstage/templates/sign-in/sign-in.locked-account.html
@@ -8,6 +8,6 @@
 
     <p>You've tried to sign in a few times with the wrong details.</p>
 
-    <p class="u-mb-l">To access your account please call us on 0300 1234 932</p>
+    <p class="u-mb-l">To access your account please call us on 0300 1234 932 or email <a href="mailto:surveys@ons.gov.uk">surveys@ons.gov.uk</a></p>
 
 {% endblock main %}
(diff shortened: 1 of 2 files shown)


frontstage/templates/sign-in/sign-in.locked-account.html
```diff
@@ -1,16 +1,13 @@
...
-    <h1 class="saturn">We've now locked this account</h1>
-    <p>You've tried to sign in a few times with incorrect details.<br />
-    As a result of this, we've now locked this account for security reasons.</p>
+    <h1 class="saturn">Trouble signing in?</h1>
 
-    <p>Don’t worry, to unlock your account you just need to contact us in one of the following ways:</p>
+    <p>You've tried to sign in a few times with the wrong details.</p>
 
-    <p>Email: <a href="mailto:surveys.ons@gov.uk">surveys.ons@gov.uk</a></p>
+    <p class="u-mb-l">To access your account please call us on 0300 1234 932</p>
```
andrewmil: Are we supposed to also offer an email address to contact or is it just the phone number?

andrewmil: Also if you'd like you can make the number a link so people can call from their laptops if you'd like

`<a href="tel:03001234931">0300 1234 932</a>`

nicprice: You're right, I'll add in the email address.

For the telephone number as a link we've considered this and for time being not doing this, but would be good to discuss post go-live.

PR #199, opened 2017-09-13
Accessibility - Changes to html in footer

nicprice:
Without alt text, the image in the footer failed accessibility compliance.

Added "OGL" as alt text.


nicprice:
Yes thanks @andrewmil these changes aren't visible, but the OGL image now has alternative text, and the list headings in the footer are not longer <h3>s.

PR #200, opened 2017-09-13
adding if for seperate error messages

rhiGriff:
Added if statement to html to accommodate Upload CI error messages

### DIFF
--- a/frontstage/templates/surveys/surveys-upload-failure.html
+++ b/frontstage/templates/surveys/surveys-upload-failure.html
@@ -11,17 +11,17 @@
 
         <div class="panel panel--error u-mb-s" data-qa="error">
             <div class="panel__header">
-                <h1 class="panel__title venus">Something went wrong</h1>
+                 {% if ".xlsx format" in error_info.text %}
+                    <h1 class="panel__title venus">Error uploading - incorrect file type</h1>
+                 {% elif "50 characters" in error_info.text %}
+                    <h1 class="panel__title venus">Error uploading - file name too long</h1>
+                 {% elif "20MB" in error_info.text %}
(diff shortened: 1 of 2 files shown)


frontstage/templates/surveys/surveys-upload-failure.html
```diff
@@ -11,17 +11,23 @@
...
-                    <li>{{ error_info.text }}</li>
-                </ul>
-
-                <p class="error-message">
-                    Please try uploading your spreadsheet again
-                </p>
+                {% if ".xlsx format" not in error_info.text and "50 characters" not in error_info.text and "20MB" not in error_info.text%}
+                    <p class="error-message">Please try uploading your spreadsheet again</p>
+                {% else %}
+                    <p class="error-message">{{ error_info.text }}</p>
```
andrewmil: If the collection instrument falls over/gives out a 500 this is either going to be empty or a huge long exception string.
Should probably have a proper default error message

rhiGriff: Ii checked with Nick on the default error and that's what he said, if none of the above so: extension, name length or file size then it's something went wrong

frontstage/templates/surveys/surveys-upload-failure.html
```diff
@@ -11,17 +11,23 @@
 
         <div class="panel panel--error u-mb-s" data-qa="error">
             <div class="panel__header">
-                <h1 class="panel__title venus">Something went wrong</h1>
+                 {% if ".xlsx format" in error_info.text %}
```
jonmorphic: Would be better to have a designed error protocol between services, e.g a json-formatted response containing a code, with a translation function to map codes to text.

PR #201, opened 2017-09-13
Feature/us012  reset forgotten password

andrewmil:
- Updated password blueprint templates to match story
- Added password reset link expired template `passwords/link-expired.html`
- Added code to passwords blueprint to make and process required calls to party service
- Should be fully tested

https://trello.com/c/02XvOKBs/488-us012-reset-forgotten-password-8sp

For functionality to work will require the party service endpoints to be available

### DIFF
--- a/frontstage/models.py
+++ b/frontstage/models.py
@@ -163,6 +163,12 @@ class ResetPasswordForm(FlaskForm):
                                                 message=app.config['PASSWORD_CRITERIA_ERROR_TEXT'])])
     password_confirm = PasswordField('Re-type new password')
 
+    @staticmethod
+    def validate_password(form, field):
+        password = field.data
+        if password.isalnum() or not any(char.isupper() for char in password) or not any(char.isdigit() for char in password):
+            raise ValidationError(app.config['PASSWORD_CRITERIA_ERROR_TEXT'])
+
 
 class RespondentStatus(enum.IntEnum):
     CREATED = 0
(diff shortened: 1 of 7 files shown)


frontstage/views/passwords.py
```diff
@@ -15,26 +18,29 @@
...
-        return redirect(url_for('passwords_bp.forgot_password_check_email',
-                                _external=True,
-                                _scheme=getenv('SCHEME', 'http')))
+        post_data = {"email_address": email_address}
+
+        url = app.config['RAS_PARTY_RESET_PASSWORD_REQUEST'].format(app.config['RAS_PARTY_SERVICE'])
+        response = requests.post(url, auth=app.config['BASIC_AUTH'], json=post_data, verify=False)
+        if response.status_code == 404:
+            logger.warning('Email address is not registered')
+            template_data = {"error": {"type": {"Email address ia not registered"}}}
```
tej99: Typo, 'is' instead of 'ia'

frontstage/models.py
```diff
@@ -163,6 +163,12 @@ class ResetPasswordForm(FlaskForm):
                                                 message=app.config['PASSWORD_CRITERIA_ERROR_TEXT'])])
     password_confirm = PasswordField('Re-type new password')
 
+    @staticmethod
+    def validate_password(form, field):
```
jonmorphic: Could we have created a custom field validator to do this?

frontstage/views/passwords.py
```diff
@@ -15,26 +18,29 @@
...
-
-        # TODO do some kind of back end processing to validate the email address, error handling, logging
-        # print('Email address=' + email_address)
-
-        return redirect(url_for('passwords_bp.forgot_password_check_email',
-                                _external=True,
-                                _scheme=getenv('SCHEME', 'http')))
+        post_data = {"email_address": email_address}
+
+        url = app.config['RAS_PARTY_RESET_PASSWORD_REQUEST'].format(app.config['RAS_PARTY_SERVICE'])
```
jonmorphic: I don't think you introduced this, but I find this approach to constructing a url a bit odd. :)

andrewmil: I agree, it's been here since I started working on frontstage though so I've ran with it.
After BRES we can go through the whole thing and update

PR #202, opened 2017-09-14
Max content length and routing

tej99:
- Added request size limit (currently hard coded to 20MB)
- Added ErrorHandler to handle when request is too large

### DIFF
--- a/frontstage/__init__.py
+++ b/frontstage/__init__.py
@@ -15,6 +15,7 @@ app = Flask(__name__)
 
 app_config = 'config.{}'.format(os.environ.get('APP_SETTINGS', 'Config'))
 app.config.from_object(app_config)
+app.config['MAX_CONTENT_LENGTH'] = 20 * 1024 * 1024
 
 log_level = 'DEBUG' if app.config['DEBUG'] else None
 logger_initial_config(service_name='ras-frontstage', log_level=log_level)
--- a/frontstage/__init__.py
+++ b/frontstage/__init__.py
@@ -15,7 +15,6 @@ app = Flask(__name__)
 
 app_config = 'config.{}'.format(os.environ.get('APP_SETTINGS', 'Config'))
 app.config.from_object(app_config)
-app.config['MAX_CONTENT_LENGTH'] = 20 * 1024 * 1024
 
 log_level = 'DEBUG' if app.config['DEBUG'] else None
 logger_initial_config(service_name='ras-frontstage', log_level=log_level)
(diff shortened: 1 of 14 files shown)


frontstage/__init__.py
```diff
@@ -15,6 +15,7 @@
 
 app_config = 'config.{}'.format(os.environ.get('APP_SETTINGS', 'Config'))
 app.config.from_object(app_config)
+app.config['MAX_CONTENT_LENGTH'] = 20 * 1024 * 1024
```
andrewmil: This should be set in config.py

frontstage/error_handlers.py
```diff
@@ -47,3 +50,12 @@ def connection_error_jwt_validation(error):  # pylint: disable=unused-argument
...
                             _scheme=getenv('SCHEME', 'http')))
+
+
+@app.errorhandler(RequestEntityTooLarge)
+def request_entity_too_large_error(error):
+    case_id = request.args.get('case_id', None)
+    logger.error('Upload failed', status_code=413, case_id=case_id)
+    error_info = {'status code': 413, 'text': 'The spreadsheet must be smaller than 20MB in size.'}
+    return render_template('surveys/surveys-upload-failure.html', _theme='default', error_info=error_info,
+                           case_id=case_id)
```
andrewmil: We should probably test this as well but could potentially go in without

andrewmil: And even thought this is a failure we may want to restrict access to this page to authorised users

frontstage/error_handlers.py
```diff
@@ -2,8 +2,11 @@
 from os import getenv
 
 from flask import redirect, url_for
+from flask import render_template
+from flask import request
```
andrewmil: Also you can do all these flask imports on the one line

frontstage/views/surveys.py
```diff
@@ -249,6 +260,47 @@ def upload_survey(session):
...
+@surveys_bp.route('/upload_failed', methods=['GET'])
+@jwt_authorization(request)
+def upload_failed(session):
+    """Logged in page for users only."""
+    party_id = session.get('party_id', 'no-party-id')
+    case_id = request.args.get('case_id', None)
+    error_info = request.args.get('error_info', None)
+    logger.error('Upload failed', party_id=party_id, case_id=case_id)
+
+    if error_info == "type":
```
andrewmil: could set these when you set error_info first to avoid more if statements

tej99: If i set error_info to the dict before hand i cant pass it as a string query arg.

PR #204, opened 2017-09-15
Fix/remove oauth token from registration

andrewmil:
We were getting a token from the Ouath2 server which we were sending to the party service. This was never getting used however so we the code/testing around this.


frontstage/views/register.py
```diff
@@ -297,10 +251,10 @@ def register_enter_your_details():
             raise ExternalServiceError(result)
         else:
             return render_template('register/register.almost-done.html', _theme='default', email=email_address)
-        # TODO We need to add an exception timeout catch and handle this type of error
```
tej99: I think calls to external dependencies should be wrapped with exception timeout catch. Could put it in this card or leave the todo.

andrewmil: Yeah I agree, I think what we should do going forward is build up a requests library for calling each service so we can deal with this kind of exception handling away from the main logic of the code.

That's probably a task to do going through the whole app rather than one bit at a time though as there's so many requests in frontstage

PR #208, opened 2017-09-21
Fix/refactoring secure message blueprint

andrewmil:
Refactoring of the structure of the secure-messaging blueprint and templates in frontstage

Previously there was a very large amount of duplicated code in this blueprint and was hard to follow. This PR tries to reuse more of the code and make it easier to understand.

There are now only 3 endpoints in the blueprint rather than the previous 6
- `/create-message` which deals with all calls to create/modify a message in the secure message service (both drafts and messages)
- `messages_get/<label>` which shows a list of messages with the corresponding tag ('INBOX, 'SENT' or 'DRAFT')
- `message_get/<label>/<message_id>` Allows the user to view a message, this works for all message type labels

The message templates have also been refactored.
No changes should have been made visually but now the `secure-messages-view.html` is used to generate the html for viewing any type of message and also for the old create-message view.


andrewmil:
Good shout I will do that, I want to do more testing as well and compare the visuals with what we currently have, Don't want to have accidentally introduced bugs

frontstage/views/secure_messaging.py
```diff
@@ -39,16 +125,18 @@ def get_party_ru_id(party_id):
...
+        logger.error('Failed to retrieve collection case id')
         raise ExternalServiceError(collection_response)
     collection_response_json = collection_response.json()
-    return collection_response_json[0].get('id')
+    case_id = collection_response_json[0].get('id')
+    logger.debug('Successfully received collection case', case_id=case_id)
+    return case_id
 
 
 def get_survey_id(party_id):
```
tej99: This function's request handling needs a second look.

frontstage/views/secure_messaging.py
```diff
@@ -17,6 +17,91 @@
...
+    """Gets users messages"""
+    party_id = session['party_id']
+    loggerb = logger.bind(party_id=party_id, label=label)
+    messages = get_messages(label, loggerb)
+    unread_msg_total = get_unread_message_total(loggerb)
+    return render_template('secure-messages/secure-messages.html', _theme='default', messages=messages['messages'],
+                           links=messages['_links'], label=label, total=unread_msg_total['total'])
+
+
 def get_party_ru_id(party_id):
```
tej99: Could you double check the log levels in this function. Don't think it should be an error for a 404

frontstage/views/secure_messaging.py
```diff
@@ -39,16 +125,18 @@ def get_party_ru_id(party_id):
 
 def get_collection_case(party_id):
     url = app.config['RM_CASE_GET_BY_PARTY'].format(app.config['RM_CASE_SERVICE'], party_id)
-    logger.debug('Retrieving collection case id', party_id=party_id)
+    logger.debug('Retrieving collection case id')
     collection_response = requests.get(url, auth=app.config['BASIC_AUTH'])
     if collection_response.status_code == 204:
-        logger.error('No case found', party_id=party_id)
+        logger.error('No case found')
```
tej99: Shouldn't be an error if the service has handled the request and given a 204

frontstage/views/secure_messaging.py
andrewmil: Are we sure that this shouldnt be an error? If the user has an account and is sending a secure message they should have all of these things, if they don't something has gone wrong

andrewmil:
I've changed those error log messages to warning messages. It shouldn't be possible for a user to be logged in and not have these fields

PR #211, opened 2017-09-25
Email address exposed in some urls and logs

andrewmil:
Now sending encoded email token in get_respondent_by_email party endpoint

Added Encoder class to handle encoding

This change works in conjunction with https://github.com/ONSdigital/ras-party/pull/60

https://trello.com/c/dUefc87O/667-email-address-exposed-in-some-urls-and-logs


config.py
```diff
@@ -130,6 +131,7 @@ class DevelopmentConfig(Config):
     DEBUG = True
     TEMPLATES_AUTO_RELOAD = True
     SECRET_KEY = os.getenv('SECRET_KEY', 'ONS_DUMMY_KEY')
```
LJBabbage: The default in party for SECRET_KEY is different (aardvark) and as we are now using it, it will fall over. This is kind of irrelevant as we inject SECRET_KEY into all the important environments, but it could trip up someone running it locally so worth changing

LJBabbage:
I will raise a card for defaults between environments and get these merged in once Travis is done

jonmorphic:
That’s not the same thing. It’s SECURITY_USER_NAME/SECURITY_USER_PASSWORD that need to be aligned.

config.py
rjweeks: Should we delete the default and issues errors for unset environment variables, like you did in Secure Message?

rjweeks: Delete default for SECRET_KEY?

PR #212, opened 2017-09-25
Add ons email and remove help with email

tej99:
Incorrect message when the trying to login without verifying the account.

- Added ons email
- Remove 'help with email' section

https://trello.com/c/kERbocRp/675-defect-47-us007-found-in-release-80-sit-incorrect-message-when-the-trying-to-login-without-verifying-the-account

### DIFF
--- a/frontstage/templates/sign-in/sign-in.account-not-verified.html
+++ b/frontstage/templates/sign-in/sign-in.account-not-verified.html
@@ -7,8 +7,6 @@
     <h1 class="saturn">Trouble signing in?</h1>
 
     <p>Please follow the link in the verification email we sent you.</p>
-    <p>If you haven't received an email please call us on 0300 1234 932</p>
-
-    {% include "register/register.help-with-email.html" %}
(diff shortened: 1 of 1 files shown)


frontstage/templates/sign-in/sign-in.account-not-verified.html
```diff
@@ -7,8 +7,6 @@
     <h1 class="saturn">Trouble signing in?</h1>
 
     <p>Please follow the link in the verification email we sent you.</p>
-    <p>If you haven't received an email please call us on 0300 1234 932</p>
-
-    {% include "register/register.help-with-email.html" %}
+    <p>If you haven't received an email please call us on 0300 1234 932 or email <a href="mailto:surveys@ons.gov.uk">surveys@ons.gov.uk</a></p>
```
andrewmil: Not required but you could add a link to the phone number too with `<a href="tel:03001234931">0300 1234 932</a>`.

It's inconsistent across frontstage whether we include that or not though

tej99: Yeah Nic has commented the changes on the trello card. As for the tel link i'm going to leave it out as the other template trouble signing in page doesn't do it.

PR #213, opened 2017-09-26
Adding help to header

nicprice:
- Put html code in the header partial
- Put new help.js in with the scripts
**- Only hiding help info and showing button if javascript is running**
- Would be better to remove initial transition when hiding help info on page load
- Still need to improve treatment for small screen esp. when signed in


nicprice:
Thanks @andrewmil - good spot! I've fixed this now and pushed up

nicprice:
Is it caching? The "Sign out" link shouldn't be there unless you're signed in.

nicprice:
@andrewmil @saedwards - something's not right with the transition, so closing PR and will need further investigation

nicprice:
Have left branch with latest changes for someone to pick up, or I can pick up next Tuesday

PR #218, opened 2017-10-03
[no subject] should show if no subject

tej99:
Currently no link is shown for a message in the message list view if there is no subject.
Messages/Drafts with no subject should show up as '[no subject]' in message list view.

trello card: https://trello.com/c/ICXHAdV6/676-secure-messages-sent-with-only-spaces-in-subject-cannot-be-viewed-in-rops


.gitignore
```diff
@@ -94,4 +94,4 @@ ENV/
 
 # Rope project settings
 .ropeproject
-.build
+.build
```
andrewmil: missing line at end of file

frontstage/views/secure_messaging.py
```diff
@@ -79,7 +79,7 @@ def message_get(session, label, message_id):
     else:
         draft = {}
 
-    if label == 'INBOX':
+    if label == 'UNREAD':
```
andrewmil: good spot this will save a lot of extra calls

frontstage/filters/subject_filter.py
```diff
@@ -0,0 +1,8 @@
+def subject_filter(subject):
+    """
+    Filter to format subject if subject is empty or has only spaces.
+    """
+    if subject is None or len(subject.strip()) < 1:
+        subject = '[no subject]'
```
andrewmil: We're actually not testing this functionality, might be an idea to add that test in

PR #220, opened 2017-10-05
Adding help to header

nicprice:
I've added the "Help" button and information to the header.

On large and medium screen the help information is hidden, and only revealed when the Help button is activated.

On small screen we show the help information all the time, and never show the Help button.

Screenshots for "Not signed in" and "Signed in" versions:

header-help-large-not-signed-in-closed
header-help-large-not-signed-in-open
header-help-large-signed-in-closed
header-help-large-signed-in-open
header-help-small-not-signed-in
header-help-small-signed-in


nicprice:
Ideally it would be better not to have the transition when the page loads.

andrewmil:
Changes made which should remove the initial transition

PR #224, opened 2017-10-12
Removed fieldset tag from secure message form

nicprice:
No visual changes, but html no longer has <fieldset> tag in it wrapped around the form input fields because not necessary (and previous version had fieldset with no legend which failed accessibility).

### DIFF
--- a/frontstage/templates/secure-messages/message-form.html
+++ b/frontstage/templates/secure-messages/message-form.html
@@ -27,7 +27,6 @@
             <div class="panel panel--simple panel--error panel--spacious" id="subject-error">
                 <p class="error-message">{{ errors['subject'][0] }}</p>
         {% endif %}
-        <fieldset>
             <label class="venus" for="secure-message-subject">Subject</label>
             <input
                 id="secure-message-subject"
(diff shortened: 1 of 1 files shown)


frontstage/templates/secure-messages/message-form.html
```diff
@@ -47,20 +45,18 @@
 
     {% if "body" in errors %}
         <div class="panel panel--simple panel--error panel--spacious" id="body-error">
-            <p class="error-message">{{ errors['body'][0] }}</p>
+          <p class="error-message">{{ errors['body'][0] }}</p>
```
andrewmil: line should be 2 spaces further in

frontstage/templates/secure-messages/message-form.html
```diff
@@ -47,20 +45,18 @@
...
-            {% else %}
-            <label class="venus" for="secure-message-body">Message</label>
-            {% endif %}
-            <textarea
-                id="secure-message-body"
-                class="input input--textarea input--textarea-message"
-                rows="10"
-                name="secure-message-body">{{ draft['body'] }}</textarea>
-        </fieldset>
+          {% if message %}
```
andrewmil: This chunk that was previously inside the fieldset tag can be unindented now

PR #225, opened 2017-10-20
Tbl087/moving secure message requests to api

andrewmil:
- Added an `api_call` function which will handle all calls to frontstage-api
- Redirected the requests for the `get secure messages` tabs through the frontstage-api
- Amended tests to cover the new functionality

These changes require the fronstage-api to connect to


config.py
```diff
@@ -108,6 +108,15 @@ class Config(object):
     LABELS_GET_API_URL = RAS_SECURE_MESSAGE_SERVICE + 'labels?name=unread'
     THREAD_GET_API_URL = RAS_SECURE_MESSAGE_SERVICE + 'thread/{}'
 
+    RAS_FRONTSTAGE_API_HOST = os.getenv('RAS_FRONTSTAGE_API_HOST', 'localhost')
+    RAS_FRONTSTAGE_API_PORT = os.getenv('RAS_FRONTSTAGE_API_PORT', 8082)
+    RAS_FRONTSTAGE_API_PROTOCOL = os.getenv('RAS_FRONTSTAGE_API_PROTOCOL', 'http')
+    RAS_FRONTSTAGE_API_SERVICE = '{}://{}:{}/'.format(RAS_FRONTSTAGE_API_PROTOCOL,
+                                                      RAS_FRONTSTAGE_API_HOST,
+                                                      RAS_FRONTSTAGE_API_PORT)
+    GET_MESSAGES_URL = 'messages_list'
```
tej99: MESSAGES_API_URL and MESSAGE_LIMIT now redundant

frontstage/views/secure_messaging.py
```diff
@@ -96,10 +97,10 @@ def messages_get(session, label="INBOX"):
     """Gets users messages"""
     party_id = session['party_id']
     loggerb = logger.bind(party_id=party_id, label=label)
-    messages = get_messages(label, loggerb)
-    unread_msg_total = get_unread_message_total(loggerb)
+    messages = get_messages_list(label, loggerb)
+    unread_msg_total = messages.get('total')
```
tej99: Possibly rename 'total' to 'total_unread' in the frontage_api as the context has now changed from only unread labels to message list. 
Or rename the response in SM but that would be a breaking change if other apis are using the labels endpoint.
Or we can add that attribute in the message list in SM message list response instead of having to call another endpoint.

tests/test_data/cases_todo_in_progress.json
```diff
@@ -0,0 +1,48 @@
+[
```
tej99: did you mean to add this?

andrewmil: As in the bracket or the file?

tej99: The file, doesn't look like it's to do with this refactor and doesn't seem like it's being used either.

frontstage/views/secure_messaging.py
andrewmil: I agree with that change, we can make it in the next update we make to the frontstage-api

tests/test_data/cases_todo_in_progress.json
andrewmil: Yeah weird I feel like I remember using this at some point but apparently not, I'll remove it

config.py
andrewmil: Removed

frontstage/common/api_call.py
```diff
@@ -0,0 +1,39 @@
...
+from structlog import wrap_logger
+
+from frontstage import app
+from frontstage.exceptions.exceptions import FailedRequest, InvalidRequestMethod
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def api_call(method, endpoint, parameters=None, json=None, headers=None):
```
benjefferies: Is there a need to handle anything other than a 'get' method as it's currently the only usage?

frontstage/common/api_call.py
```diff
@@ -0,0 +1,39 @@
...
+        elif method == 'POST':
+            response = requests.post(url, headers=headers, json=json)
+        elif method == 'PUT':
+            response = requests.put(url, headers=headers, json=json)
+        else:
+            logger.error('Invalid request method', method=str(method), url=url)
+            raise InvalidRequestMethod(method, url)
+    except ConnectTimeout as e:
+        logger.error('Connection to remote server timed out', method=method, url=url, exception=str(e))
+        raise FailedRequest(method, url, e)
```
benjefferies: Is there much value in wrapping and raising another type of exception?

frontstage/common/api_call.py
```diff
@@ -0,0 +1,39 @@
...
+
+    except ConnectionError as e:
+        logger.error('Failed to connect to external service', method=method, url=url, exception=str(e))
+        raise FailedRequest(method, url, e)
```
benjefferies: Is there much value in wrapping and raising another type of exception?

frontstage/common/api_call.py
andrewmil: This method is going to handle all requests from frontstage in the near future, I added these in so it's ready for these in advance.

frontstage/common/api_call.py
andrewmil: I think we'll want to catch any failed calls to the API explicitly so we can flag it up somewhere (Splunk etc). All the calls frontstage makes will be going through it so it'll be important to know when there's failures there.
I could make another `APIFailure` exception maybe which could handle it?

andrewmil: I could add an error handler for it which returns the general error page and logs a message we can look for in splunk?

benjefferies: I was thinking we'd could just catch ConnectTimeout and ConnectionError in the error handler

frontstage/views/secure_messaging.py
```diff
@@ -160,32 +161,20 @@ def get_survey_id(party_id):
     return survey_name
 
 
-def get_messages(label, logger):
-    headers = {"Authorization": request.cookies['authorization']}
+def get_messages_list(label, logger):
```
benjefferies: I know you didn't add it but why are we passing in the logger?

config.py
```diff
@@ -108,6 +106,15 @@ class Config(object):
...
     THREAD_GET_API_URL = RAS_SECURE_MESSAGE_SERVICE + 'thread/{}'
 
+    RAS_FRONTSTAGE_API_HOST = os.getenv('RAS_FRONTSTAGE_API_HOST', 'localhost')
+    RAS_FRONTSTAGE_API_PORT = os.getenv('RAS_FRONTSTAGE_API_PORT', 8083)
+    RAS_FRONTSTAGE_API_PROTOCOL = os.getenv('RAS_FRONTSTAGE_API_PROTOCOL', 'http')
+    RAS_FRONTSTAGE_API_SERVICE = '{}://{}:{}/'.format(RAS_FRONTSTAGE_API_PROTOCOL,
+                                                      RAS_FRONTSTAGE_API_HOST,
+                                                      RAS_FRONTSTAGE_API_PORT)
+    GET_MESSAGES_URL = 'messages_list'
+    UNREAD_MESSAGES_TOTAL_URL = 'unread_message_total'
```
benjefferies: Is this used?

frontstage/views/secure_messaging.py
```diff
@@ -160,32 +161,20 @@ def get_survey_id(party_id):
...
     logger.debug('Attempting to retrieve messages')
-    url = app.config['MESSAGES_API_URL']
-    if label is not None:
-        url = url + "&label=" + label
-    response = requests.get(url, headers=headers)
+    headers = {"Authorization": request.cookies['authorization']}
```
benjefferies: Is this something that could be pushed into `api_call` as it seems like a generic header that will (probably) always need to be present?

config.py
andrewmil: Nope this got absorbed into the frontstage api so I'll remove

frontstage/views/secure_messaging.py
andrewmil: This is only sent in the secure messages requests so isn't a generic to all requests. The other services are using BASIC_AUTH currently, would be good to have this standardised so we can make it generic though

frontstage/views/secure_messaging.py
andrewmil: Rather than defining `party_id`, `message_id` etc in every relevant log line they get bound to a logger with `logger.bind`, e.g. on line 26. From that point onwards everywhere that logger is called with include `party_id`. 
Once I bind anything to a logger I pass it through the other functions so it retains these details without having to explicitly write them out on every log line

benjefferies: I think you get that behaviour without assigning to logger and passing it through

andrewmil: I'll test it out now

andrewmil: Just had a play around with it and I couldn't recreate the behaviour without doing it the way it's currently being implemented. If we wanting to change this should we raise it as another ticket rather than doing it as part of this request?

benjefferies: Cool it's strange I was sure it did when I did this on EQ (https://github.com/ONSdigital/eq-survey-runner/blob/master/app/views/questionnaire.py#L49) but if it's not a quick one lets investigate on another ticket

andrewmil: That is a bit strange, we do define the logger slightly differently so that may be it. 
`logger = get_logger()` vs `logger = wrap_logger(logging.getLogger(__name__))`
Either way there's quite a few places we'd have to change it so probably should do on another card.

benjefferies:
It would be good to squash the commits @andrewmil

PR #226, opened 2017-10-23
Add default manifest.yml

benjefferies:
Add manifest file for cloudfoundry deployments

1. Are the defaults set to sensible values?
1. Are we missing any manifest properties that should be there?
1. Could we strip them down further?


benjefferies:
This will bring consistency and reproducibility to deployments when they're not deployed via jenkins. All values are defined in the jenkins plugin are applied over the manifest. I don't believe the manifest is even considered with the settings we have. 

I have tested this by upping the memory in dev and it did apply it although I had to delete and deploy the app for those changes to be applied

pricem14pc:
That's correct. The RAS Jenkins jobs are setup not to use manifests. The environment variables would also need to be provisioned via the manifest. The RM jobs use the following manifest (e.g.)..

https://github.com/ONSdigital/rm-case-service/blob/master/manifest-template.yml

And a Jenkins bash script to replace the environment specific values. However, these values (inc. secrets) are in plain text in the Jenkins configs...

sed -i -- 's/SPACE/prod/g' *template.yml
sed -i -- 's/INSTANCES/2/g' *template.yml
sed -i -- 's/DATABASE/<<DATABASE>>b/g' *template.yml
sed -i -- 's/DOMAIN/<<DOMAIN>>/g' *template.yml
sed -i -- 's/REPLACE_PORT/443/g' *template.yml
sed -i -- 's/REPLACE_PROTOCOL/https/g' *template.yml
sed -i -- 's/ENDPOINT_ENABLED/\"false\"/g' *template.yml
sed -i -- 's/true/\"true\"/g' *template.yml
sed -i -- 's/REPLACE_BA_USERNAME/<<USERNAME>>/g' *template.yml
sed -i -- 's/REPLACE_BA_PASSWORD/<<PASSWORD>>g' *template.yml
echo "    logging_level_uk.gov.ons.ctp: WARN" >> manifest-template.yml
echo "    reportSettings_cronExpression: \"0 0 23 * * *\"" >> manifest-template.yml

benjefferies:
I guess it would only make sense to add environmental variables that don't change per environment? and they if they don't change per environment it would be nicer to remove them as environmental variables?

PR #227, opened 2017-10-25
Tbl087/moving more secure message requests to api

andrewmil:
- All requests used in the secure messaging blueprint are now made through the frontstage api
- In cases where the API returns an error, the code it returns is used to decided what to do next
- When an error is returned from the API which cannot be recovered from the exception ApiError is raised which returns the 'Server error' page

This PR works in conjunction with changes in the frontstage api at https://github.com/ONSdigital/ras-frontstage-api/pull/6


frontstage/views/secure_messaging.py
```diff
@@ -95,195 +48,108 @@ def message_get(session, label, message_id):
...
+        raise ApiError('FA000')
 
     messages_list = json.loads(response.text)
+
+    # Handle Api Error codes
+    error_code = messages_list.get('error', {}).get('code')
+    if error_code == 'FA001':
+        logger.error('Failed to retrieve messages list')
+        raise ApiError('FA001')
+    elif error_code == 'FA002':
```
tej99: I think... {"unread_messages_total": { "error": { "code": "FA002",}}} is returned so error_code would never be 'FA002'

andrewmil: You're right, I've made the changes you suggested in frontstage api and it fixed this mistake I'm pretty sure

PR #230, opened 2017-10-31
Defect 60 - Check oauth before reset password

tej99:
Check if account is locked in oauth and serve a 'something went wrong' page if it is.

https://trello.com/c/8DJShebh/658-defect-60-us012-found-in-release10-app-is-allowing-the-respondent-to-reset-a-password-when-the-account-is-locked


frontstage/templates/passwords/reset-password.trouble.html
```diff
@@ -0,0 +1,17 @@
+{% import 'partials/section.html' as section %}
```
andrewmil: Is this template from some design document?

frontstage/views/passwords.py
```diff
@@ -24,6 +25,46 @@ def forgot_password():
         email_address = request.form.get('email_address')
         post_data = {"email_address": email_address}
 
+        try:
```
andrewmil: There's probably a bunch of reused code between this and the the sign-in page. Could look at refactoring it together in the move to frontstage-api

frontstage/views/passwords.py
```diff
@@ -24,6 +25,46 @@ def forgot_password():
...
+
+            oauth2_response = requests.post(url=token_url, data=data,
+                                            headers=headers, auth=(app.config['RAS_FRONTSTAGE_CLIENT_ID'],
+                                                                   app.config['RAS_FRONTSTAGE_CLIENT_SECRET']))
+            # Check to see that this user account is not locked
+            if oauth2_response.status_code == 401:
+                oauth2Error = json.loads(oauth2_response.text)
+                if 'Unauthorized user credentials' in oauth2Error['detail']:
+                    logger.warning('Email address is not registered')
+                    template_data = {"error": {"type": {"Email address is not registered"}}}
```
andrewmil: This template data isn't used anywhere currently, this is an existing problem with the original form however rather than something that's been added here I think.

frontstage/views/passwords.py
```diff
@@ -24,6 +25,46 @@ def forgot_password():
...
+                    logger.error('OAuth 2 server generated 401 which is not understood',
+                                 oauth2error=oauth2Error.get('detail'))
+
+                return render_template('passwords/reset-password.trouble.html', _theme='default',
+                                       data={"error": {"type": "failed"}})
+            if oauth2_response.status_code != 201:
+                logger.error('Unknown error from the OAuth2 server')
+                raise ExternalServiceError(oauth2_response)
+        except requests.ConnectionError as e:
+            logger.warning('Connection error between the server and the OAuth2 service of: {}'.format(exception=str(e)))
```
andrewmil: Could of untested lines here and at line 56

frontstage/views/passwords.py
tej99: Can do the refactoring as part of this ticket if it makes the switch to the frontage-api easier?

andrewmil: I'm porting the sign-in stuff over already right now so probably easier if I do it :+1:

frontstage/views/passwords.py
benjefferies: We could catch requests.ConnectionError in the error handlers. It exposes the url it failed to connect to

frontstage/views/passwords.py
```diff
@@ -24,6 +25,46 @@ def forgot_password():
...
+
+            data = {
+                'grant_type': 'reset_password',
+                'client_id': app.config['RAS_FRONTSTAGE_CLIENT_ID'],
+                'client_secret': app.config['RAS_FRONTSTAGE_CLIENT_SECRET'],
+                'username': email_address,
+            }
+            headers = {
+                'Accept': 'application/json',
+                'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
```
benjefferies: Should content-type be `application/json`?

frontstage/views/passwords.py
```diff
@@ -24,6 +25,46 @@ def forgot_password():
...
+                                                                   app.config['RAS_FRONTSTAGE_CLIENT_SECRET']))
+            # Check to see that this user account is not locked
+            if oauth2_response.status_code == 401:
+                oauth2Error = json.loads(oauth2_response.text)
+                if 'Unauthorized user credentials' in oauth2Error['detail']:
+                    logger.warning('Email address is not registered')
+                    template_data = {"error": {"type": {"Email address is not registered"}}}
+                    return render_template('passwords/forgot-password.html', _theme='default', form=form,
+                                           data=template_data)
+                elif 'User account locked' in oauth2Error['detail']:
```
benjefferies: Maybe http 403 would be good enough for a locked account?

frontstage/views/passwords.py
andrewmil: I haven't tested it but if I remember right the Django server is a bit picky about what you put as content type and this was the only way we could get it working.
I'm working on a similar call in frontstage-api so I can test

tej99: No the Django oAuth server is expecting an encoded payload, breaks if i set content-type to application/json

frontstage/views/passwords.py
tej99: I agree but changing Django oAuth to return 403 for this case will make responses inconsistent with the current logic and trying to fix the inconsistencies would introduce breaking changes.

PR #231, opened 2017-10-31
TBL087/Moving Sign-In calls to Frontstage-Api

andrewmil:
This PR works in conjunction with this frontstage-api PR https://github.com/ONSdigital/ras-frontstage-api/pull/8

- Moved sign in calls to single frontstage-api call which returns oauth2 token details and party_id
- Updated testing

### DIFF
--- a/frontstage/views/secure_messaging.py
+++ b/frontstage/views/secure_messaging.py
@@ -111,14 +111,12 @@ def send_message(party_id, is_draft):
     # If form errors are returned render them
     if response.status_code == 400:
         logger.debug('Form submitted with errors', party_id=party_id)
-        error_message = json.loads(response.text)
-        message = error_message.get('error', {}).get('data', {}).get('thread_message')
-        errors = error_message['error']['data']['form_errors']
+        error_message = json.loads(response.text).get('error', {}).get('data')
         return render_template('secure-messages/secure-messages-view.html',
                                _theme='default',
-                               message=message,
+                               message=error_message.get('thread_message'),
                                draft=message_json,
-                               errors=errors)
+                               errors=error_message.get('form_errors'))
 
     if response.status_code != 200:
         logger.debug('Failed to send message')
--- a/frontstage/views/secure_messaging.py
+++ b/frontstage/views/secure_messaging.py
@@ -108,23 +108,23 @@ def send_message(party_id, is_draft):
         message_json["msg_id"] = request.form['msg_id']
     response = api_call('POST', endpoint, parameters={"is_draft": is_draft}, json=message_json, headers=headers)
 
-    if response.status_code != 200:
-        logger.debug('Failed to send message')
-        raise ApiError(response)
-
-    sent_message = json.loads(response.text)
-
     # If form errors are returned render them
-    if sent_message.get('error'):
+    if response.status_code == 400:
         logger.debug('Form submitted with errors', party_id=party_id)
-        message = sent_message.get('error', {}).get('data', {}).get('thread_message')
-        errors = sent_message['error']['data']['form_errors']
+        error_message = json.loads(response.text)
+        message = error_message.get('error', {}).get('data', {}).get('thread_message')
+        errors = error_message['error']['data']['form_errors']
         return render_template('secure-messages/secure-messages-view.html',
                                _theme='default',
                                message=message,
                                draft=message_json,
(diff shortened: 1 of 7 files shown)


frontstage/views/sign_in.py
```diff
@@ -31,70 +31,38 @@ def login():
...
+        sign_in_data = {
+            "username": username,
+            "password": password
+        }
+        response = api_call('POST', app.config['SIGN_IN_URL'], json=sign_in_data)
+
+        # Handle OAuth2 authentication errors
+        if response.status_code == 502:
+            error_json = json.loads(response.text).get('error')
+            if error_json.get('status_code') == 401:
```
benjefferies: Why isn't the `response.status_code == 401`?

frontstage/views/sign_in.py
```diff
@@ -31,70 +31,38 @@ def login():
...
                     return render_template('sign-in/sign-in.html', _theme='default', form=form, data={"error": {"type": "failed"}})
-                elif 'User account locked' in oauth2Error['detail']:
-                    logger.warning('User account is locked on the OAuth2 server')
-                    return render_template('sign-in/sign-in.trouble.html', _theme='default', form=form,
-                                           data={"error": {"type": "account locked"}})
-                elif 'User account not verified' in oauth2Error['detail']:
-                    logger.warning('User account is not verified on the OAuth2 server')
-                    return render_template('sign-in/sign-in.account-not-verified.html', _theme='default', form=form,
-                                           data={"error": {"type": "account not verified"}})
+                elif 'User account locked' in error_message:
```
benjefferies: Could this be http 403?

frontstage/views/sign_in.py
andrewmil: 502 is the correct status code here I think, Bad Gateway indicating an upstream service has returned an invalid response? I can then pass info about that failure in the returned json

andrewmil: 401 will be used for when Basic Auth etc to the frontstage-api fails

frontstage/views/sign_in.py
andrewmil: We could make this return 403 from the django server? Then pass that through the api

frontstage/views/sign_in.py
benjefferies: I'd disagree. We'd be confusing a microservice with proxy server in that case. 

However orchestrating these changes to return a 401 or a 403 might not be very easy since the pattern has already been introducedl

andrewmil: It's not loads of hassle if it does have to be done. This pattern does make things quite clean for the most part other than the cases where we have to return more specific error messages from the origin api's like this one

andrewmil: yeah looking through it would be simple to change the api to pass through the status code from the failed service but this would be the 3rd time I've flipped it one way or the other 👀

frontstage/views/sign_in.py
tej99: Changing this now will introduce breaking changes as there are at least three separate services using the django server

benjefferies: Fair point

benjefferies: Yeah I don't think we should change it in the django server but as we're introducing the frontstage-api we can do the horrible mapping there and have a clean API which returns the relevant status codes

andrewmil:
Those comments by @benjefferies should be dealt with now

frontstage/views/secure_messaging.py
```diff
@@ -108,23 +108,23 @@ def send_message(party_id, is_draft):
...
-
     # If form errors are returned render them
-    if sent_message.get('error'):
+    if response.status_code == 400:
         logger.debug('Form submitted with errors', party_id=party_id)
-        message = sent_message.get('error', {}).get('data', {}).get('thread_message')
-        errors = sent_message['error']['data']['form_errors']
+        error_message = json.loads(response.text)
+        message = error_message.get('error', {}).get('data', {}).get('thread_message')
+        errors = error_message['error']['data']['form_errors']
```
tej99: error_message.get('error', {}).get('data', {}).get('form_errors')? Makes it consistent and robust

andrewmil: Went one better and we able to remove those lines

PR #232, opened 2017-10-31
Tbl087/adding basic auth to frontstage api calls

andrewmil:
Adding basic auth to frontstage-api requests

Works in conjunction with https://github.com/ONSdigital/ras-frontstage-api/pull/9

### DIFF
--- a/frontstage/common/api_call.py
+++ b/frontstage/common/api_call.py
@@ -19,11 +19,11 @@ def api_call(method, endpoint, parameters=None, json=None, headers=None):
             url = url + parameter_string
 
     if method == 'GET':
-        response = requests.get(url, headers=headers)
+        response = requests.get(url, headers=headers, auth=app.config['BASIC_AUTH'])
     elif method == 'POST':
-        response = requests.post(url, headers=headers, json=json)
+        response = requests.post(url, headers=headers, json=json, auth=app.config['BASIC_AUTH'])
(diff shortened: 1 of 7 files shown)


frontstage/common/api_call.py
```diff
@@ -19,11 +19,11 @@ def api_call(method, endpoint, parameters=None, json=None, headers=None):
...
 
     if method == 'GET':
-        response = requests.get(url, headers=headers)
+        response = requests.get(url, headers=headers, auth=app.config['BASIC_AUTH'])
     elif method == 'POST':
-        response = requests.post(url, headers=headers, json=json)
+        response = requests.post(url, headers=headers, json=json, auth=app.config['BASIC_AUTH'])
     elif method == 'PUT':
-        response = requests.put(url, headers=headers, json=json)
+        response = requests.put(url, headers=headers, json=json, auth=app.config['BASIC_AUTH'])
```
jwgwalton: Need to update TestApiCall to include basic auth. The mocking in the current tests is invalid as it is returning a 200 even without basic auth.

andrewmil: I'm not sure what you mean? we don't pass in an auth parameter it's just pulled from config so do we need to change anything about the way it's called in tests?

jwgwalton: I think I'm going mad, you're right.

PR #234, opened 2017-11-03
Tbl087/moving enrolment calls to api

andrewmil:
This PR removes the calls frontstage was making to external services during the enrolment process and replaces them with single calls to the frontstage-api

This PR works in conjunction with this frontstage-api PR https://github.com/ONSdigital/ras-frontstage-api/pull/11

### How to test
I have enrolment codes now so if you needs to test against the int environment just give me a message and I'll send one


frontstage/common/api_call.py
```diff
@@ -18,14 +18,16 @@ def api_call(method, endpoint, parameters=None, json=None, headers=None):
...
     elif method == 'POST':
         response = requests.post(url, headers=headers, json=json, auth=app.config['BASIC_AUTH'])
     elif method == 'PUT':
         response = requests.put(url, headers=headers, json=json, auth=app.config['BASIC_AUTH'])
     else:
-        logger.error('Invalid request method', method=str(method), url=url)
+        logger.error('Invalid request method', method=method, url=url)
         raise InvalidRequestMethod(method, url)
 
+    logger.debug('Frontstage-api response', method=method, url=url, status_code=response.status_code)
```
davidmort: the sdc [logging standards](https://digitaleq.atlassian.net/wiki/spaces/SDC/pages/69632229/Logging) suggest that you should log out "status" instead of "status_code" when logging http requests to reduce verbosity

davidmort: the sdc [logging standards](https://digitaleq.atlassian.net/wiki/spaces/SDC/pages/69632229/Logging) suggest that you should log out "status" instead of "status_code" when logging http requests to reduce verbosity

andrewmil: Will change, I think we log out `status_code` in quite a lot of places across RAS so we might have to change in a few places

PR #235, opened 2017-11-03
Tbl087/moving passwords to frontstage api calls

tej99:
This PR removes the calls frontstage was making to external services during the reset password process and replaces them with calls to the frontstage-api

This PR works in conjunction with this frontstage-api PR ONSdigital/ras-frontstage-api#12


frontstage/views/passwords.py
```diff
@@ -23,56 +22,35 @@ def forgot_password():
...
-        # Check to see that this user account is not locked
-        if oauth2_response.status_code == 401:
-            oauth2Error = json.loads(oauth2_response.text)
-            if 'Unauthorized user credentials' in oauth2Error['detail']:
-                logger.warning('Email address is not registered')
+        if response.status_code == 401:
+            error_json = json.loads(response.text).get('error')
+            error_message = error_json.get('data', {}).get('detail')
+            if 'Unauthorized user credentials' in error_message:
+                logger.info('Requesting password change for unregistered email on OAuth2 server')
```
benjefferies: Instead of having an explicit line of log we could just log the error_message. Then you'd have something like:
```python
if response.status_code == 401:
    error_json = json.loads(response.text).get('error')
    error_message = error_json.get('data', {}).get('detail')
    logger.info(error_message=error_message)
    if 'Unauthorized user credentials' in error_message:
        template_data = {"error": {"type": {"Email address is not registered"}}}
        return render_template('passwords/forgot-password.html', _theme='default', form=form, data=template_data)

    return render_template('passwords/reset-password.trouble.html', _theme='default',
                           data={"error": {"type": "failed"}})
```

frontstage/views/passwords.py
```diff
@@ -23,56 +22,35 @@ def forgot_password():
...
-            logger.error('Unknown error from the OAuth2 server')
-            raise ExternalServiceError(oauth2_response)
-
-        url = app.config['RAS_PARTY_RESET_PASSWORD_REQUEST'].format(app.config['RAS_PARTY_SERVICE'])
-        response = requests.post(url, auth=app.config['BASIC_AUTH'], json=post_data, verify=False)
-        if response.status_code == 404:
-            logger.warning('Email address is not registered')
+
+        elif response.status_code == 404:
+            logger.error('Requesting password change for email registered on OAuth2 server but not in party service')
```
benjefferies: Should this be error level logging or just info?

frontstage/views/passwords.py
```diff
@@ -23,56 +22,35 @@ def forgot_password():
...
             template_data = {"error": {"type": {"Email address is not registered"}}}
             return render_template('passwords/forgot-password.html', _theme='default', form=form, data=template_data)
+
         if response.status_code != 200:
-            logger.error('Failed to send password change request email')
-            raise ExternalServiceError(response)
+            logger.error('Unable to send password change request')
```
benjefferies: If we logged the stacktrace I guess this line of logging would be unnecessary and would caused duplicate error logging as we will log when the ApiError is raised?

frontstage/views/passwords.py
```diff
@@ -92,32 +70,36 @@ def forgot_password_check_email():
 @passwords_bp.route('/reset-password/<token>', methods=['GET', 'POST'])
```
benjefferies: We should split the GET and POST handlers out into two methods really. Appreciate it wasn't introduced in your changes

frontstage/views/passwords.py
```diff
@@ -92,32 +70,36 @@ def forgot_password_check_email():
...
-        response = requests.put(url, auth=app.config['BASIC_AUTH'], json=put_data, verify=False)
+        url = app.config['CHANGE_PASSWORD']
+        response = api_call('PUT', url, json=put_data)
 
-        if response.status_code != 200:
+        if response.status_code == 200:
+            logger.info('Successfully change user password', token=token)
+            return redirect(url_for('passwords_bp.reset_password_confirmation'))
+        else:
             logger.error('Failed to change user password', token=token)
```
benjefferies: Should this now raise an Error?

frontstage/views/passwords.py
```diff
@@ -92,32 +70,36 @@ def forgot_password_check_email():
...
+        response = api_call('GET', url, parameters=parameters)
+
+    if response.status_code == 409:
+        logger.warning('Token expired', token=token)
+        return render_template('passwords/password-expired.html', _theme='default')
+    elif response.status_code == 404:
+        logger.warning('Invalid token sent to party service', token=token)
+        return redirect(url_for('error_bp.not_found_error_page'))
+    elif response.status_code != 200:
+        logger.error('Party service failed to verify token')
```
benjefferies: If we logged the stacktrace I guess this line of logging would be unnecessary and would caused duplicate error logging as we will log when the ApiError is raised?

frontstage/views/passwords.py
andrewmil: Yeah this looks like a place where we'd raise `ApiError`.
Currently I think it just reloads the page

frontstage/views/passwords.py
tej99: It's logged at error level because Frontstage API will only return 404 if the user exists in the OAuth server but not in the party service. This should never happen, if it does though something has gone wrong with the user's account and should be investigated.

frontstage/views/passwords.py
tej99: If the form is valid then api call is made to CHANGE_PASSWORD endpoint else the VERIFY_PASSWORD_TOKEN endpoint is called. Other than the 200 response from the CHANGE_PASSWORD call all other responses are handled after to avoid duplication.

frontstage/views/passwords.py
tej99: This is being done throughout front stage I'll raise a ticket to make these changes throughout to keep the service consistent.

frontstage/views/passwords.py
```diff
@@ -83,17 +70,20 @@ def forgot_password():
...
     form = ResetPasswordForm(request.form)
-    url = app.config['RAS_PARTY_VERIFY_PASSWORD_TOKEN'].format(app.config['RAS_PARTY_SERVICE'], token)
-    response = requests.get(url, auth=app.config['BASIC_AUTH'], verify=False)
+
+    url = app.config['VERIFY_PASSWORD_TOKEN']
+    parameters = {"token": token}
+    response = api_call('GET', url, parameters=parameters)
+
     if response.status_code == 409:
         logger.warning('Token expired', token=token)
```
andrewmil: These error lines aren't covered by tests atm

PR #236, opened 2017-11-13
Tbl087/moving surveys calls to api

andrewmil:
All calls made in the surveys blueprint have been moved into the Frontstage-API. In the blueprint these have been replaced with a single call to the api for each endpoint.

### To test
Run Frontstage-API from this branch https://github.com/ONSdigital/ras-frontstage-api/pull/15 alongside this PR. 
Test loading up the surveys list, the access survey page, downloading collection instruments and uploading collection instruments.


tests/test_data/surveys_list.json
```diff
@@ -0,0 +1,79 @@
...
+    "survey": {
+      "id": "cb0711c3-0ac8-41d3-ae0e-567e5ea1ef87",
+      "shortName": "BRES",
+      "longName": "Business Register and Employment Survey",
+      "surveyRef": "221"
+    },
+    "collection_instrument_size": 5302,
+    "status": "None"
+  }
+]
```
gemmai95: No newline at EOF

frontstage/views/surveys.py
```diff
@@ -1,181 +1,159 @@
...
-    return render_template('surveys/surveys-history.html',  _theme='default', data_array=data_array, history=True)
+    party_id = session['party_id']
+    logger.info('Retrieving surveys history list', party_id=party_id)
+    surveys_list = get_surveys_list(party_id, 'history')
+    logger.info('Successfully retrieved surveys history list', party_id=party_id)
+    return render_template('surveys/surveys-history.html',  _theme='default', surveys_list=surveys_list, history=True)
+
+
+def get_surveys_list(party_id, list_type):
+    logger.debug('Retrieving surveys list', party_id=party_id, list_type=list_type)
```
rhiGriff: Should this be the same as above either being an info log or a debug?

frontstage/views/surveys.py
```diff
@@ -1,181 +1,159 @@
...
+    params = {
+        "party_id": party_id,
+        "list": list_type
+    }
+    response = api_call('GET', app.config['SURVEYS_LIST'], parameters=params)
+    if response.status_code != 200:
+        logger.error('Failed to retrieve surveys list', party_id=party_id, list_type=list_type)
+        raise ApiError(response)
+    surveys_list = json.loads(response.text)
+    logger.debug('Successfully retrieved surveys list', party_id=party_id, list_type=list_type)
```
rhiGriff: Same as above.

frontstage/views/surveys.py
andrewmil: My pattern for using debug/info logs is to have info logs in the direct route functions then debug for any functions they call. Not sure if that's the ideal way to go but I think it's consistent across frontstage(-api)

gemmai95: Could it not be more consistent to have it the same across all functions?

andrewmil: We need the important broad strokes logged at info level but the more involved stuff we want in debug so that's what I've tried to do here. Surely not a good idea to only have one or the other?

frontstage/views/surveys.py
gemmai95: Do you need this log line here in debug and above in info?

andrewmil: Actually your right for this function it doesn't make sense, I'll update it

frontstage/views/surveys.py
```diff
@@ -1,181 +1,155 @@
...
 @surveys_bp.route('/upload_survey', methods=['POST'])
 @jwt_authorization(request)
 def upload_survey(session):
-    """Logged in page for users only."""
-    party_id = session.get('party_id', 'no-party-id')
-    case_id = request.args.get('case_id', None)
-    logger.info('Attempting to upload survey', party_id=party_id, case_id=case_id)
-
-    if request.content_length > app.config['MAX_UPLOAD_LENGTH']:
-        error_info = "size"
```
tej99: Did we agree on keeping this check here as well as/instead of in frontsage-api?

andrewmil: I'll add this check back into frontstage now, We probably need a more robust way to be validating the files though

tej99: Yeah for sure

PR #241, opened 2017-11-16
[tbl111] Use pipenv for deterministic builds

jcox-dev:
Pipenv enables us to lock down and guarantee what dependencies we get
when we install dependencies. This works by using the pip hashes when
running 'pipenv install --dev'. The hashes are stored in a generated
file call Pipfile.lock.

We no longer need requirement files instead pipenv uses a file called
Pipfile. The documentation can be found
https://github.com/kennethreitz/pipenv.


andrewmil:
How does pipenv handle python versions? 
If we specify a specific version to use does it look for that on your machine or does it just use your default?

jcox-dev:
@andrewmil I've updated the README to show how pipenv will defer to pyenv if a specified (in the Pipfile) python version is missing.

PR #243, opened 2017-11-20
Use unittest instead of testfixtures

tej99:
Testfixtures package caused deprecation warnings in tests: 
tests/app/test_logger_config.py::TestLoggerConfig::test_indent_type_error
  /Users/patelt/.virtualenvs/ras-frontstage-NfGiVM_Z/lib/python3.5/site-packages/testfixtures/utils.py:36: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() instead
    to_add = len(getargspec(wrapped)[0][len(args):])

We can get the same functionality from the unittest package to test logs.


gemmai95:
Do you guys not use pytest rather than unitest?

tej99:
@gemmai95 nope looks like pytest is only used to run the tests

PR #244, opened 2017-11-20
Add error handler tests

andrewmil:
Added tests for the error handlers file.

Also made a change in the error handlers, no longer catching 500 status codes and instead catching all other exceptions which are not caught specifically.


frontstage/error_handlers.py
```diff
@@ -43,7 +36,14 @@ def connection_error(error):
...
 @app.errorhandler(JWTValidationError)
-def connection_error_jwt_validation(error):  # pylint: disable=unused-argument
+def jwt_validation_error(error):  # pylint: disable=unused-argument
     return redirect(url_for('error_bp.not_logged_in_error_page',
                             _external=True,
                             _scheme=getenv('SCHEME', 'http')))
+
+
+@app.errorhandler(Exception)
+def server_error(error):  # pylint: disable=unused-argument
```
gemmai95: Why is the error not logged out? I noticed this with most of the repos just wondering if there is a reason the error itself isn't logged?

tests/app/test_error_handlers.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertTrue('Server error'.encode() in response.data)
+
+    @requests_mock.mock()
+    def test_jwt_validation_error(self, mock_request):
+        mock_request.post(url_oauth, exc=JWTValidationError)
+
+        response = self.app.post('sign-in', data=self.sign_in_form, follow_redirects=True)
+
+        self.assertEqual(response.status_code, 403)
+        self.assertTrue('Error - Not signed in'.encode() in response.data)
```
tej99: No newline at end of file

PR #245, opened 2017-11-20
Add create app function

andrewmil:
App initialisation has been moved to a create_app function in it's own file to make testing easier.

Comments have also been added to this process to make the steps to initialise the app more clear

Additional tests have also been created to test the app initialisation process


frontstage/cloud/cloudfoundry.py
```diff
@@ -12,18 +12,6 @@ def __init__(self):
...
-        return self._cf_env.port
-
-    @property
-    def host(self):
-        return self._host
-
-    @property
-    def protocol(self):
-        return self._protocol
-
```
tej99: Why have you deleted this?

andrewmil: we don't use it

tej99: The variable names indicate they're 'private' so to access them the property methods should be used

PR #246, opened 2017-11-20
Add info endpoint test

andrewmil:
Added a test to exercise the info endpoint using git_info file


git_info
```diff
@@ -0,0 +1,3 @@
+{
+  "test": "test"
+}
```
tej99: What is this file for?

tests/app/test_info.py
```diff
@@ -0,0 +1,16 @@
...
+class TestInfo(unittest.TestCase):
+
+    def setUp(self):
+        self.app = app.test_client()
+
+    def test_info_with_git_info(self):
+        response = self.app.get("/info")
+
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue('"name": "ras-frontstage"'.encode() in response.data)
```
gemmai95: Could you not use assert in rather than use assert true in?

rhiGriff:
Should the same test be added into the behave tests to test the info endpoint?

git_info
andrewmil: :+1: this is a good idea

tests/app/test_info.py
andrewmil: This makes a lot of sense, I think we use this assertTrue method in pretty much all the tests in frontstage/frontstage-api but I'll have a look at changing them

andrewmil:
@rhiGriff Probably, I haven't looked at the behave tests in a long time and wasn't aware we were running them anywhere currently. Are we going to use them for CI tests in the future?

git_info
andrewmil: Though frontstage-api also uses this method and the file does get overwritten when it is deployed

rhiGriff:
Behave tests have been used but it's not certain what will be used, although it is worth keeping the behave tests up to date

git_info
tej99: Fair play, just think it's better to not have a default file for this. Similar to when we have env vars set to default values that can easily make their way to production.

tej99: https://github.com/ONSdigital/ras-frontstage-api/pull/22added the change to frontstage-api

PR #247, opened 2017-11-23
Pin the python buildpack version to manifest

jcox-dev:
This ensures a later version of Pipenv is used and avoids [setuptools issue](https://github.com/cloudfoundry/python-buildpack/issues/87).


jcox-dev:
Not needed now that we've upgraded the version of our buildpacks in Cloud Foundry to 1.6.2.

PR #248, opened 2017-11-23
Use FlaskForm forms for secure messaging view

andrewmil:
Aligned the secure message form template with the rest of frontstage, using FlaskForm's.

Several hidden fields have to be passed through when loading this form to initialise it with the right data for both a draft and a viewed thread message.

### DIFF
--- a/frontstage/models.py
+++ b/frontstage/models.py
@@ -1,6 +1,3 @@
-"""
-This module contains the data model for the application
-"""
 import logging
 import enum
 import phonenumbers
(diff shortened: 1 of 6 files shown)


frontstage/models.py
```diff
@@ -133,17 +69,12 @@ def validate_password(form, field):
 
 
 class LoginForm(FlaskForm):
-    """Login form."""
-
     username = StringField('Email Address', [InputRequired("Email Address is required"),
```
gemmai95: Is there reasoning behind removing all the docstrings?

andrewmil: We don't use them consistently at the moment.

And also a docstring that just repeats the name of the class doesn't seem that useful

rhiGriff: Wouldn't it be more useful to update the docstring rather than to delete it?

andrewmil: I don't see the point in writing up one doctring for this form right now. 
We should probably agree as a team how we're going to document our code across RAS?

rhiGriff: I thought as a dev team we decided on PEP 8 if so docstrings are part of PEP 8 are they not?

andrewmil: Yep I believe so https://www.python.org/dev/peps/pep-0257/
I feel like adding these in is a much wider task than is covered in this PR however seeing as there's so much to be documented

gemmai95: I think this needs to be brought to the attention of Dave.

gemmai95: Or add a todo comment to make sure they get done later.

frontstage/models.py
```diff
@@ -170,6 +98,33 @@ def validate_password(form, field):
...
+    thread_id = HiddenField('Thread id')
+    hidden_subject = HiddenField('Hidden Subject')
+
+    @staticmethod
+    def validate_subject(form, field):
+        subject = form['hidden_subject'].data if form['hidden_subject'].data else field.data
+        if len(subject) > 96:
+            raise ValidationError('Subject field length must not be greater than 100')
+        if form.send.data and not subject:
+            raise ValidationError('Please enter a subject')
```
tej99: I may be wrong but i think in the current functionality users can save a draft with no subject. Looks like this change will take that ability away. Need to add this validation only if it's a send submit.

frontstage/models.py
```diff
@@ -170,6 +98,33 @@ def validate_password(form, field):
...
+
+    @staticmethod
+    def validate_body(form, field):
+        body = field.data
+        if len(body) > 10000:
+            raise ValidationError('Body field length must not be greater than 10000')
+        if form.send.data and not body:
```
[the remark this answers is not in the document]
andrewmil: `form.send.data` is only true when the user is sending a message so the body check isn't carried out when a draft is sent.

A test covering this scenario was introduced in this PR too, https://github.com/ONSdigital/ras-frontstage/pull/248/files#diff-528390fa8838df9f7c381705e07e2978R204

PR #250, opened 2018-01-04
align ports and basic auth

rhiGriff:
Align default port numbers for all apps and default basic auth. Port numbers for all apps can be found here: https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/94863418/Developer+Environment+Port+Mappings

### DIFF
--- a/run.py
+++ b/run.py
@@ -15,5 +15,5 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 if __name__ == '__main__':
     port = app.config['PORT']
-    logger.info(f'Starting listening port "{port}"')
+    logger.info('Starting listening port: ', port=port)
     app.run(debug=app.config['DEBUG'], host='0.0.0.0', port=int(port))
--- a/run.py
+++ b/run.py
@@ -13,6 +13,6 @@ from structlog import wrap_logger  # NOQA  # pylint: disable=wrong-import-positi
 logger = wrap_logger(logging.getLogger(__name__))
 
 if __name__ == '__main__':
-    port = os.getenv('FS_DEV_PORT', 8080)
+    port = os.getenv('FS_DEV_PORT', 8082)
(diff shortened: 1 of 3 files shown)


run.py
```diff
@@ -13,6 +13,6 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 if __name__ == '__main__':
-    port = os.getenv('FS_DEV_PORT', 8080)
+    port = os.getenv('FS_DEV_PORT', 8082)
```
benjefferies: Dockerfile needs updating too

rhiGriff: This shouldn't affect docker

benjefferies: The expose port should reflect the port the app is listening on https://github.com/ONSdigital/ras-frontstage/blob/master/Dockerfile#L6. You'll also need to update the docker compose files https://github.com/ONSdigital/ras-rm-docker-dev/blob/master/.env

madeye-matt: @benjefferies I've spoken to Rhi - I am going to take a look at ras-rm-docker-dev repo - see if we can align the ports there with the ports in Rhi's list (from link at top)

rhiGriff:
[Align all application port numbers to run locally](https://trello.com/c/DNzTTGD7/31-align-all-application-port-numbers-to-run-locally)

run.py
jcox-dev: This is doing something different to the other repos and not using the config for the port number. Is that ok?

andrewmil: This is something left over from way back when frontstage was started, would be good to bring it in line with the other repos

jcox-dev: Attached a simple patch for [using config](https://github.com/ONSdigital/ras-frontstage/files/1606695/use_config.txt).

jcox-dev:
@gemmai95 true that. I should have just dismissed the "requested changes" 👍

PR #255, opened 2018-01-26
Us334 add survey design code

emilioward:
Added in HTML templates for adding a survey, confirmation page and error screens. Highlighting of adding a new survey has an environment variable to turn it on or off. 
This has been discussed as a best approach for this feature in case of failures in the future. 
Tested against frontstage api during development.


config.py
```diff
@@ -56,17 +59,20 @@ class Config(object):
...
     DEVELOPMENT = True
     DEBUG = True
     TEMPLATES_AUTO_RELOAD = True
     SECRET_KEY = os.getenv('SECRET_KEY', 'ONS_DUMMY_KEY')
-    JWT_SECRET = os.getenv('JWT_SECRET', 'vrwgLNWEffe45thh545yuby')
+    JWT_SECRET = os.getenv('JWT_SECRET', 'testsecret')
     SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME', 'test_user')
     SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD', 'test_password')
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
+    NEW_SURVEY_NOTIF_HIGHLIGHTING = os.getenv('NEW_SURVEY_NOTIF_HIGHLIGHTING', '1')
```
gemmai95: You don't need this both here and in config. It's only if it's needed but not in config or the default is different. As Development config inherits from the config object where it is already defined.

gemmai95:
[US334 Add a Survey (XL)](https://trello.com/c/sHDqihr7/23-us334-add-a-survey-xl)

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+@jwt_authorization(request)
+def add_survey_submit(session):
+    party_id = session['party_id']
+
+    logger.info('Assigning new survey to a user')
+    encrypted_enrolment_code = request.args.get('encrypted_enrolment_code')
+    enrolment_code = cryptographer.decrypt(encrypted_enrolment_code.encode()).decode()
+    json_params = {
+        "enrolment_code": enrolment_code,
+        "initial": True,
```
tej99: Still some questions around the need of this for adding a survey.

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+cryptographer = Cryptographer()
 
 
 @surveys_bp.route('/', methods=['GET'])
 @jwt_authorization(request)
 def logged_in(session):
     party_id = session.get('party_id')
     surveys_list = get_surveys_list(party_id, 'todo')
-    return render_template('surveys/surveys-todo.html', _theme='default', surveys_list=surveys_list)
+    template_data = {"survey": {"justAdded": "false"}}
```
tej99: Can we not just pass a 'just_added' boolean flag here and all the other endpoints instead of passing a dictionary with a string 'true'/'false'?

emilioward: I just followed the same format that was done throughout frontstage with the template data flags as strings. I think we could change it but if we do that shouldn't we change it and make it consistent with all the other endpoints too?

tej99: The case for using template_data in other places is valid because it is used for error messages not just as a flag.

config.py
```diff
@@ -57,6 +60,8 @@ class Config(object):
     ACCESS_CASE = 'surveys/access-case'
     DOWNLOAD_CI = 'surveys/download-ci'
     UPLOAD_CI = 'surveys/upload-ci'
+    ADD_SURVEY = 'surveys/add-survey'
+    CONFIRM_ADD_ORGANISATION_SURVEY = 'surveys/add-survey/confirm-add-organisation-survey'
```
andrewmil: This is fine right now because it's consistent with the rest of the repo but I think it'd be good to move these out of `config.py` in the future

frontstage/templates/partials/header.html
```diff
@@ -30,7 +30,7 @@
 
     {% if 'authorization' in request.cookies %}
     <div class="sign-out header__sign-out">
-      <a class="u-ml-s nowrap" href="{{ url_for('sign_in_bp.logout') }}">Sign out</a>
+      <a id="SIGN_OUT_BTN" class="u-ml-s nowrap" href="{{ url_for('sign_in_bp.logout') }}">Sign out</a>
```
andrewmil: is there a reason this id is all caps?

frontstage/templates/surveys/surveys-add.html
```diff
@@ -0,0 +1,57 @@
+{% import 'partials/section.html' as section %}
```
andrewmil: It would be good to go through these html files and make the indenting consistent, the browser wont care but they're be easier to manage if it's more consistent.

frontstage/templates/surveys/surveys-confirm-organisation.html
```diff
@@ -0,0 +1,31 @@
+{% import 'partials/section.html' as section %}
```
andrewmil: Is this exactly the same as the other `confirm-organisation-survey.html` page? 
If it is it'd be better to reuse it rather than adding a new one

frontstage/views/surveys.py
andrewmil: I'm with Tej on this one, this should be a boolean

config.py
```diff
@@ -32,6 +32,9 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
+    # Survey highlighting config
+    NEW_SURVEY_NOTIF_HIGHLIGHTING = os.getenv('NEW_SURVEY_NOTIF_HIGHLIGHTING', '1')
```
andrewmil: I think this works better as a boolean rather than a number

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+
+    sorted_surveys_list = sorted(surveys_list, key=lambda k: k['collection_exercise']['scheduledReturnDateTime'],
+                                 reverse=True)
+
+    # Checks if new survey is added and renders "survey added" notification
+    if request.args.get('new_survey') and ['NEW_SURVEY_NOTIF_HIGHLIGHTING'] == 1:
```
andrewmil: We shouldn't need this big if statement here
We should be able to do something like
`survey_added = True if (request.args.get('new_survey') and ['NEW_SURVEY_NOTIF_HIGHLIGHTING']) else False`

andrewmil: This way we'd only need one return statement here

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+                           _theme='default', sorted_surveys_list=sorted_surveys_list)
 
 
 @surveys_bp.route('/history', methods=['GET'])
 @jwt_authorization(request)
 def surveys_history(session):
     party_id = session['party_id']
     surveys_list = get_surveys_list(party_id, 'history')
-    return render_template('surveys/surveys-history.html',  _theme='default',
+    template_data = {"survey": {"justAdded": "false"}}
```
andrewmil: If we do as Tej recommended above and have `survey_added` be a boolean we could leave this unchanged

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+            'enrolment_code': enrolment_code,
+            'initial': True
+        }
+        response = api_call('POST', app.config['VALIDATE_ENROLMENT'], json=request_data)
+
+        # Handle API errors
+        if response.status_code == 404:
+            logger.info('Enrolment code not found')
+            template_data = {"error": {"type": "failed"}}
+            return render_template('surveys/surveys-add.html', _theme='default',
```
andrewmil: These return statements are all identical, we can probably do this in less lines

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+        encrypted_enrolment_code = cryptographer.encrypt(enrolment_code.encode()).decode()
+        logger.info('Successful enrolment code submitted')
+        return redirect(url_for('surveys_bp.survey_confirm_organisation',
+                                encrypted_enrolment_code=encrypted_enrolment_code,
+                                _external=True,
+                                _scheme=getenv('SCHEME', 'http')))
+
+    elif request.method == 'POST' and not form.validate():
+        logger.info('Invalid character length, must be 12 characters')
+        template_data = {"error": {"type": "failed"}}
```
andrewmil: We don't need to define template_data here, just pass the json into the template directly, i.e.
`return render_template('surveys/surveys-add.html', _theme='default', form=form, data={"error": {"type": "failed"}})`

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+            template_data = {"error": {"type": "failed"}}
+            return render_template('surveys/surveys-add.html', _theme='default',
+                                   form=form, data=template_data), 200
+        elif response.status_code == 400:
+            logger.info('Enrolment code already used')
+            template_data = {"error": {"type": "failed"}}
+            return render_template('surveys/surveys-add.html', _theme='default',
+                                   form=form, data=template_data), 200
+        elif response.status_code != 200:
+            logger.error('Failed to submit enrolment code')
```
rhiGriff: Can we add status=response.status_code to log line?

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+        # Get and decrypt enrolment code
+        encrypted_enrolment_code = request.args.get('encrypted_enrolment_code', None)
+        enrolment_code = cryptographer.decrypt(encrypted_enrolment_code.encode()).decode()
+
+        logger.info('Attempting to retrieve data for confirm add organisation/survey page')
+        response = api_call('POST', app.config['CONFIRM_ADD_ORGANISATION_SURVEY'],
+                            json={'enrolment_code': enrolment_code})
+
+        if response.status_code != 200:
+            logger.error('Failed to retrieve data for confirm add organisation/survey page')
```
rhiGriff: Same here

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+
+    return render_template('surveys/surveys-add.html', _theme='default',
+                           form=form, data={"error": {}})
+
+
+@surveys_bp.route('/add-survey/confirm-organisation-survey', methods=['GET'])
+@jwt_authorization(request)
+def survey_confirm_organisation(session):
+
+    if request.method == 'GET':
```
andrewmil: There is only a `GET` method here so we don't need to check the method

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+    json_params = {
+        "enrolment_code": enrolment_code,
+        "initial": True,
+        "party_id": party_id
+    }
+    response = api_call('POST', app.config['ADD_SURVEY'],
+                        json=json_params)
+
+    if response.status_code != 200:
+        logger.error('Failed to assign user to a survey')
```
rhiGriff: And here

frontstage/views/surveys.py
andrewmil: Yeah did we get confirmation about what cases need to be posted during this process?

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+
+    logger.info('Assigning new survey to a user')
+    encrypted_enrolment_code = request.args.get('encrypted_enrolment_code')
+    enrolment_code = cryptographer.decrypt(encrypted_enrolment_code.encode()).decode()
+    json_params = {
+        "enrolment_code": enrolment_code,
+        "initial": True,
+        "party_id": party_id
+    }
+    response = api_call('POST', app.config['ADD_SURVEY'],
```
andrewmil: This can be over one line

frontstage/views/surveys.py
```diff
@@ -1,37 +1,164 @@
...
+                               organisation_name=response_json['organisation_name'],
+                               survey_name=response_json['survey_name'])
+
+
+@surveys_bp.route('/add-survey/add-survey-submit', methods=['GET'])
+@jwt_authorization(request)
+def add_survey_submit(session):
+    party_id = session['party_id']
+
+    logger.info('Assigning new survey to a user')
```
andrewmil: For most of these log lines we have access to the `party_id` it'd probably be good practice to include it wherever we can

frontstage/templates/partials/header.html
rhiGriff: id Being all caps makes no difference, it depends on who adds it etc, there are some that are all caps and others that are not

andrewmil: :+1: looks like there's no standards around capitalisation anyway so will be fine

frontstage/views/surveys.py
andrewmil: This is logged in `ApiError`

config.py
emilioward: Yeah I agree

frontstage/templates/surveys/surveys-confirm-organisation.html
emilioward: Yes it is the same as the register page, I thought it warranted a new html page as this is for adding a survey for the future if changes needed to be made then the foundation is there. I do see your point but it depends how this screen is going to be iterated.

andrewmil: If it's the exact same I think we should just reuse it

emilioward: Sorry my mistake these pages are different, they point to different places once you continue

frontstage/views/surveys.py
```diff
@@ -1,35 +1,156 @@
...
-    return render_template('surveys/surveys-todo.html', _theme='default', surveys_list=surveys_list)
+
+    sorted_surveys_list = sorted(surveys_list, key=lambda k: k['collection_exercise']['scheduledReturnDateTime'],
+                                 reverse=True)
+
+    # Checks if new survey is added and renders "survey added" notification
+    if request.args.get('new_survey') and ['NEW_SURVEY_NOTIF_HIGHLIGHTING']:
+        logger.info('New survey added highlighted')
+        return render_template('surveys/surveys-todo.html', _theme='default',
+                               surveyAdded=True, sorted_surveys_list=sorted_surveys_list,
```
andrewmil: `surveyAdded` should be `survey_added`

frontstage/views/surveys.py
```diff
@@ -1,35 +1,156 @@
...
     party_id = session.get('party_id')
     surveys_list = get_surveys_list(party_id, 'todo')
-    return render_template('surveys/surveys-todo.html', _theme='default', surveys_list=surveys_list)
+
+    sorted_surveys_list = sorted(surveys_list, key=lambda k: k['collection_exercise']['scheduledReturnDateTime'],
+                                 reverse=True)
+
+    # Checks if new survey is added and renders "survey added" notification
+    if request.args.get('new_survey') and ['NEW_SURVEY_NOTIF_HIGHLIGHTING']:
+        logger.info('New survey added highlighted')
```
andrewmil: These lines aren't tested

frontstage/views/surveys.py
```diff
@@ -1,35 +1,156 @@
...
+                                encrypted_enrolment_code=encrypted_enrolment_code,
+                                _external=True,
+                                _scheme=getenv('SCHEME', 'http')))
+
+    elif request.method == 'POST' and not form.validate():
+        logger.info('Invalid character length, must be 12 characters')
+        return render_template('surveys/surveys-add.html', _theme='default', form=form,
+                               data={"error": {"type": "failed"}})
+
+    return render_template('surveys/surveys-add.html', _theme='default',
```
andrewmil: We don't test this line

frontstage/views/surveys.py
```diff
@@ -1,35 +1,156 @@
...
+                           enrolment_code=enrolment_code,
+                           encrypted_enrolment_code=encrypted_enrolment_code,
+                           organisation_name=response_json['organisation_name'],
+                           survey_name=response_json['survey_name'])
+
+
+@surveys_bp.route('/add-survey/add-survey-submit', methods=['GET'])
+@jwt_authorization(request)
+def add_survey_submit(session):
+    party_id = session['party_id']
```
andrewmil: None of this function is tested by the looks of it

PR #259, opened 2018-01-29
Update logging standards

rhiGriff:


frontstage/views/passwords.py
```diff
@@ -58,7 +58,7 @@ def post_forgot_password():
                                    form=form, data=template_data)
 
         if response.status_code != 200:
-            logger.error('Unable to send password change request')
+            logger.error('Unable to send password change request', status=response.status_code)
```
andrewmil: Status code is logged out when we call ApiError so do we need to include it here?

PR #263, opened 2018-02-02
Remove gulp.js functionality

andrewmil:
I don't believe anyone has been using this frontend functionality when we've been developing
If we're not going to use it then it's just unnecessarily complicating the repo so I'm proposing removing it entirely

In the future I believe we'll be moving towards using a CDN similar to EQ and that should give us everything we need


andrewmil:
@benjefferies I don't believe it will
Right now gulp would be used to compile scss/js from `common/assets/styles` folder at the top level into the `css_responsive` and `css_fixed` files inside the frontstage folder. But we would do this before pushing to the repo, when we build/deploy it just uses what is in `ras-frontstage/frontstage/assets` and `ras-frontstage/frontstage/styles`.

When anyone has been developing on frontstage they've just been directly editing these files and I can't see anything wrong with that right now. I'm pretty confident that none of this stuff is used when we deploy or run frontstage locally

PR #264, opened 2018-02-02
Separate endpoints out into their own files

andrewmil:
### !DONT PANIC!
This PR isn't as large as it looks, it's just a structural change

Several of the files in `/views` were getting quite large so here I've separated out the endpoints into their own folders/files. For example the `surveys.py` file has been split into several files such as `add_survey.py`, `surveys_list.py` inside a `/surveys` folder.
The blueprints used by these endpoints are created in the `__init__.py` files inside these new folders


andrewmil:
[Refactor Frontstage](https://trello.com/c/SUJzOu8N/52-refactor-frontstage)

frontstage/views/__init__.py
```diff
@@ -10,12 +10,27 @@
 from frontstage.views.info import info_bp
 
 
+# Import endpoints and register blueprints
```
dcdarrell9: Why aren't these imports at the top?

andrewmil: :+1: originally I thought that the blueprints had to be registered before we import the endpoints so this is leftover from that

PR #269, opened 2018-02-26
Us062 access eq

LJBabbage:
Required for US062: EXT - Access an eQ, this is the frontend. Do not merge until https://github.com/ONSdigital/ras-frontstage-api/pull/39 has been merged


tests/app/test_surveys.py
```diff
@@ -89,17 +90,25 @@ def test_surveys_todo_fail(self, mock_request):
     def test_access_survey(self, mock_request):
```
emilioward: Is it worth calling this `test_access_survey_seft` to differentiate between the access eq tests. Just to make it clearer?

PR #270, opened 2018-02-27
Sm137 send message from todo list

gemmai95:
Adds functionality for respondent to send message from their survey todo list and work with version 2 of the secure messaging API. 

- Add buttons to survey todo and history lists for each survey to send a message in regards to RU and survey.

- Change json sent to frontstage api to include case id, survey id as well as ru_ref to handle new version of secure messaging API. For multiple surveys.

Links to other PRs:

https://github.com/ONSdigital/ras-frontstage-api/pull/41
https://github.com/ONSdigital/rasrm-acceptance-tests/pull/60
https://github.com/ONSdigital/ras-secure-message/pull/153


gemmai95:
[SM137: EXT Send SM from To do List](https://trello.com/c/UU26Qrr6/76-sm137-ext-send-sm-from-to-do-list)

frontstage/views/secure_messaging/messages_get.py
```diff
@@ -17,12 +17,14 @@
 @secure_message_bp.route('/messages/', methods=['GET'])
 @secure_message_bp.route('/messages/<label>', methods=['GET'])
 @jwt_authorization(request)
-def messages_get(session, label="INBOX"):
+def messages_get(session, label="INBOX", new_message=False):
     messages_list = get_messages_list(label)
     messages = messages_list['messages']
+    if request.args.get('new_message'):
```
JamesGardiner:
```Python
try:
    new_message = request.args['new_message']
except KeyError:
	sys.exc_clear() # clears the exception - we don't need it

return render_template('secure-messages/secure-messages.html',
					   _theme='default',
                       messages=messages['messages'],
					   links=messages['_links'],
					   label=label,
                       total=unread_msg_total,
					   new_message=new_message)

File: frontstage/views/secure_messaging/create_message.py [hunk 2]
```diff
@@ -16,32 +16,35 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-@secure_message_bp.route('/create-message', methods=['GET', 'POST'])
+@secure_message_bp.route('/create-message/', methods=['GET', 'POST'])
 @jwt_authorization(request)
 def create_message(session):
+    case_id = request.args.get('case_id')
+    survey = request.args.get('survey')
+    ru_ref = request.args.get('ru_ref')
```
AdamHawtin: Should there be some sort of verification that these aren't empty?
If a use navigates to the `/create-messages URL` without the query strings, the page will currently still try to send the message with these fields missing.

frontstage/views/secure_messaging/messages_get.py
ajmaddaford: I don't think the exception is needed, `new_message = request.args.get('new_message', None)` should work.

frontstage/views/secure_messaging/create_message.py
```diff
@@ -50,9 +53,13 @@ def send_message(party_id, is_draft):
     subject = form['subject'].data if form['subject'].data else form['hidden_subject'].data
     message_json = {
         'msg_from': party_id,
+        'msg_to': ['GROUP'],
```
tej99: Are you aware this currently breaks response operations messages view? As the msg_to field does not get populated when retrieving a message.

AndrewTorrance: Not following how it will break it ?

tej99: If a respondent sends a message to internal user, when an internal user tries to go to the messages list view they will be presented with a 500 internal server error. This is because the msg_to field comes back as an empty list. Looks like the reason for this is because internal_user_service.get_user_details is not implemented yet.

tej99: Never mind looks like there's a PR already to fix this https://github.com/ONSdigital/ras-secure-message/pull/156

AndrewTorrance: There is a change imminent which will implement that , the original implementation is flawed , it should have returned dummy data . We still have to handle the case of the msg_to being group when we retrieve data .

AdamHawtin: I think this is actually a bug in r-ops, I've raised it with out team. It happens with messages sent from the current master branch.

frontstage/views/secure_messaging/create_message.py
gemmai95: In the current implementation they should not be able to send a message apart from through the survey list

gemmai95: But yes atm this should not work without these so should have validation

PR #273, opened 2018-03-02
Adding response headers to stop cached page being dispayed. This will…

pricem14pc:
… ensure latest statues are displayed

As we're 302 redirecting when an EQ is launched, an immediate browser 'back' causes a cached My Surveys page to be displayed. This still shows the EQ as 'Not Started' which is incorrect (as the 302 has now changed the EQ status from 'Not Started' to 'Started'). This header causes the browser to round trip to the server to display the correct updated status of 'Started'

### DIFF
--- a/frontstage/views/surveys/surveys_list.py
+++ b/frontstage/views/surveys/surveys_list.py
@@ -24,13 +24,14 @@ def logged_in(session):
                                  reverse=True)
 
     response = make_response(render_template('surveys/surveys-todo.html',
-                           just_added_case_id=request.args.get('case_id'),
-                           _theme='default', sorted_surveys_list=sorted_surveys_list))
+                             just_added_case_id=request.args.get('case_id'),
+                             _theme='default', sorted_surveys_list=sorted_surveys_list))
 
     # Ensure any return to list of surveys (e.g. browser back) round trips the server to display the latest statuses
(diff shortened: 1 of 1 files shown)


frontstage/views/surveys/surveys_list.py
```diff
@@ -23,10 +23,14 @@ def logged_in(session):
     sorted_surveys_list = sorted(surveys_list, key=lambda k: k['collection_exercise']['scheduledReturnDateTime'],
                                  reverse=True)
 
-    return render_template('surveys/surveys-todo.html',
+    response = make_response(render_template('surveys/surveys-todo.html',
                            just_added_case_id=request.args.get('case_id'),
```
andrewmil: These lines should be aligned with the open bracket above

frontstage/views/surveys/surveys_list.py
```diff
@@ -23,10 +23,14 @@ def logged_in(session):
...
-                           _theme='default', sorted_surveys_list=sorted_surveys_list)
+                           _theme='default', sorted_surveys_list=sorted_surveys_list))
 
+    # Ensure any return to list of surveys (e.g. browser back) round trips the server to display the latest statuses
+    response.headers.set("Cache-Control", "no-cache, max-age=0, must-revalidate, no-store")
+
+    return response
```
andrewmil: There should be 2 empty lines between functions here

PR #276, opened 2018-03-08
Fix reply to message

warrenbailey:
RU_REF, CASE_ID and Survey were all missing from the template so were not sent by the frontstage application.

Case ID is now also optional in secure messaging so doesn't always need to be sent


gemmai95:
Just remove the test I added to fail without the case_id and it should be good.

PR #283, opened 2018-03-16
Set expiry on auth token

JamesGardiner:
This PR fixes a bug where authorization tokens are not having their expiry set in the client's browser.

### DIFF
--- a/frontstage/common/authorisation.py
+++ b/frontstage/common/authorisation.py
@@ -58,12 +58,11 @@ def jwt_authorization(request):
                 raise JWTValidationError
 
             if app.config['VALIDATE_JWT']:
-                valid = validate(jwt)
-            if valid:
-                session_handler.update_session()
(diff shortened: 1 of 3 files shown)


frontstage/common/authorisation.py
```diff
@@ -58,12 +58,11 @@ def extract_session_wrapper(*args, **kwargs):
...
-            if valid:
-                session_handler.update_session()
-                return original_function(jwt, *args, **kwargs)
-            else:
-                logger.warning('Token is not valid for this request')
-                raise JWTValidationError
+                if validate(jwt):
+                    session_handler.update_session()
+                    return original_function(jwt, *args, **kwargs)
+                else:
```
andrewmil: We don't need this else as there's a return in the if statement

dcdarrell9: What about if validate(jwt) is False?

PR #292, opened 2018-03-29
Moved duplicate code into single partial

dcdarrell9:
The same code was being used in multiple files, so created a partial to re-use

### DIFF
--- a/frontstage/views/surveys/surveys_list.py
+++ b/frontstage/views/surveys/surveys_list.py
@@ -14,7 +14,7 @@ from frontstage.views.surveys import surveys_bp
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-@surveys_bp.route('/', methods=['GET'])
+@surveys_bp.route('/todo', methods=['GET'])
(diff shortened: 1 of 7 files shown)


frontstage/views/surveys/surveys_list.py
```diff
@@ -14,7 +14,7 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-@surveys_bp.route('/', methods=['GET'])
+@surveys_bp.route('/todo', methods=['GET'])
```
andrewmil: We can have this at `/todo` but we also want `/` to redirect here I think when the user is signed in? I think when a user signs in they are redirected to `/` so that would need to be updated as well

dcdarrell9: When I sign in I get redirected to `/todo` but I'll have a look see 👍

dcdarrell9: So it redirects to `url_for('surveys_bp.logged_in')` which is this route, not named the best, but works.

PR #293, opened 2018-03-29
Upgrade Python to 3.6 and update dependencies

insacuri:
Don't merge until it's correct to do so.  This branch is a test to see how difficult it would be to upgrade and what would need to happen.  With this branch, the tests pass locally with fairly little effort as the only change was to unpin the version of coverage (which is probably a good thing as I imagine it was an old version)


JamesGardiner:
LGTM - though using `3.6` will use 3.6.4, not 3.6.3

ajmaddaford:
> Don't merge until it's correct to do so

@insacuri Any reason not to get this merged?

insacuri:
@ajmaddaford only fear of breaking stuff?  Rhi checked for me, and the buildpack they have in cloudfoundry goes up to 3.6.3 so I think this can be merged.

PR #295, opened 2018-04-06
Remove old messages

insacuri:
Branched off https://github.com/ONSdigital/ras-frontstage/pull/294 (which needs to go in first).  This PR deletes a bunch of now not used messages code and some of the draft code that was previously disabled by removing the save draft buttons


frontstage/views/secure_messaging/create_message.py
```diff
@@ -77,3 +77,26 @@ def send_message(party_id, is_draft, case_id, survey, ru_ref):
...
+
+
+def create_headers():
+    encoded_jwt = SessionHandler().get_encoded_jwt(request.cookies['authorization'])
+    headers = {"jwt": encoded_jwt}
+    return headers
+
+
+def get_message(message_id, label, party_id):
+    logger.debug('Attempting to retrieve message', message_id=message_id, party_id=party_id)
```
atrozzo: I think this is a good info level, give me all teh information about the message id and party id.

frontstage/views/secure_messaging/create_message.py
```diff
@@ -77,3 +77,26 @@ def send_message(party_id, is_draft, case_id, survey, ru_ref):
...
+
+    headers = create_headers()
+    endpoint = 'secure-messaging/message'
+    parameters = {"message_id": message_id, "label": label, "party_id": party_id}
+    response = api_call('GET', endpoint, parameters=parameters, headers=headers)
+
+    if response.status_code != 200:
```
atrozzo: I partially agree with this solution. I think the 200 range shouls always be sucsessfull. One day we should start to create some differentiation about the possible reposnse. http://www.flaskapi.org/api-guide/status-codes/ But I think it s ok for now.

frontstage/views/secure_messaging/create_message.py
```diff
@@ -77,3 +77,26 @@ def send_message(party_id, is_draft, case_id, survey, ru_ref):
...
+        logger.error('Failed to retrieve message', message_id=message_id, party_id=party_id)
+        raise ApiError(response)
+
+    message = json.loads(response.text)
+    logger.debug('Retrieved message successfully', message_id=message_id, party_id=party_id)
```
atrozzo: as aboove consistent with the other log message or are all debug or all info these kinf of.

frontstage/views/secure_messaging/create_message.py
insacuri: What do you mean give you all the information?  That's literally all the information this function gets, other than label, but it doesn't feel worthwhile logging that out.

frontstage/views/secure_messaging/create_message.py
insacuri: Agreed, I think this is code that was already there so until we figure out what should be used (my vote is for raise_for_status()), I'll leave it alone as it works.

frontstage/views/secure_messaging/create_message.py
insacuri: I think this was how it was when I got here, but I'll update it as I think they should all be info level anyway.

frontstage/controllers/conversation_controller.py
```diff
@@ -0,0 +1,96 @@
...
+
+    try:
+        return response.json()
+    except JSONDecodeError:
+        logger.exception("the response could not be decoded", thread_id=thread_id)
+        raise ApiError(response)
+
+
+def get_conversation_list():
+    logger.info("Retrieving threads list")
```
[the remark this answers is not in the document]
insacuri: I took a look at this, I'm not sure how we'd give extra information in the log line.  The user_id is inside the token and is never actually decoded anywhere within frontstage.  Other then that, I'm not sure what I'd add

dcdarrell9:
Looks good 👍 I think we could look to improve test coverage in conversation_controller/message view, it is quite low

dcdarrell9:
Do the `secure-messages` html files get used anymore? I know the inbox/sent tabs are still there in them, so I think we may not need them

JamesGardiner:
Until #294 is merged any chance you can point the merge to that branch - it will make reviewing easier.

PR #296, opened 2018-04-09
Remove unused cryptographer import and creation

insacuri:


insacuri:
@dcdarrell9 I've looked and it is being used in a bunch of places.  This was the only place where it was being created and not used.

PR #297, opened 2018-04-09
Remove theme

insacuri:
Needs to be merged in after https://github.com/ONSdigital/ras-frontstage/pull/294 as it got branched from that. 
This removes the need to have _theme in every view as until we actually need to have different styles as it makes render_templates in view look messy with this random extra parameter needing to be passed.
In fixed.css.map and responsive.css.map, it still has the previous values so I'm not sure if these need to be deleted or regenerated, so don't merge until this gets resolved.


config.py
```diff
@@ -36,6 +36,15 @@ class Config(object):
     RAS_FRONTSTAGE_API_SERVICE = '{}://{}:{}/'.format(RAS_FRONTSTAGE_API_PROTOCOL,
                                                       RAS_FRONTSTAGE_API_HOST,
                                                       RAS_FRONTSTAGE_API_PORT)
+
+    RAS_SECURE_MESSAGING_PROTOCOL = os.getenv('RAS_SECURE_MESSAGING_PROTOCOL', 'http')
+    RAS_SECURE_MESSAGING_HOST = os.getenv('RAS_SECURE_MESSAGING_HOST', 'localhost')
+    RAS_SECURE_MESSAGING_PORT = os.getenv('RAS_SECURE_MESSAGING_PORT', 5050)
+
+    RAS_SECURE_MESSAGING_SERVICE = '{}://{}:{}/'.format(RAS_SECURE_MESSAGING_PROTOCOL,
```
rhiGriff: Can we please keep these consistent with what's been used in Jenkins eg. RAS_SECURE_MESSAGE_SERVICE_PORT etc.

frontstage/common/message_helper.py
```diff
@@ -0,0 +1,81 @@
...
+        logger.exception("Failed to retrieve Subject from thread")
+        raise
+
+
+def get_from_name(message):
+    try:
+        if message['from_internal']:
+            return "ONS Business Surveys Team"
+        msg_from = message['@msg_from']
+        return "{} {}".format(msg_from.get('firstName'), msg_from.get('lastName'))
```
rhiGriff: Should we be using the f"{variable}" format now? 🎨

insacuri: frontstage is on python 3.5, f-strings weren't introduced until python 3.6 :(  I've made a PR to start the process because all our services should be using the latest version where possible.

insacuri:
Closing because there is a v2 version that is cleaner

PR #302, opened 2018-04-11
add banner html and config flag

emilioward:
This PR is to add a banner to the sign in page of frontstage to alert users of downtime to the service. This is enabled or disabled through a config flag and is needed for this weekend

https://trello.com/c/ZuZx85OD

screen shot 2018-04-11 at 14 04 16

### DIFF
--- a/config.py
+++ b/config.py
@@ -20,6 +20,7 @@ class Config(object):
     GOOGLE_ANALYTICS = os.getenv('GOOGLE_ANALYTICS', None)
     SELENIUM_TEST_URL = os.environ.get('SELENIUM_TEST_URL', 'http://localhost:8080')
     NON_DEFAULT_VARIABLES = ['SECRET_KEY', 'SECURITY_USER_NAME', 'SECURITY_USER_PASSWORD', 'JWT_SECRET']
+    AVAILABILITY_BANNER = os.getenv('AVAILABILITY_BANNER', None)
 
     REDIS_HOST = os.getenv('REDIS_HOST', 'localhost')
     REDIS_PORT = os.getenv('REDIS_PORT', 6379)
--- a/frontstage/templates/sign-in/sign-in.html
+++ b/frontstage/templates/sign-in/sign-in.html
@@ -39,6 +39,17 @@
         <h1 class="saturn">You've activated your account</h1>
         <p>You may now sign in.</p>
         {% else %}
+
+        {% if config.AVAILABILITY_BANNER %}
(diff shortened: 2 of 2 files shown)


frontstage/templates/sign-in/sign-in.html
```diff
@@ -39,6 +39,17 @@ <h1 class="panel__title venus">{{ errorType[error][0] }}</h1>
         <h1 class="saturn">You've activated your account</h1>
         <p>You may now sign in.</p>
         {% else %}
+
+        {% if config.AVAILABILITY_BANNER %}
+        <div class="panel panel--simple panel--error">
```
davidmort: not sure we should be using an error styling for this message, shouldn't it be an information panel? see the sdc-pattern-library https://onsdigital.github.io/sdc-global-design-patterns/components/detail/panel.html

emilioward:
[We need to have a banner put up for RAS/RM for the weekend (13th April)](https://trello.com/c/ZuZx85OD/65-we-need-to-have-a-banner-put-up-for-ras-rm-for-the-weekend-13th-april)

frontstage/templates/sign-in/sign-in.html
emilioward: Cool, im going to have a discussion with Nic this morning

insacuri:
I know this is for tomorrow and might've needed to be rushed out but this feels like a quick and dirty solution.
Every time you want to put up a banner to announce downtime you have to do a code change and a release?  That's really not a good solution.  Is the intention of this to be temporary?  If so, can we have a trello card/github issue created so that this can be looked at properly.

emilioward:
Hi @insacuri yeah I agree its not the best solution and not the way I would do it going forward, this is just a temporary solution for now. We do need either a trello card or github issue to look into this properly.

config.py
```diff
@@ -20,6 +20,7 @@ class Config(object):
     GOOGLE_ANALYTICS = os.getenv('GOOGLE_ANALYTICS', None)
     SELENIUM_TEST_URL = os.environ.get('SELENIUM_TEST_URL', 'http://localhost:8080')
     NON_DEFAULT_VARIABLES = ['SECRET_KEY', 'SECURITY_USER_NAME', 'SECURITY_USER_PASSWORD', 'JWT_SECRET']
+    AVAILABILITY_BANNER = os.getenv('AVAILABILITY_BANNER', None)
```
dcdarrell9: Might look nicer with 'False' instead of 'None'?

PR #303, opened 2018-04-11
Remove old messages v2

insacuri:
This is based off https://github.com/ONSdigital/ras-frontstage/pull/295 , but taken off master after the reply functionality was merged in as it caused a number of odd looking merge conflicts.
This PR removes the old message functionality (that was replace by conversations) and more of the draft functionality that is no longer used.
Testing this should be as easy as creating a message for a survey, ensuring errors render correctly and the page looks right.  Then look in the messages tab, ensure all the details are there (the date/time specifically, as at one point it was missing for a really odd reason).  Reply to the message and see if the details for that are still there too.
Finally, check the message in response-operations-ui to ensure that everything in the message looks as it's supposed to.


frontstage/views/secure_messaging/create_message.py
```diff
@@ -73,14 +62,22 @@ def send_message(party_id, is_draft, case_id, survey, ru_ref):
     # If message has previously been saved as a draft add through the message id
     if form["msg_id"].data:
         message_json["msg_id"] = form['msg_id'].data
-    response = api_call('POST', endpoint, parameters={"is_draft": is_draft},
+    # Without is_draft parameter, date/time on the message doesn't get saved correctly,
+    # resulting in missing date/time in conversation list.
```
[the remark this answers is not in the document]
insacuri: Probably yea.   I'm not proud of this bit, but because we're going to be reworking secure-message soon, I don't want to dunk a lot of time into figuring out how 'is_draft' makes the time not appear when you get the thread list down the line.
The comment at least gives some context as to why that seems to 'randomly' be there and will hopefully be fixed down the line.

frontstage/templates/secure-messages/secure-messages-view.html
```diff
@@ -28,12 +16,56 @@
...
+                    {% if "body" in errors %}
+                        <div class="panel panel--simple panel--error panel--spacious" id="body-error">
+                            <p class="error-message">{{ errors['body'][0] }}</p>
+                    {% endif %}
+                    {% if message %}
+                        <label class="venus" for="body">Reply</label>
+                    {% else %}
+                        <label class="venus" for="body">Message</label>
+                    {% endif %}
+                    {{ form.body(id='secure-message-body', class_='input input--textarea input--textarea-message', rows='10') }}
```
ryangrundy7: Should this have something like maxlength='10000'?

ryangrundy7: Should this have something like maxlength='10000'?

insacuri: I only removed the included template and dropped the content of it into here.  If it's wrong here then it's wrong on master too and should have a branch taken off and fixed there.
I'm trying to reduce the scope of the changes in this branch because I'm deleting/changing quite a few files.

insacuri:
Ran both unit and acceptance tests and it passed (apart from 1 collection exercise test that has nothing to do with messaging functionality)

PR #306, opened 2018-04-12
Fix Secure Message Env Variable Names

AdamHawtin:
The secure message config variables were looking for the wrong environment variable names, this change corrects them.


dcdarrell9:
@JamesGardiner No, these were changed and I changed them back to the original but forgot parts.. oops

PR #311, opened 2018-04-13
Remove unused behave tests and test running scripts

insacuri:
Testing this should be simple.  The unit and acceptance tests should still pass.  All the code (I believe) wasn't being run anywhere and can be safely deleted.
The dependancies deleted are ones that were being used by the behave tests so this should also not have any adverse effects.  Someone else should try and regenerate the lock file to make sure it's not too different from what it regenerated into.


ajmaddaford:
There are a lot of other dependencies updated, can we just remove behave (maybe using the pipenv --keep-outdated flag)?

dcdarrell9:
Looks good to me, these were never being used. But what Andrew said about the outdated would be good ^

PR #312, opened 2018-04-17
Apply security headers

benjefferies:
We need to add the following headers to response-operations-ui and
frontstage

` Strict-Transport-Security: max-age=31536000; includeSubDomains` -
Enforce all communication over TLS

`Content-Security-Policy: default-src 'self'` - Mitigate and report XSS
attacks. Do we need to allow google analytics?

`X-XSS-Protection: 1` - Enables XSS filtering (usually default in
browsers). If a cross-site scripting attack is detected, the browser
will sanitize the page (remove the unsafe parts).

`X-Content-Type-Options: nosniff` - Prevent mime sniffing

`Referrer-Policy: same-origin` - Only send referrer information to the
same domain


ajmaddaford:
eQ uses [Talisman](https://github.com/GoogleCloudPlatform/flask-talisman) to set up the secure headers - https://github.com/ONSdigital/eq-survey-runner/blob/master/app/setup.py#L163, also setting the CSP policy for CDN and google analytics access.

frontstage/create_app.py
```diff
@@ -43,9 +52,21 @@ def create_app_object():
...
     return app
+
+
+
+def setup_secure_headers(app):
+    Talisman(
+        app,
+        content_security_policy=CSP_POLICY,
+        content_security_policy_nonce_in=['script-src'],
+        session_cookie_secure=True, # Will setting True this cause issues when https is terminated?
```
ajmaddaford: This won't cause issues when https is terminated before the app as the client (browser) will be sending it over https. The secure flag tells the browser to only send the cookie over https.

Setting the secure flag on the cookie will cause an issue in environments without https e.g. the docker compose setup.

benjefferies: Cool I'll make this configurable then

benjefferies:
@rjweeks will the content security policy work with google analytics?

frontstage/create_app.py
```diff
@@ -13,6 +14,14 @@
 
 cf = ONSCloudFoundry()
 
+CSP_POLICY = {
+    'default-src': ["'self'", 'https://cdn.ons.gov.uk', ],
+    'font-src': ["'self'", 'data:', 'https://cdn.ons.gov.uk', ],
+    'script-src': ["'self'", 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ],
+    'connect-src': ["'self'", 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ],
+    'img-src': ["'self'", 'data:', 'https://www.google-analytics.com', 'https://cdn.ons.gov.uk', ]
```
ajmaddaford: There are more sites resources are loaded from:
```
https://maxcdn.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css
https://code.jquery.com/jquery-3.1.1.min.js
https://fonts.gstatic.com/s/lato/v11/<lots of files>.woff2
```

config.py
```diff
@@ -20,6 +20,7 @@ class Config(object):
     GOOGLE_ANALYTICS = os.getenv('GOOGLE_ANALYTICS', None)
     NON_DEFAULT_VARIABLES = ['SECRET_KEY', 'SECURITY_USER_NAME', 'SECURITY_USER_PASSWORD', 'JWT_SECRET']
     AVAILABILITY_BANNER = os.getenv('AVAILABILITY_BANNER', False)
+    SECURE_COOKIES = os.getenv('SECURE_COOKIES', True)
```
ajmaddaford: When running locally on Docker, the secure flag isn't set for me on the session cookie.

ajmaddaford: I think the getting of booleans from the environment needs to be looked at. The only way to get this to evaluate to `False` is to set the env var to empty, which doesn't seem possible in the docker-compose setup. This should evaluate environment variable booleans in a similar way to survey runner (https://github.com/ONSdigital/eq-survey-runner/blob/master/app/settings.py#L15).

benjefferies:
This will break things if it goes in as it is. CSP is preventing executing inline javascript. We will need to add logic to generate and inject a nonce into the html like https://github.com/ONSdigital/eq-survey-runner/blob/b5168bb07890cefd64658f0ef0f426b810f7d3e2/app/talisman.py.

It will need more work so closing the PR and re-opening when this work is done

PR #313, opened 2018-04-18
Removing caching on frontstage

ryangrundy7:
This PR stops the browser from caching on Frontstage

To test:

- Click on an unread message and when navigating back to the messages page, that message should become read.


insacuri:
Just want to double check, these headers don't affect the caching of styles?

frontstage/create_app.py
```diff
@@ -46,6 +50,8 @@ def create_app_object():
     @app.after_request
     def apply_caching(response):
```
ajmaddaford: As `X-Frame-Options` isn't a cache header the method should be renamed something like `apply_headers`

PR #315, opened 2018-04-19
Simplify info endpoint

insacuri:
The git_info file didn't provide any value.  This makes the info endpoint return more sensible values.


insacuri:
@jwgwalton It shouldn't affect it. I checked and all it's doing it curling the endpoint but not actually checking the response (other then for a 200 response)

andrewmil:
@insacuri Not sure, I can't remember where the idea for this came from, might be best to check with Mark Price in case it's being used for something

insacuri:
Talked to @davidmort about it and it is used and useful.  This'll get looked at when we start using concourse in the future

PR #316, opened 2018-04-23
Add google analytics events.

rjweeks:


dcdarrell9:
if you have a look in eq, there might be a nicer way to add these in if we're going to add them in more places
https://github.com/ONSdigital/eq-survey-runner/search?utf8=%E2%9C%93&q=ga&type=

frontstage/templates/surveys/surveys-access.html
```diff
@@ -30,7 +30,8 @@ <h2 class="neptune">Download</h2>
 
 
 <a id="download-survey-button" class="btn action-button btn--primary" target="_blank"
-   href="{{ url_for('surveys_bp.download_survey', cid=collection_instrument_id, case_id=case_id) }}">
+   href="{{ url_for('surveys_bp.download_survey', cid=collection_instrument_id, case_id=case_id) }}"
```
tej99: Doesn't look like cid=collection_instrument_id is actually needed. download_survey just requires case_id.

frontstage/views/surveys/upload_survey.py
```diff
@@ -58,6 +58,7 @@ def upload_survey(session):
         return redirect(url_for('surveys_bp.upload_failed',
                                 _external=True,
                                 case_id=case_id,
+                                party_id=party_id,
```
tej99: Might not need to pass the party_id

frontstage/views/surveys/upload_survey_failed.py
```diff
@@ -14,7 +15,17 @@
 @jwt_authorization(request)
 def upload_failed(session):
     case_id = request.args.get('case_id', None)
+    party_id = request.args.get('party_id', None)
```
tej99: Can get party_id from the session instead of passing it in. party_id = session['party_id']

PR #318, opened 2018-04-23
fix availability banner

emilioward:
This PR is to fix availability banner config to be set to true and false as an environment variable for Jenkins properly as before it was evaluated incorrectly. Tested this on my own machine


JamesGardiner:
@davidmort I'm closing this until we get clear story requirements.

PR #319, opened 2018-04-25
adding hash to symbols for password

rhiGriff:
Allowing the hash symbol to be accepted in the password to stop the client side error message appearing when a hash is entered in the password. This has now been extended to add other special characters that are missing.


rhiGriff:
[Ext - Password Criteria is not correct - # is not being accepted on the Front end.](https://trello.com/c/wRnVd6AR/205-ext-password-criteria-is-not-correct-is-not-being-accepted-on-the-front-end)

PR #320, opened 2018-04-26
[TBL] Use the CI service directly for download and upload

jcox-dev:
Change ras-frontstage to directly call collection instrument service instead of proxying through ras-frontstage-api.

1. Development - Code needs to be changed in ras-frontstage and ras-frontstage-api.
2. Debugging - Unnecessary hops makes debugging more difficult
3. Code reviews - split over multiple repos

This PR may look like a lot of code changes but most of it has been copied over from fronstage-api. The main thing is to test is running it against the acceptance tests and ensuring uploading and downloading a survey works as expected for a respondent.

This is in an effort to deprecate [frontstage-api](https://github.com/ONSdigital/ras-frontstage-api) by refactoring to use in-house controllers that replicate the implementation from fronstage-api thus reducing networking overhead.


jcox-dev:
[ras-frontstage should directly call collection instrument service (not via ras-frontstage-api)](https://trello.com/c/4iNfdqga/103-ras-frontstage-should-directly-call-collection-instrument-service-not-via-ras-frontstage-api)

rhiGriff:
@jcox-dev There seems to be more to this PR than just using the collection instrument service directly?

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,160 @@
...
+
+def post_case_event(case_id, party_id, category, description):
+    logger.debug('Posting case event', case_id=case_id)
+    validate_case_category(category)
+    url = f"{app.config['RM_CASE_SERVICE']}/cases/{case_id}/events"
+    message = {
+        'description': description,
+        'category': category,
+        'partyId': party_id,
+        'createdBy': 'RAS_FRONTSTAGE_API'
```
gemmai95: The API part needs removed.

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,160 @@
...
+    }
+    response = request_handler('POST', url, auth=app.config['BASIC_AUTH'], json=message)
+
+    if response.status_code != 201:
```
gemmai95: Probably use response.raise_status here

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,57 @@
...
+from frontstage.common.request_handler import request_handler
+from frontstage.controllers import case_controller
+from frontstage.exceptions.exceptions import ApiError, CiUploadError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def upload_collection_instrument(upload_file, case_id, party_id):
+    logger.info('Attempting to upload collection instrument', case_id=case_id)
```
gemmai95: This should be a debug level log and log party_id as well probably.

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,57 @@
...
+    case_controller.post_case_event(case_id,
+                                    party_id=party_id,
+                                    category=category,
+                                    description=f'Survey response for case {case_id} uploaded by {party_id}')
+
+    if response.status_code == 400:
+        logger.exception('Invalid file uploaded', case_id=case_id)
+        raise CiUploadError(response, message=response.text)
+
+    if not response.ok:
```
gemmai95: Maybe use response.raise_for_status here again

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -0,0 +1,57 @@
...
+        logger.exception('Failed to upload collection instrument', case_id=case_id)
+        raise ApiError(response)
+
+    logger.debug('Successfully uploaded collection instrument', case_id=case_id)
+
+
+def download_collection_instrument(collection_instrument_id, case_id, party_id):
+    logger.debug('Downloading collection instrument', collection_instrument_id=collection_instrument_id)
```
gemmai95: Should the party id be logged here too?

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,128 @@
...
+    response = request_handler('PUT', url, auth=app.config['BASIC_AUTH'], json=post_data)
+
+    if response.status_code != 200:
+        raise ApiError(url=url, status_code=response.status_code, description='Failed to change password party')
+
+    logger.debug('Successfully changed password party')
+
+
+def create_account(registration_data):
+    logger.debug('Creating account')
```
gemmai95: Should add a success log line also.

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,128 @@
...
+
+    if response.status_code != 200:
+        logger.error('Failed to verify email address', token=token)
+        raise ApiError(url=url, status_code=response.status_code)
+
+    logger.debug('Successfully verified email address', token=token)
+
+
+def add_survey(party_id, enrolment_code):
+    logger.debug('Adding a survey')
```
gemmai95: Party id should likely be logged here

config.py
```diff
@@ -30,20 +30,45 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
+    RM_CASE_SERVICE_HOST = os.getenv('RM_CASE_SERVICE_HOST', 'localhost')
+    RM_CASE_SERVICE_PORT = os.getenv('RM_CASE_SERVICE_PORT', 8171)
+    RM_CASE_SERVICE_PROTOCOL = os.getenv('RM_CASE_SERVICE_PROTOCOL', 'http')
+    RM_CASE_SERVICE = f'{RM_CASE_SERVICE_PROTOCOL}://{RM_CASE_SERVICE_HOST}:{RM_CASE_SERVICE_PORT}'
```
gemmai95: We are moving away from this approach now it should just be CASE_SERVICE_URL. The host, port and protocol use are being discontinued and  RAS or RM in front of service URL names is not being used anymore.

insacuri:
I'll mention it here just so it doesn't get forgotten about.  Once merged, the jenkins jobs need their config updated, otherwise this'll fail spectacularly!

frontstage/common/request_handler.py
```diff
@@ -0,0 +1,32 @@
+import logging
```
insacuri: Why do we need a request handler?  If it's not too difficult, calls should be in the controller rather than passing it off to a helper module. 
If it's too much hassle for this PR (which is pretty big to be fair), could you make a card for it?

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,160 @@
...
+
+def check_case_permissions(party_id, case_party_id, case_id=None):
+    logger.debug('Party requesting access to case', party_id=party_id, case_id=case_id, case_party_id=case_party_id)
+    if party_id != case_party_id:
+        raise NoSurveyPermission(party_id, case_id, case_party_id)
+
+    logger.debug('Party has permission to access case', party_id=party_id, case_id=case_id, case_party_id=case_party_id)
+
+
+def build_full_case_data(case):
```
insacuri: It's probably best to leave this piece of functionality in frontstage-api for now.    To make the PR easier to read, and to reduce the risk, it should only be the code relating to the case service that should be done here.
The config changes above should be reasonably small (they should only be urls for the collection instrument service and nothing else)

frontstage/controllers/survey_controller.py
```diff
@@ -0,0 +1,24 @@
...
+from structlog import wrap_logger
+
+from frontstage.common.request_handler import request_handler
+from frontstage.exceptions.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_survey(survey_id):
```
insacuri: This should be removed as well as it's related to my comment above.  Moving survey service code should happen in a later PR.

frontstage/common/request_handler.py
jcox-dev: @insacuri I would much prefer to use the requests lib (which is already an abstraction around urllib) directly but this was dropped in to speed up the unpicking of the ported code. 

I'll see what can be done. Consistency curse strikes again!

frontstage/controllers/case_controller.py
jcox-dev: @insacuri the whole module was copied over to support the required functionality. If this is something (among others) that can be left behind, then I'll do so.

frontstage/controllers/survey_controller.py
jcox-dev: @insacuri this controller is not used by these features, I'll get it removed from this branch 👍

config.py
jcox-dev: @gemmai95 Oh, nice! Have we got that proposition documented somewhere?

frontstage/controllers/party_controller.py
jcox-dev: @gemmai95 a lot of of this module is probably not related to this PR. I'll strip it back.

config.py
gemmai95: Unfortunately not just word of mouth atm.

jcox-dev: @gemmai95 I'll see about getting it in this PR anyway 👍

jcox-dev: Example in r-ops found [here](https://github.com/ONSdigital/response-operations-ui/blob/0e2b966d637af71a34d7879b15af53e7a7ead2df/config.py#L33).

frontstage/common/request_handler.py
jcox-dev: @insacuri Updated to remove the wrapper in latest commit.

tej99:
@jcox-dev will need to update ras-rm-docker-dev config and add the config changes to https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/406257665/Pending+Release+Config+Changes

jcox-dev:
Merged into #333.

PR #321, opened 2018-04-27
Adding validation on email address for whitespace

rhiGriff:
When submitting an email address with whitespace in the account registration the application gives a 500 error. This PR adds validation to the email address to check if there is whitespace and gives an error to the user.


rhiGriff:
[Email validation on create account](https://trello.com/c/27Wc7x7f/80-email-validation-on-create-account)

jcox-dev:
Is it possible to use an [EmailField](https://github.com/wtforms/wtforms/blob/086d0bf7d6d73ab9a05e30b861adee627916a71c/wtforms/fields/html5.py#L35) for the form's `email_address` instead?

PR #322, opened 2018-04-30
[TBL] Sign in using the OAuth 2 service directly

jcox-dev:
Change ras-frontstage to directly call the Django OAuth 2 and Party services instead of proxying through ras-frontstage-api.

Notable changes include:
- creation of the `oauth_controller` and `party_controller` (from frontstage_api)
- new exception, `OAuth2Error`, for presenting detailed information to the user
- moving from "DJANGO_" to "OAUTH_" variable names to be explicit


tests/app/mocked_services.py
```diff
@@ -0,0 +1,10 @@
+import json
+from frontstage import app
+
+
+with open('tests/test_data/party/party.json') as json_data:
+	party = json.load(json_data)
+
+
+url_oauth_token = f"{app.config['OAUTH_SERVICE_URL']}/api/v1/tokens/"
+url_get_party_email = f"{app.config['PARTY_SERVICE_URL']}/party-api/v1/respondents/email"
```
rhiGriff: Missing new line at end of file 🎨

frontstage/controllers/oauth_controller.py
```diff
@@ -0,0 +1,44 @@
...
+        'client_id': app.config['OAUTH_CLIENT_ID'],
+        'client_secret': app.config['OAUTH_CLIENT_SECRET'],
+        'username': username,
+        'password': password,
+    }
+    headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+    }
+    response = requests.post(url, headers=headers, auth=app.config['OAUTH_BASIC_AUTH'], data=data)
```
JamesGardiner: Are network level errors handled by the application, or should they be handled here?

frontstage/controllers/oauth_controller.py
```diff
@@ -0,0 +1,44 @@
...
+    except requests.exceptions.HTTPError:
+        if response.status_code == 401:
+            oauth2_error = json.loads(response.text).get('error').get('data', {}).get('detail')
+            logger.exception('Authentication error in oauth2 service', error=oauth2_error)
+            raise OAuth2Error(response, message=oauth2_error)
+        else:
+            logger.exception('Failed to retrieve OAuth2 token')
+        raise ApiError(response)
+
+    oauth2_token = json.loads(response.text)
```
JamesGardiner: Decode errors should be handled here.

frontstage/controllers/oauth_controller.py
```diff
@@ -0,0 +1,44 @@
...
+        'Accept': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+    }
+    response = requests.post(url, headers=headers, auth=app.config['OAUTH_BASIC_AUTH'], data=data)
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        if response.status_code == 401:
+            oauth2_error = json.loads(response.text).get('error').get('data', {}).get('detail')
```
JamesGardiner: Requests decoder is far better at guessing encodings than standard json.loads, so the `Response.json` method should always be preferred to this approach:

```Python
oauth2_error = response.json().get('error', {}).get('data', {}).get('detail')
```

frontstage/controllers/oauth_controller.py
JamesGardiner: Same here, use the `response.json()` method.

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,26 @@
...
+from frontstage.exceptions.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_party_by_email(email):
+    logger.debug('Retrieving party')
+    url = f"{app.config['PARTY_SERVICE_URL']}/party-api/v1/respondents/email"
+    response = requests.get(url, json={"email": email}, auth=app.config['BASIC_AUTH'])
```
JamesGardiner: Same question here as before - where are network level errors handled?

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,26 @@
...
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception('Failed to retrieve party')
+        raise ApiError(response)
+
+    logger.debug('Successfully retrieved party')
+    return json.loads(response.text)
```
JamesGardiner: response.json()

frontstage/views/sign_in/sign_in.py
```diff
@@ -30,49 +36,40 @@ def login():
...
-                logger.error('OAuth 2 server generated 401 which is not understood',
-                             oauth2error=error_message)
-                return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
-
-        if response.status_code != 200:
-            logger.error('Failed to sign in')
-            raise ApiError(response)
+            logger.error('OAuth 2 server generated 401 which is not understood',
+                         oauth2error=error_message)
+            return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
```
JamesGardiner: An explicit else, while not required, would enhance the clarity of this if/else block.

tests/app/mocked_services.py
```diff
@@ -0,0 +1,10 @@
+import json
+from frontstage import app
```
JamesGardiner: Blank line needed between different import sections

frontstage/controllers/oauth_controller.py
jcox-dev: Lazy copy pasta  👍

frontstage/controllers/oauth_controller.py
jcox-dev: Yup - [error_handlers.py](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/error_handlers.py#L22).

jcox-dev: @JamesGardiner should be handled by the [error_handlers.py](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/error_handlers.py#L30).

frontstage/controllers/oauth_controller.py
jcox-dev: Unexpected errors are handled by the generic [catch all](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/error_handlers.py#L45). I'm not sure whether there's an intention to change to a more defensive style but probably worth raising.

frontstage/views/sign_in/sign_in.py
jcox-dev: @JamesGardiner Not a fan of an extra `else` but if it reads clunky then I'll add it here.

JamesGardiner: Nope, just a suggestion 👍

jcox-dev:
[ras-frontstage should directly call django service (not via ras-frontstage-api)](https://trello.com/c/luQFWqS0/104-ras-frontstage-should-directly-call-django-service-not-via-ras-frontstage-api)

rhiGriff:
Can you add any config changes made to the confluence page please? https://digitaleq.atlassian.net/wiki/spaces/RASB/pages/406257665/Pending+Release+Config+Changes

config.py
```diff
@@ -30,6 +33,9 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
+    OAUTH_SERVICE_URL = os.getenv('OAUTH_SERVICE_URL', 'http://localhost:8040')
```
gemmai95: Don't think these should have defaults here only in the dev config.

frontstage/views/sign_in/sign_in.py
```diff
@@ -30,22 +35,17 @@ def login():
...
+        try:
+            oauth2_token = oauth_controller.sign_in(username, password)
+        except OAuth2Error as exc:
+            error_message = exc.message
+            if BAD_AUTH_ERROR in error_message:
                 return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
-            elif 'User account locked' in error_message:
+            elif ACCOUNT_LOCKED_ERROR in error_message:
                 logger.debug('User account is locked on the OAuth2 server')
                 return render_template('sign-in/sign-in.trouble.html', form=form, data={"error": {"type": "account locked"}})
```
gemmai95: I don't think this is handled by the UI anymore or it shouldn't be we don't want to be feeding back that the account is locked

config.py
jcox-dev: @gemmai95 I agree. And would be consistent with the new r-ops changes.

frontstage/views/sign_in/sign_in.py
jcox-dev: @gemmai95 can this case be removed then? (Should it be done so in this PR or separately?).

gemmai95: I have removed it in my PR.

jcox-dev: I'll drop it out here as well then.

config.py
```diff
@@ -75,6 +83,16 @@ class DevelopmentConfig(Config):
     SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD', 'secret')
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
 
+    OAUTH_URL = os.getenv('OAUTH_URL', 'http://localhost:8040')
+    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID', 'admin')
```
tej99: Looks like this is getting set to ons@ons.gov as default

config.py
```diff
@@ -75,6 +83,16 @@ class DevelopmentConfig(Config):
...
+    OAUTH_CLIENT_SECRET = os.getenv('OAUTH_CLIENT_SECRET', 'secret')
+    OAUTH_BASIC_AUTH = (OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET)
+
+    PARTY_URL = os.getenv('PARTY_URL', 'http://localhost:8081')
+    PARTY_USERNAME = os.getenv('PARTY_USERNAME', 'admin')
+    PARTY_PASSWORD = os.getenv('PARTY_PASSWORD', 'secret')
```
tej99: looks like this is getting set to password as default

jcox-dev: @tej99 sounds like the party service failing to be consistent rather than an issue here.

tej99: Sorry that was a comment for OAUTH_CLIENT_SECRET

jcox-dev:
Merged into #333.

PR #323, opened 2018-04-30
[TBL] Build case for failed uploads and survey access

jcox-dev:
Change ras-frontstage to directly call many of the other services to build case details instead of proxying through ras-frontstage-api.

Notable changes include:
- creation of the controllers
- building case details using multiple service calls


jcox-dev:
[ras-frontstage should directly call case service (not via ras-frontstage-api)](https://trello.com/c/tiVMJRsR/101-ras-frontstage-should-directly-call-case-service-not-via-ras-frontstage-api)

tests/app/mocked_services.py
```diff
@@ -0,0 +1,30 @@
+import json
+from frontstage import app
+
+with open('tests/test_data/party/business_party.json') as json_data:
```
andrewmil: :+1: Like this a lot, This could tidy up a lot of the tests across quite a few repos

Pipfile
```diff
@@ -21,10 +17,9 @@ python-jose = "==1.3.2"
 redis = "==2.10.6"
 requests = "*"
 structlog = "==17.2.0"
-
+arrow = "==0.12"
```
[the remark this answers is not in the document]
jcox-dev: @andrewmil I think pinning package versions which we can assume are safe where we can is probably good practice in the Pipfile.

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_collection_exercise(collection_exercise_id):
+    logger.debug('Attempting to retrieve collection exercise', collection_exercise_id=collection_exercise_id)
+    url = f"{app.config['COLLECTION_EXERCISE_SERVICE_URL']}/collectionexercises/{collection_exercise_id}"
+    
+    response = requests.get(url, auth=app.config['BASIC_AUTH'])
```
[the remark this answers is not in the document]
jcox-dev: @andrewmil It sounds like error handling is something to reassess with a design perspective. Currently, most should be handled by one of those defined in error_handlers.py.

Pipfile
```diff
@@ -21,10 +17,9 @@ python-jose = "==1.3.2"
 redis = "==2.10.6"
 requests = "*"
```
tej99: Could pin gunicorn and requests?
`gunicorn = "==19.8.0"`
`requests = "==2.18.4"`

config.py
```diff
@@ -71,10 +96,36 @@ class DevelopmentConfig(Config):
...
+
+    PARTY_URL = os.getenv('PARTY_URL', 'http://localhost:8081')
+    PARTY_USERNAME = os.getenv('PARTY_USERNAME', 'admin')
+    PARTY_PASSWORD = os.getenv('PARTY_PASSWORD', 'secret')
+    PARTY_AUTH = (PARTY_USERNAME, PARTY_PASSWORD)
+
+    SURVEY_URL = os.getenv('SURVEY_URL', 'http://localhost:8080')
+    SURVEY_USERNAME = os.getenv('SURVEY_USERNAME', 'admin')
+    SURVEY_PASSWORD = os.getenv('SURVEY_PASSWORD', 'secret')
+    SURVEY_AUTH = (SURVEY_USERNAME, SURVEY_PASSWORD)
```
tej99: All these will need adding in the ras-rm-docker-dev repo

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -0,0 +1,42 @@
...
+def get_collection_exercise_event(collection_exercise_id, tag):
+    logger.debug('Retrieving collection exercise event', collection_exercise_id=collection_exercise_id, tag=tag)
+    url = f"{app.config['COLLECTION_EXERCISE_URL']}/collectionexercises/{collection_exercise_id}/events/{tag}"
+
+    response = requests.get(url, auth=app.config['COLLECTION_EXERCISE_AUTH'])
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception('Failed to retrieve collection exercise event', collection_exercise_id=collection_exercise_id, tag=tag)
```
tej99: Same thing about logger.exception logging at error level when we probably don't want to do that.

jcox-dev:
Merged into #333.

PR #325, opened 2018-05-01
[TBL] Use the oauth and party service directly for password management

jcox-dev:
Change ras-frontstage to directly call the other services to build case details instead of proxying through ras-frontstage-api.

Notable changes include:
- handling of exceptions for info of service errors
- reset and change passwords functionality no longer proxied through ras-frontstage-api

### DIFF
--- a/config.py
+++ b/config.py
@@ -14,6 +14,9 @@ class Config(object):
     SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME')
     SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD')
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
+    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID')
+    OAUTH_CLIENT_SECRET = os.getenv('OAUTH_CLIENT_SECRET')
+    OAUTH_BASIC_AUTH = (OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET)
     JWT_ALGORITHM = os.getenv('JWT_ALGORITHM', 'HS256')
     JWT_SECRET = os.getenv('JWT_SECRET')
     VALIDATE_JWT = os.environ.get('VALIDATE_JWT', True)
@@ -29,6 +32,9 @@ class Config(object):
     PASSWORD_CRITERIA_ERROR_TEXT = 'Your password doesn\'t meet the requirements'
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
+
+    OAUTH_SERVICE_URL = os.getenv('OAUTH_SERVICE_URL', 'http://localhost:8040')
+    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL', 'http://localhost:8081')
 
     RAS_FRONTSTAGE_API_HOST = os.getenv('RAS_FRONTSTAGE_API_HOST', 'localhost')
     RAS_FRONTSTAGE_API_PORT = os.getenv('RAS_FRONTSTAGE_API_PORT', 8083)
@@ -46,10 +52,6 @@ class Config(object):
                                                         RAS_SECURE_MESSAGE_SERVICE_PORT)
 
     SIGN_IN_URL = 'sign-in'
-
-    REQUEST_PASSWORD_CHANGE = 'passwords/request-password-change'
-    VERIFY_PASSWORD_TOKEN = 'passwords/verify-password-token'
-    CHANGE_PASSWORD = 'passwords/change-password'
 
     VALIDATE_ENROLMENT = 'register/validate-enrolment'
     CONFIRM_ORGANISATION_SURVEY = 'register/confirm-organisation-survey'
--- a/config.py
+++ b/config.py
@@ -33,8 +33,8 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
-    OAUTH_SERVICE_URL = os.getenv('OAUTH_SERVICE_URL', 'http://localhost:8040')
-    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL', 'http://localhost:8081')
+    OAUTH_SERVICE_URL = os.getenv('OAUTH_SERVICE_URL')
+    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL')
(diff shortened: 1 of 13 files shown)


frontstage/views/passwords/reset_password.py
```diff
@@ -17,19 +17,18 @@
...
-        return render_template('passwords/password-expired.html')
-    elif response.status_code == 404:
-        logger.warning('Invalid token sent to party service', token=token)
-        return redirect(url_for('error_bp.not_found_error_page'))
-    elif response.status_code != 200:
-        logger.error('Party service failed to verify token')
-        raise ApiError(response)
+    try:
+        party_controller.verify_token(token)
+    except ApiError as exc:
```
andrewmil: Are exceptions caught by `try/except` statements before the app's error handlers?

frontstage/views/passwords/reset_password.py
```diff
@@ -17,19 +17,18 @@
...
+        if exc.status_code == 409:
+            logger.warning('Token expired', token=token)
```
andrewmil: Is there any risk in logging out this token?

jcox-dev: @andrewmil I'm not sure. 

Team, @ONSdigital/sdc-ras, what do we think?

tej99: Should be alright, It's expired. And if it's not okay, we have a bigger issue with sending the token in the url.

frontstage/controllers/oauth_controller.py
```diff
@@ -0,0 +1,40 @@
...
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+    }
+    response = requests.post(url, headers=headers, auth=app.config['OAUTH_BASIC_AUTH'], data=data)
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        if response.status_code == 401:
+            oauth2_error = response.json().get('detail')
+            logger.exception('Authentication error in OAuth2 service', error=oauth2_error)
```
tej99: logger.exception logs at Error level not sure we want to be logging at Error level for every HTTPError caught. All these logger.exception logs will make our splunk alerts explode.

config.py
```diff
@@ -30,6 +33,9 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
+    OAUTH_SERVICE_URL = os.getenv('OAUTH_SERVICE_URL')
+    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL')
```
tej99: Being a bit picky put response operations is following the pattern of not having 'SERVICE' in the config names. OAUTH_URL, PARTY_URL, SECURE_MESSAGE_URL

frontstage/views/passwords/reset_password.py
gemmai95: Rob seems to think these tokens are fine to be logged out

frontstage/controllers/oauth_controller.py
jcox-dev: @tej99 warning instead?

config.py
```diff
@@ -14,6 +14,9 @@ class Config(object):
     SECURITY_USER_NAME = os.getenv('SECURITY_USER_NAME')
     SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD')
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
+    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID')
```
gemmai95: Think these need changed to username and password.

config.py
```diff
@@ -30,6 +33,9 @@ class Config(object):
     PASSWORD_MIN_LENGTH = 8
     PASSWORD_MAX_LENGTH = 160
 
+    OAUTH_URL = os.getenv('OAUTH_URL')
+    PARTY_URL = os.getenv('PARTY_URL')
```
gemmai95: This should have a party_username and party_password for authentication.

frontstage/controllers/oauth_controller.py
jcox-dev: @tej99 I was following what was implemented in response-operations, like [this for example](https://github.com/ONSdigital/response-operations-ui/blob/12946cb861e69a8285b98fecbf490858953b1d2c/response_operations_ui/controllers/respondent_controllers.py#L27). 

It's probably worth raising with the tech leads / @ONSdigital/sdc-ras to make sure we're not filtering legitimate errors to avoid noise at a Splunk level.

config.py
jcox-dev: @gemmai95 is that the same for all services?

gemmai95: Yes. Another one of these config decisions not documented unfortunately.

config.py
jcox-dev: @gemmai95 I think we're safe with the OAUTH / DJANGO stuff. Still seems to be CLIENT_ID, etc.

config.py
jcox-dev: @gemmai95 Is that because each service may have its own username and password for API auth now?

gemmai95: Yes

jcox-dev:
[ras-frontstage should directly call party service (not via ras-frontstage-api)](https://trello.com/c/2RE2aLfK/106-ras-frontstage-should-directly-call-party-service-not-via-ras-frontstage-api)

frontstage/controllers/oauth_controller.py
tej99: It works fine in the example you were following because it is calling backstage-api which filters out errors, [like the 404](https://github.com/ONSdigital/ras-backstage/blob/da11ef32925c8ca924f977984ecd1da075bf93f8/ras_backstage/controllers/party_controller.py#L105) that should not be logging at error level.

frontstage/views/passwords/reset_password.py
```diff
@@ -48,23 +46,19 @@ def post_reset_password(token):
...
-        logger.warning('Invalid token sent to party service', token=token)
-        return redirect(url_for('error_bp.not_found_error_page'))
-    elif response.status_code != 200:
-        logger.error('Party service failed to verify token')
-        raise ApiError(response)
+    try:
+        party_controller.change_password(password, token)
+    except ApiError as exc:
+        if exc.status_code == 409:
+            logger.warning('Token expired', token=token)
```
rhiGriff: Should be logging the status code in these log lines

jcox-dev:
Merged into #333.

PR #326, opened 2018-05-03
Add inability to sign in unless active account in system

gemmai95:
Respondent should only be able to sign in to system if they are in an active state in the system. This PR goes with the ability to suspend an account but adds the overall functionality to stop people being able to log in to frontstage without an active account.


config.py
```diff
@@ -45,6 +45,10 @@ class Config(object):
                                                         RAS_SECURE_MESSAGE_SERVICE_HOST,
                                                         RAS_SECURE_MESSAGE_SERVICE_PORT)
 
+    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL')
+    PARTY_SERVICE_USERNAME = os.getenv('PARTY_SERVICE_USERNAME')
+    PARTY_SERVICE_PASSWORD = os.getenv('PARTY_SERVICE_PASSWORD')
```
emilioward: I believe we're going for a config pattern where we remove `service` in the naming

gemmai95: Where is this documented?

gemmai95: This needs discussed with the tech leads if some people are doing one thing and others doing something else for consistency purposes. In this case it doesn't make sense to me to remove the service as PARTY refers to something entirely different in regards to the applications as we have a respondent party or business party then the party service that handles this.

config.py
```diff
@@ -75,6 +79,10 @@ class DevelopmentConfig(Config):
     SECURITY_USER_PASSWORD = os.getenv('SECURITY_USER_PASSWORD', 'secret')
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
 
+    PARTY_SERVICE_URL = os.getenv('PARTY_SERVICE_URL', 'http://localhost:8081')
+    PARTY_SERVICE_USERNAME = os.getenv('PARTY_SERVICE_USERNAME', 'admin')
+    PARTY_SERVICE_PASSWORD = os.getenv('PARTY_SERVICE_PASSWORD', 'secret')
```
tej99: Are we going to have different username and password for every service? Might be useful to have PARTY_BASIC_AUTH = (PARTY_SERVICE_USERNAME, PARTY_SERVICE_PASSWORD) like how we currently use BASIC_AUTH. (or PARTY_SERVICE_BASIC_AUTH once we agree on config naming standards...)

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def get_respondent_by_email(email):
+
+    url = f'{app.config["PARTY_SERVICE_URL"]}/party-api/v1/respondents/email'
+    response = requests.get(url, json={"email": email}, auth=(app.config['PARTY_SERVICE_USERNAME'],
+                                                              app.config['PARTY_SERVICE_PASSWORD']))
+
+    if response.status_code == 404:
+        return 'Respondent not found'
```
jcox-dev: It might be better practice to raise a (custom) NotFoundError exception here instead of returning and comparing to a string here.

gemmai95: @jcox-dev this shouldn't be redirecting to the error page it just shouldn't let the person log in

jwgwalton:
I don't like the removing of SERVICE from the urls.

frontstage/controllers/party_controller.py
jcox-dev: @gemmai95 Sorry, I should have been more explicit in the example and said you could raise a (custom) NotFoundError exception in the controller and have the view handle (try/except block) it. (In place of the string comparison).

gemmai95:
[US020: INT Suspend respondents account (L) - Needed for R17](https://trello.com/c/sDXLtTMy/56-us020-int-suspend-respondents-account-l-needed-for-r17)

frontstage/views/sign_in/sign_in.py
```diff
@@ -34,6 +35,12 @@ def login():
             "username": username,
             "password": password
         }
+
+        respondent = party_controller.get_respondent_by_email(username)
+        if not respondent or respondent['status'] != 'ACTIVE':
+            logger.debug("Respondent not able to sign in as don't have an active account in the system.")
+            return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
```
tej99: Should the type be more specific? account locked?

frontstage/views/sign_in/sign_in.py
```diff
@@ -42,13 +49,6 @@ def login():
             error_message = error_json.get('data', {}).get('detail')
             if 'Unauthorized user credentials' in error_message:
                 return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
-            elif 'User account locked' in error_message:
```
tej99: This means users now have no idea wether they can't access their account because it's locked or if they need to verify their account. Quite a big change in the user journey which might need more thinking.

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,15 @@
+import requests
+
+from frontstage import app
+
+
+def get_respondent_by_email(email):
+
```
jcox-dev: Some (debug level) logging in here would be useful as well as error handling and logging of the response, e.g.:

```python
def get_respondent_by_email(email):
    logger.debug('Retrieving respondent by email')
    ...
        logger.debug('Failed to retrieve respondent by email')
        return

    try:
        response.raise_for_status()
        logger.exception('Failed to retrieve respondent by email')
        raise ApiError(response)

    logger.debug('Successfully retrieved respondent by email')
    return response.json()
```

frontstage/views/sign_in/sign_in.py
gemmai95: This is a decision that has been made by IA. We should not be giving them the info of if an account exists in the system.

gemmai95: @tej99 It should now be treated as an invalid login.

gemmai95: @tej99 This is being checked with Dan

frontstage/controllers/party_controller.py
gemmai95: Read my mind haha

emilioward:
Looks good I noticed a few linting errors from running `pipenv check --style . --max-line-length 120` just a heads up

JamesGardiner:
Waiting for Dan from IA. Closing until then.

PR #327, opened 2018-05-09
[TBL] Enrolment to call the services directly

jcox-dev:
Change ras-frontstage to directly call the services directly for enrolment and registration instead of proxying through ras-frontstage-api.

NB: the jenkins / concourse jobs need their config updated, otherwise this'll fail spectacularly!


config.py
```diff
@@ -30,6 +30,31 @@ class Config(object):
...
+    CASE_USERNAME = os.getenv('CASE_USERNAME')
+    CASE_PASSWORD = os.getenv('CASE_PASSWORD')
+    CASE_AUTH = (CASE_USERNAME, CASE_PASSWORD)
+
+    COLLECTION_EXERCISE_URL = os.getenv('COLLECTION_EXERCISE_URL')
+    COLLECTION_EXERCISE_USERNAME = os.getenv('COLLECTION_EXERCISE_USERNAME')
+    COLLECTION_EXERCISE_PASSWORD = os.getenv('COLLECTION_EXERCISE_PASSWORD')
+    COLLECTION_EXERCISE_AUTH = (COLLECTION_EXERCISE_USERNAME, COLLECTION_EXERCISE_PASSWORD)
+
+    IAC_URL = os.getenv('IAC_URL')
```
gemmai95: These need added in docker-dev too.

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,97 @@
...
+            if case_group_id == case.get('caseGroup', {}).get('id'):
+                return case['id']
+
+
+def validate_case_category(category):
+    logger.debug('Validating case category', category=category)
+    categories = get_case_categories()
+    category_names = [cat['name'] for cat in categories]
+    if category not in category_names:
+        raise InvalidCaseCategory(category)
```
gemmai95: Should probably have a success log here.

config.py
jcox-dev: I think they should all now be listed on the confluence page.

gemmai95: They still need to be added to the ras-rm-docker-dev and to the concourse pipeline before adding these changes to master.

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,97 @@
+import logging
+
+from flask import current_app as app
+import requests
+from structlog import wrap_logger
```
JamesGardiner: I've noticed a few different approaches to defining import order. None are right/wrong but we should probably settle on and use it (Black would be useful for this)

frontstage/views/register/activate_account.py
```diff
@@ -15,22 +14,24 @@
...
+        party_controller.verify_email(token)
+    except ApiError as exc:
+        # Handle api errors
+        if exc.status_code == 409:
+            logger.info('Expired email verification token', token=token)
+            return render_template('register/register.link-expired.html')
+        elif exc.status_code == 404:
+            logger.warning('Unrecognised email verification token', token=token)
+            return redirect(url_for('error_bp.not_found_error_page'))
+        elif exc.status_code != 200:
```
JamesGardiner: Should an ApiError ever have a 200 status code?

frontstage/views/register/confirm_organisation_survey.py
```diff
@@ -20,19 +19,31 @@ def register_confirm_organisation_survey():
...
-
-    response_json = json.loads(response.text)
-    logger.info('Successfully retrieved data for confirm organisation/survey page')
+    # Validate enrolment code before retrieving organisation data
+    iac_controller.validate_enrolment_code(enrolment_code)
+
+    logger.info('Attempting to retrieve data for confirm organisation/survey page',
+                enrolment_code=enrolment_code)
+    try:
+        # Get organisation name
```
JamesGardiner: Try blocks should be limited in their scope to isolate the error succinctly. This block has too much logic in it IMHO.

jcox-dev: I'm also a fan of limiting a try/except's scope but any one of the function calls within could raise an `ApiError` (when using one of the other services), so rather than blocking out each call I thought it made sense to capture and fail hard at any point during the confirmation process.

frontstage/views/register/activate_account.py
jcox-dev: Will change to an `else`.

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,98 @@
...
+        log_level = logger.warning if response.status_code == 404 else logger.exception
+        log_level('Failed to retrieve case', enrolment_code=enrolment_code)
+        raise ApiError(response)
+
+    logger.debug('Successfully retrieved case by enrolment code', enrolment_code=enrolment_code)
+    return response.json()
+
+
+def get_case_by_party_id(party_id, case_events=False):
+    logger.debug('Attempting to retrieve case by party id', party_id=party_id)
```
gemmai95: Does this not return a list of cases for the party id rather than just a singular case? This should be reflected in the logs.

jcox-dev:
Merged into #333.

PR #328, opened 2018-05-10
Us110 display trading as

rhiGriff:
Displaying trading as only if there is a value.

### DIFF
--- a/frontstage/templates/surveys/surveys.html
+++ b/frontstage/templates/surveys/surveys.html
@@ -32,12 +32,12 @@
             <span class="survey-list-item__label mercury">SURVEY: </span>
             <span id="SURVEY_NAME">{{ survey.survey.longName }}</span><br/>
             <span id="REPORTING_UNIT_DETAILS_{{ survey.case.caseGroup.sampleUnitRef }}" class="survey-list-item pluto">
-                        Business: {{ survey.business_party.name }}<br/>
-                        {% if survey.business_party.trading_as %}
-                        Trading as: {{survey.business_party.trading_as }}<br/>
-                        {% endif %}
-                        Ref: {{survey.case.caseGroup.sampleUnitRef }}
-                    </span>
+                Business: {{ survey.business_party.name }}<br/>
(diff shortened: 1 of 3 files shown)


rhiGriff:
[US110: EXT Display "trading as" in To do and History [M]](https://trello.com/c/pXNYDmoS/204-us110-ext-display-trading-as-in-to-do-and-history-m)

tests/app/test_surveys.py
```diff
@@ -70,6 +73,42 @@ def test_surveys_todo(self, mock_request):
         self.assertTrue('Business Register and Employment Survey'.encode() in response.data)
         self.assertTrue('RUNAME1_COMPANY4 RUNNAME2_COMPANY4'.encode() in response.data)
 
+    @requests_mock.mock()
```
JamesGardiner: What happens if the '/surveys/todo` endpoint returns an error?

rhiGriff: This should be covered in another test as these tests are only checking for the trading as and not if the page can be displayed.

frontstage/templates/surveys/surveys.html
```diff
@@ -31,9 +31,12 @@
...
             <span id="SURVEY_NAME">{{ survey.survey.longName }}</span><br/>
-            <span class="survey-list-item pluto">
-                        BUSINESS: {{ survey.business_party.name }}<br/>
-                        REF: {{survey.case.caseGroup.sampleUnitRef }}
+            <span id="REPORTING_UNIT_DETAILS_{{ survey.case.caseGroup.sampleUnitRef }}" class="survey-list-item pluto">
+                        Business: {{ survey.business_party.name }}<br/>
+                        {% if survey.business_party.trading_as %}
+                        Trading as: {{survey.business_party.trading_as }}<br/>
+                        {% endif %}
+                        Ref: {{survey.case.caseGroup.sampleUnitRef }}
```
emilioward: Can we fix the indentation spacing on this please

PR #329, opened 2018-05-11
[TBL] Call the services directly to list the surveys for a respondent

jcox-dev:
Change ras-frontstage to directly call the services directly for listing surveys instead of proxying through ras-frontstage-api.

NB: the jenkins / concourse jobs need their config updated, otherwise this'll fail spectacularly!


tests/app/mocked_services.py
```diff
@@ -0,0 +1,45 @@
+import json
+
+from frontstage import app
+
+
+with open('tests/test_data/party/business_party.json') as json_data:
```
rhiGriff: Shouldn't be using as json_data here it should be something like as fp or as f

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,104 @@
...
+        "collection_instrument_type": collection_instrument['type'],
+        "collection_instrument_size": collection_instrument['len'],
+        "go_live": collection_exercise_go_live
+    }
+    logger.debug('Successfully built case data', case_id=case['id'])
+    return survey_data
+
+
+def calculate_case_status(case, collection_instrument_type):
+    logger.debug('Getting the status of case')
```
gemmai95: Can we log out the case id here if possible?

frontstage/controllers/party_controller.py
```diff
@@ -0,0 +1,35 @@
...
+
+from frontstage.exceptions.exceptions import ApiError
+
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+def get_party_by_business_id(party_id, collection_exercise_id=None):
+    logger.debug('Attempting to retrieve party by business',
+                 collection_exercise_id=collection_exercise_id,
```
gemmai95: Don't think we need collection instrument id logged out here

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,104 @@
...
+        status = 'Downloaded'
+
+    logger.debug('Retrieved the status of case', case=case['id'], status=status)
+    return status
+
+
+def format_collection_exercise_dates(collection_exercise):
+    logger.debug('Formatting collection exercise dates')
+    input_date_format = 'YYYY-MM-DDThh:mm:ss'
+    output_date_format = 'D MMM YYYY'
```
emilioward: is that meant to be DD MM YYYY?

PR #331, opened 2018-05-14
Refactor to send directly to secure message api

dcdarrell9:
Removed usage of frontstage-api calls to secure-message and call directly

### DIFF
--- a/frontstage/controllers/conversation_controller.py
+++ b/frontstage/controllers/conversation_controller.py
@@ -1,3 +1,4 @@
+import json
 from json import JSONDecodeError
 import logging
 
@@ -66,7 +67,8 @@ def get_conversation_list():
 
 
 def send_message(message_json):
-    logger.info("About to send message")
+    party_id = json.loads(message_json).get('msg_from')
+    logger.info("About to send message", party_id=party_id)
     url = '{}v2/messages'.format(current_app.config["RAS_SECURE_MESSAGING_SERVICE"])
     headers = _create_send_message_headers()
 
@@ -74,9 +76,10 @@ def send_message(message_json):
 
     try:
         response.raise_for_status()
+        logger.info("Send message", party_id=party_id)
         return response.json()
     except HTTPError:
-        logger.exception("Message sending failed due to API Error")
+        logger.exception("Message sending failed due to API Error", party_id=party_id)
         raise ApiError(response)
 
 
--- a/frontstage/controllers/conversation_controller.py
+++ b/frontstage/controllers/conversation_controller.py
@@ -75,9 +75,9 @@ def send_message(message_json):
         response.raise_for_status()
         logger.info("Message sent successfully")
         return response.json()
-    except HTTPError as ex:
+    except HTTPError:
         logger.exception("Message sending failed due to API Error")
-        raise ApiError(ex.response)
+        raise ApiError(response)
 
 
 def _create_get_conversation_headers():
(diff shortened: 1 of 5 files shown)


dcdarrell9:
[Refactor frontstage to send messages using the secure-message API directly (5)](https://trello.com/c/DGyASdWF/177-refactor-frontstage-to-send-messages-using-the-secure-message-api-directly-5)

frontstage/views/secure_messaging/create_message.py
```diff
@@ -8,6 +8,7 @@
 
 from frontstage.common.api_call import api_call
 from frontstage.common.session import SessionHandler
```
MIslam92: Forgot to remove these unused imports?

frontstage/controllers/conversation_controller.py
```diff
@@ -1,3 +1,4 @@
+import json
```
MIslam92: This import is not used either?

frontstage/views/secure_messaging/create_message.py
```diff
@@ -8,6 +8,7 @@
 
 from frontstage.common.api_call import api_call
```
MIslam92: Unused import

frontstage/views/secure_messaging/create_message.py
```diff
@@ -8,6 +8,7 @@
...
 from frontstage.common.session import SessionHandler
+from frontstage.controllers import conversation_controller
 from frontstage.exceptions.exceptions import ApiError
```
MIslam92: Unused import again

dcdarrell9:
The `make test` must not be running the lint, my bad, I'll see if I can get it running with that

frontstage/controllers/conversation_controller.py
```diff
@@ -72,10 +72,11 @@ def send_message(message_json):
...
         response = session.post(url, headers=headers, data=message_json)
         response.raise_for_status()
-    except HTTPError as ex:
+        logger.info("Message sent successfully")
+        return response.json()
+    except HTTPError:
         logger.exception("Message sending failed due to API Error")
-        raise ApiError(ex.response)
-    logger.info("Message sent successfully")
+        raise ApiError(response)
```
MIslam92: `raise ApiError(response)` throws the classic `Local variable might be referenced before assignment` warning so might be worth setting response to none prior to the try/except?

insacuri: I think you can put the post outside the try/except to fix the issue.  I think that's what we do in response-operations-ui but you'll have to double check

frontstage/controllers/conversation_controller.py
```diff
@@ -72,10 +72,11 @@ def send_message(message_json):
     try:
         response = session.post(url, headers=headers, data=message_json)
         response.raise_for_status()
-    except HTTPError as ex:
+        logger.info("Message sent successfully")
```
insacuri: Could you put a ```python

dcdarrell9: Line 64 in create_message.py gives us the message id and party id, so I think that should be fine 👍

insacuri: It's not good enough, lots of the time when you're debugging an issue and looking through logs, you filter on an ID to begin with, see what messages you get then go from there.  And we want to make that part as easy as possible whereever we can

Think about in SDX, pretty much all of the lines have a tx_id on them and when you search in splunk, it's really easy to see the whole flow.  If you want anything that hasn't got that id then it's more of a pain to find generally

MIslam92:
@dcdarrell9 Squash and merge?

PR #332, opened 2018-05-14
[TBL] Generate the EQ URL within the frontstage service

jcox-dev:
Change ras-frontstage to directly call the services directly for generating a URL for EQ instead of proxying through ras-frontstage-api.

NB: this replicates a lot of the functionality already in https://github.com/ONSdigital/ras-frontstage/pull/323 for accessing case.

NB: the jenkins jobs / concourse need their config updated, otherwise this'll fail spectacularly!


frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,194 @@
...
+def get_case_by_case_id(case_id):
+    logger.debug('Retrieving case', case_id=case_id)
+    url = f"{app.config['CASE_URL']}/cases/{case_id}"
+    response = requests.get(url, auth=app.config['CASE_AUTH'])
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        log_level = logger.warning if response.status_code == 404 else logger.exception
+        log_level('Failed to retrieve case', case_id=case_id)
```
rhiGriff: Can we add the status code here so it's visible in Splunk please?

frontstage/controllers/case_controller.py
```diff
@@ -0,0 +1,194 @@
...
+def get_case_categories():
+    logger.debug('Retrieving case categories')
+    url = f"{app.config['CASE_URL']}/categories"
+    response = requests.get(url, auth=app.config['CASE_AUTH'])
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        log_level = logger.warning if response.status_code == 404 else logger.exception
+        log_level('Failed to get case categories')
```
rhiGriff: Same here as well please to log the status code.

frontstage/controllers/eq_payload.py
```diff
@@ -0,0 +1,120 @@
+import logging
```
insacuri: Doesn't seem right to me that this mixed in with controllers as it's more of a model if we're still using the MVC analogy.  I've seen other places put it in common/ but I don't know if that's right either...
I won't hold up the PR on this, but if you feel the same way then maybe move it somewhere more sensible?

config.py
```diff
@@ -83,3 +135,6 @@ class TestingConfig(DevelopmentConfig):
     os.environ['OAUTHLIB_INSECURE_TRANSPORT'] = '1'
     JWT_SECRET = 'testsecret'
     REDIS_DB = os.getenv('REDIS_DB', 13)
+    JSON_SECRET_KEYS = open("./tests/test_data/jwt-test-keys/test_key.json").read()
+    EQ_URL = 'https://eq-test/session?token='
+    ACCOUNT_SERVICE_URL = 'http://frontstage-url/surveys'
```
tej99: 'http://localhost:8082/surveys/todo'?

config.py
```diff
@@ -75,6 +102,31 @@ class DevelopmentConfig(Config):
...
+
+    PARTY_URL = os.getenv('PARTY_URL', 'http://localhost:8081')
+    PARTY_USERNAME = os.getenv('PARTY_USERNAME', 'admin')
+    PARTY_PASSWORD = os.getenv('PARTY_PASSWORD', 'secret')
+    PARTY_AUTH = (PARTY_USERNAME, PARTY_PASSWORD)
+
+    SURVEY_URL = os.getenv('SURVEY_URL', 'http://localhost:8080')
+    SURVEY_USERNAME = os.getenv('SURVEY_USERNAME', 'admin')
+    SURVEY_PASSWORD = os.getenv('SURVEY_PASSWORD', 'secret')
+    SURVEY_AUTH = (SURVEY_USERNAME, SURVEY_PASSWORD)
```
tej99: Might be useful to add default values for EQ_URL and JSON_SECRET_KEYS for the DevelopmentConfig

jcox-dev: @tej99 any suggestion for the JSON_SECRET_KEYS default?

frontstage/common/eq_payload.py
```diff
@@ -0,0 +1,120 @@
+import logging
+import time
+import uuid
+
+import iso8601
+from flask import current_app
+from frontstage.controllers import (collection_exercise_controller, collection_instrument_controller,
```
gemmai95: Can we fix this import order please? In app imports shouldn't be with flask imports

tests/app/mocked_services.py
```diff
@@ -0,0 +1,53 @@
...
+with open('tests/test_data/case/case.json') as fp:
+    case = json.load(fp)
+
+with open('tests/test_data/case/categories.json') as json_data:
+    categories = json.load(json_data)
+
+with open('tests/test_data/collection_exercise/collection_exercise.json') as fp:
+    collection_exercise = json.load(fp)
+
+with open('tests/test_data/collection_exercise/collection_exercise_before_go_live.json') as json_data:
```
gemmai95: Any reason this is still json_data rather than fp?

jcox-dev:
Merged into #333.

PR #342, opened 2018-05-24
Upping to v0.4.2

pricem14pc:


config.py
```diff
@@ -6,7 +6,7 @@
 class Config(object):
     DEBUG = False
     TESTING = False
-    VERSION = '0.4.0'
+    VERSION = '0.4.2'
```
JamesGardiner: What happened to 0.4.1?

pricem14pc: Its on a branch

JamesGardiner: 👍

PR #344, opened 2018-05-25
Fix secure message usage of requests session

benjefferies:
# Background
ras-frontstage uses Session (http://docs.python-requests.org/en/master/user/advanced/) to get retry functionality if a HTTP request fails. When a Session is created the TCP connection is continued to be reused which leads to long running/open TCP connections. Cloundfoundry may do some housekeeping and close long running TCP connections incase they are connection that haven't correctly been closed. If that connection is closed it should reconnect but we're not experiencing this behaviour gracefully when running the services in CF. (https://docs.cloudfoundry.org/services/route-services.html)

All of the above was causing us to experience 500 errors occasionally when sending messages.

# Whats changed
Instead of creating one session at the beginning when the application starts we will now create a new session for each request.

# How to review/test
* Existing tests should pass
* Deploy this branch to cloudfoundry and check no 500 errors occur. I have tested this re-running the same test for over and hour without failure
* Deploy this branch to cloudfoundry and kicked off the secure message acceptance tests builds 12-31 to check the stability of the tests with an audit

### DIFF
--- a/frontstage/controllers/conversation_controller.py
+++ b/frontstage/controllers/conversation_controller.py
@@ -16,7 +16,7 @@ from requests.exceptions import HTTPError, RequestException
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-def get_session():
+def _get_session():
     session = requests.Session()
     retries = Retry(total=10, backoff_factor=0.1)
     session.mount('http://', HTTPAdapter(max_retries=retries))
@@ -30,10 +30,12 @@ def get_conversation(thread_id):
     headers = _create_get_conversation_headers()
     url = '{}v2/threads/{}'.format(current_app.config["RAS_SECURE_MESSAGING_SERVICE"], thread_id)
 
+    with _get_session() as session:
+        response = session.get(url, headers=headers)
+
     try:
-        response = get_session().get(url, headers=headers)
         response.raise_for_status()
-    except (HTTPError, RequestException):
+    except RequestException:
         logger.exception("Thread retrieval failed", thread_id=thread_id)
         raise ApiError(response)
     logger.info("Thread retrieval successful", thread_id=thread_id)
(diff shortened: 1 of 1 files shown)


frontstage/controllers/conversation_controller.py
```diff
@@ -30,7 +31,7 @@ def get_conversation(thread_id):
     url = '{}v2/threads/{}'.format(current_app.config["RAS_SECURE_MESSAGING_SERVICE"], thread_id)
 
     try:
-        response = session.get(url, headers=headers)
+        response = get_session().get(url, headers=headers)
```
benjefferies: The documentation suggests Session() should be used as a context manager to ensure the connection is closed after the request i.e. `with get_session() as s:` http://docs.python-requests.org/en/master/user/advanced/

jcox-dev: I expect we could also decorate the session-based functions with `@get_session`, (after tweaking it a bit) which would then inject the session variable as a parameter to be used throughout.

benjefferies: Fair point @jcox-dev. I find it's a little harder to follow when there is a decorator that injects variables

frontstage/controllers/conversation_controller.py
```diff
@@ -16,11 +16,12 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-# Configure number of retries on requests
-session = requests.Session()
-retries = Retry(total=10, backoff_factor=0.1)
-session.mount('http://', HTTPAdapter(max_retries=retries))
-session.mount('https://', HTTPAdapter(max_retries=retries))
+def _get_session():
```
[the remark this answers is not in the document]
jcox-dev: Ignore. Session variable that is returned by `_get_session()` is a context manager.

PR #351, opened 2018-06-11
Remove redis dump and add to gitignore

gemmai95:
# Motivation and Context

Removes accidentally committed file and stops files of this type being added in the future.

# What has changed

Remove dump.rdb and add to git ignore.

# How to test?

# Links


.gitignore
```diff
@@ -13,6 +13,9 @@ __pycache__/
 # C extensions
 *.so
 
+# Redis files
+dump.rdb
```
JamesGardiner: Should probably be `*.rdb`

PR #358, opened 2018-06-25
Output time to match BST timezone in secure message

shahi645:
# Motivation and Context
The Business have been informed that the time stamp for secure messaging doesn't reflect BST.

# What has changed
Time displayed for users in secure message is BST

# How to test?
Run RAS/RM, send a message to an internal user, the message should display the correct time

### DIFF
--- a/frontstage/common/message_helper.py
+++ b/frontstage/common/message_helper.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import datetime, date
+from dateutil.tz import tz
 
 from structlog import wrap_logger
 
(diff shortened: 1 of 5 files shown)


frontstage/common/message_helper.py
```diff
@@ -81,3 +82,16 @@ def get_formatted_date(datetime_string, string_format='%Y-%m-%d %H:%M:%S'):
     elif time_difference.days == -1:
         return "Yesterday at {}".format(datetime_parsed.strftime('%H:%M'))
     return "{} {}".format(datetime_parsed.strftime('%d %b %Y').title(), datetime_parsed.strftime('%H:%M'))
+
+
+def get_current_date_bst(datetime_string):
```
dcdarrell9: Only thing with this is that it will work for respondents in the uk, but not in other places with different timezones

shahi645: good point, could add an if statement to determine locale, where this is called, not sure on the requirement though. Ticket only stated that the displayed time should be BST

PR #363, opened 2018-07-09
Add validation for enrolment code length

shahi645:
# Motivation and Context
Currently when a user enters an enrolment code longer or shorter than 12 character, nothing happens. This introduces validation on that error. Also we were counting leading/trailing white spaces as characters, so this PR stops that also.

# How to test?
Run Frontstage enter an enrolment code too long/short and view validation error, try with spaces they shouldn't count

### DIFF
--- a/frontstage/models.py
+++ b/frontstage/models.py
@@ -14,35 +14,37 @@ logger = wrap_logger(logging.getLogger(__name__))
 
 
 class EnrolmentCodeForm(FlaskForm):
-    enrolment_code = StringField('Enrolment Code', [InputRequired(), Length(min=12, max=12,
-                                                                            message='Your code must be 12 characters')])
+    enrolment_code = StringField('Enrolment Code', [InputRequired(), Length(min=12,
+                                                                            max=12,
+                                                                            message='Please re-enter '
+                                                                                    'the code and try again')])
 
 
 class RegistrationForm(FlaskForm):
     first_name = StringField('First name',
-                             validators=[InputRequired("First name is required"),
+                             validators=[InputRequired('First name is required'),
                                          Length(max=254,
                                                 message='Your first name must be less than 254 characters')])
     last_name = StringField('Last name',
-                            validators=[InputRequired("Last name is required"),
+                            validators=[InputRequired('Last name is required'),
                                         Length(max=254, message='Your last name must be less than 254 characters')])
     email_address = StringField('Enter your email address',
-                                validators=[InputRequired("Email address is required"),
+                                validators=[InputRequired('Email address is required'),
                                             Email(message="Your email should be of the form 'myname@email.com' "),
                                             Length(max=254,
                                                    message='Your email must be less than 254 characters')])
     password = PasswordField('Create a password',
-                             validators=[DataRequired("Password is required"),
+                             validators=[DataRequired('Password is required'),
                                          EqualTo('password_confirm', message=app.config['PASSWORD_MATCH_ERROR_TEXT']),
                                          Length(min=app.config['PASSWORD_MIN_LENGTH'],
                                                 max=app.config['PASSWORD_MAX_LENGTH'],
                                                 message=app.config['PASSWORD_CRITERIA_ERROR_TEXT'])])
     password_confirm = PasswordField('Re-type your password')
     phone_number = StringField('Enter your phone number',
-                               validators=[DataRequired("Phone number is required"),
+                               validators=[DataRequired('Phone number is required'),
                                            Length(min=9,
                                                   max=15,
-                                                  message="This should be a valid phone number between 9 and 15 digits")],
+                                                  message='This should be a valid phone number between 9 and 15 digits')],
                                default=None)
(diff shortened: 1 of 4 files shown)


frontstage/templates/errors/validation_error.html
```diff
@@ -0,0 +1,13 @@
+{% block main %}
```
andrewmil: 👍  for having this as a separate component
Would there we a way to make it more general so we can use it for the rest of our validation too and just pass in variables; `title`, `message` `focus_on` etc

frontstage/templates/register/register.enter-enrolment-code.html
```diff
@@ -30,7 +29,7 @@ <h1 class="panel__title venus">Enrolment code not valid</h1>
 
         <h1 class="saturn">Create an account</h1>
 
-        {% if errorType == "failed" %}
+        {% if errorType == "failed" or errors|length > 0 %}
```
andrewmil: Is there any way to combine `errorType` and `errors` would be nice if all our errors got passed in in the same way

shahi645: I agree with this, although i have no idea where the errorType comes from, will take a look into this

frontstage/models.py
```diff
@@ -14,43 +14,45 @@
...
+                                                  message='This should be a valid phone number between 9 and 15 digits')],
                                default=None)
     enrolment_code = HiddenField('Enrolment Code')
 
     @staticmethod
     def validate_phone_number(form, field):
         try:
             logger.debug('Checking this is a valid GB phone number')
-            input_number = phonenumbers.parse(field.data, "GB")  # Tell the parser we are looking for a GB number
+            input_number = phonenumbers.parse(field.data, 'GB')  # Tell the parser we are looking for a GB number
```
gavinleeedwards: Is there a reason why we're only accepting GB numbers when a lot of respondents are overseas?

frontstage/views/register/create_account.py
```diff
@@ -18,6 +18,8 @@
 def register():
     cryptographer = Cryptographer()
     form = EnrolmentCodeForm(request.form)
+    if form.enrolment_code.data:
+        form.enrolment_code.data = form.enrolment_code.data.strip()
```
damien-lloyd: Would you explain why we're checking for form enrolment_code data then assigning new data to the the thing we were checking against?

shahi645: just stripping leading/trailing whitespaces

frontstage/models.py
```diff
@@ -65,9 +67,9 @@ def validate_phone_number(form, field):
     def validate_email_address(form, field):
         logger.debug('Checking if the email address contains a space')
         # this extends the email validator to check if there is whitespace in the email
-        if " " in field.data:
+        if ' ' in field.data:
             logger.debug('Space found in email address')
-            raise ValidationError("Your email should be of the form myname@email.com")
+            raise ValidationError('Your email should be of the form myname@email.com')
```
gavinleeedwards: Could the field be trimmed of any spaces at the beginning/end before being checking for spaces within the address?

frontstage/models.py
shahi645: I'm not sure, theres a car for this work, so probably look to add it on there:

shahi645: Ok looking like we are going to remove this added validation, so just to ignore this @gavinleeedwards

frontstage/models.py
shahi645: yes we can do it, probably in the scope of a different ticket though

frontstage/models.py
gavinleeedwards: All good if these bits are captured on other tickets :+1:

frontstage/templates/errors/validation_error.html
shahi645: done

frontstage/templates/register/register.enter-enrolment-code.html
shahi645: So currently not possible to do without putting the IAC checks in the form. Spoke to Lewis, he's keen on keeping the two things separate

frontstage/views/register/create_account.py
andrewmil: ^ And we check if it exists because you can't perform the `.strip()` on `None`

frontstage/templates/errors/validation_error.html
```diff
@@ -0,0 +1,13 @@
...
+
+<div class="panel panel--error">
+    <div class="panel__header">
+        <h1 class="panel__title venus">{{ panel_title }}</h1>
+    </div>
+    <div class="panel__body" data-qa="error-body">
+        <p class="mars"><a onclick="focusOn({{ focus_on }});" href="#enrolment_code">{{ error_message }}</a></p>
```
gavinleeedwards: Should the anchor be passed in as well if this is to be generic and reusable?

PR #365, opened 2018-07-12
Log Exception traces as JSON in Frontstage

MIslam92:
Currently in Splunk the errors are displaying 'incorrectly' now with the updated logging, it should appear much more readable in Splunk.

Card: https://trello.com/c/CYTY3NKF/229-log-exception-traces-as-json-in-frontstage-s

### How to review 

- Check when you get an exception if it outputs correctly in the log as a single JSON block.
- Run tests the usual.

### DIFF
--- a/frontstage/common/message_helper.py
+++ b/frontstage/common/message_helper.py
@@ -26,7 +26,7 @@ def get_message_subject(message):
         subject = message["subject"]
         return subject
     except KeyError:
-        logger.error("Failed to retrieve Subject from thread")
+        logger.error('Failed to retrieve Subject from thread')
         raise
 
 
@@ -37,7 +37,7 @@ def get_from_name(message):
     try:
         msg_from = message['@msg_from']
     except KeyError:
-        logger.error("Failed to retrieve name from message", message_id=message.get('msg_id'))
+        logger.error('Failed to retrieve name from message', message_id=message.get('msg_id'))
         raise
 
     return "{} {}".format(msg_from.get('firstName'), msg_from.get('lastName'))
@@ -47,7 +47,7 @@ def get_ru_ref_from_message(message):
     try:
         return message['@ru_id']['id']
     except (KeyError, TypeError):
-        logger.error("Failed to retrieve RU ref from message", message_id=message.get('msg_id'))
+        logger.error('Failed to retrieve RU ref from message', message_id=message.get('msg_id'))
         raise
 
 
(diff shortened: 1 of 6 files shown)


frontstage/common/message_helper.py
```diff
@@ -26,7 +26,7 @@ def get_message_subject(message):
         subject = message["subject"]
         return subject
     except KeyError:
-        logger.exception("Failed to retrieve Subject from thread")
+        logger.error("Failed to retrieve Subject from thread")
```
shahi645: 👍

frontstage/views/secure_messaging/message_get.py
```diff
@@ -28,7 +28,7 @@ def view_conversation(session, thread_id):
     try:
         refined_conversation = [refine(message) for message in reversed(conversation['messages'])]
     except KeyError as e:
-        logger.exception("Message is missing important data", thread_id=thread_id, party_id=party_id)
+        logger.error("Message is missing important data", thread_id=thread_id, party_id=party_id)
```
shahi645: these should be single quotes but we dont enforce it so 🤷‍♂️

MIslam92: Might as well do it now 🤷‍♀️

frontstage/error_handlers.py
```diff
@@ -39,15 +39,17 @@ def connection_error(error):
...
-def jwt_validation_error(error):  # pylint: disable=unused-argument
+def jwt_validation_error(error):
+    del error
     return redirect(url_for('error_bp.not_logged_in_error_page',
                             _external=True,
                             _scheme=getenv('SCHEME', 'http')))
 
 
 @app.errorhandler(Exception)
 def server_error(error):  # pylint: disable=unused-argument
```
atrozzo: Not sure but , I think the exception details will get lost in the log using error . The use of error for such checked exception i think would be more appropriate as warning level. They are actually not real errors . 
If you were actually parsing the raising exception here, you would not provably need to change the logging from exception to error. Plus tell if I am wrong but the handler will handle just the one are rised, so no need to change where the exception is not raised.

frontstage/controllers/conversation_controller.py
```diff
@@ -130,6 +130,6 @@ def remove_unread_label(message_id):
         try:
             response.raise_for_status()
         except HTTPError:
-            logger.exception('Failed to remove unread label', message_id=message_id, status=response.status_code)
+            logger.error('Failed to remove unread label', message_id=message_id, status=response.status_code)
```
atrozzo: not sure you need to change it here if you do not raise teh exception .

frontstage/controllers/conversation_controller.py
```diff
@@ -36,7 +36,7 @@ def get_conversation(thread_id):
             response.raise_for_status()
         except HTTPError:
             raise ApiError(logger, response,
-                           log_level='exception',
+                           log_level='error',
```
shahi645: good catch @gavinleeedwards 👍

PR #367, opened 2018-07-16
Stop frontstage redirecting to error pages

shahi645:
# Motivation and Context
Currently when an error is generated in frontstage the user is redirected to an error page e.g. `/errors/500/`. This is unnecessary and causes issues:
 - It's harder to track what went wrong if a respondent sends us the URL 
 - It's harder to debug issues as it's not straightforward to link the error page request with the failure
 - It stops the user from refreshing the page to retry
 - The error page itself can be requested, so errors can be generated by anyone which could skew metrics

# What has changed
Instead of redirecting to an endpoint, the correct template is just rendered leaving url untouched. We now also log the url and status code on error.

# How to test?
- Run Frontstage and recreate errors from error handlers, by stopping other services, e.g. `OAuth` for `sign in`.
- Does the correct url remain in place and also get logged out with status code
- Also run acceptance_tests


frontstage/views/passwords/reset_password.py
```diff
@@ -20,11 +20,11 @@ def get_reset_password(token, form_errors=None):
...
     except ApiError as exc:
         if exc.status_code == 409:
-            logger.warning('Token expired', status_code=exc.status_code, token=token)
+            logger.warning('Token expired', url=exc.url, status_code=exc.status_code, token=token)
             return render_template('passwords/password-expired.html')
         elif exc.status_code == 404:
-            logger.warning('Invalid token sent to party service', status_code=exc.status_code, token=token)
-            return redirect(url_for('error_bp.not_found_error_page'))
+            logger.warning('Invalid token sent to party service', url=exc.url, status_code=exc.status_code, token=token)
+            return render_template('errors/404-error.html'), 404
```
ajmaddaford: Instead of calling `render_template` we should call `abort(404)` to use the common error handler so we're not duplicating the template reference.

frontstage/views/register/activate_account.py
```diff
@@ -25,7 +25,7 @@ def register_activate_account(token):
             return render_template('register/register.link-expired.html')
         elif exc.status_code == 404:
             logger.warning('Unrecognised email verification token', token=token)
```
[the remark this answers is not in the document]
shahi645: Good catch can do here, so will add 👍

frontstage/error_handlers.py
```diff
@@ -14,49 +13,38 @@
...
+    return render_template('errors/500-error.html'), 500
 
 
 @app.errorhandler(InvalidEqPayLoad)
 def eq_error(error):
     logger.error('Failed to generate EQ URL', error=error.message)
-    return redirect(url_for('error_bp.server_error_page',
-                            _external=True,
-                            _scheme=getenv('SCHEME', 'http')))
+    return render_template('errors/500-error.html'), 500
```
ajmaddaford: Every error needs to log the status code and url. I'd say the status code should be logged as `status_code` rather than just `status` as that's what's done in eQ (see https://github.com/ONSdigital/eq-survey-runner/blob/master/app/views/errors.py#L71) and it would be good to get some consistency across the platform.

frontstage/views/register/activate_account.py
```diff
@@ -21,13 +21,14 @@ def register_activate_account(token):
     except ApiError as exc:
         # Handle api errors
         if exc.status_code == 409:
-            logger.info('Expired email verification token', token=token)
+            logger.info('Expired email verification token', token=token, url=exc.url, status_code=exc.status_code)
```
ajmaddaford: I think we should use use `api_url` and `api_status_code` in any place where we're logging the url and status code of the API we're calling. This will allow us to easily differentiate between this services errors and the API errors in the logs.

shahi645:
@jcox-dev Theres a ticket for it, we just separated it...so it coming up! 😉

PR #369, opened 2018-07-19
Change logging levels where appropriate

shahi645:
# Motivation and Context
We had some logging in front stage that was error level when they shouldn't have been. This was leading to splunk alerts populating our channel when these events occured.

# What has changed
- changed from error/warning to info

- deleted unused templates

- refactor for linting

# How to test?
Check out logging for some of the events, run tests and acceptance tests


frontstage/controllers/iac_controller.py
```diff
@@ -31,7 +31,7 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
     if validate and not response.json().get('active', False):
         raise ApiError(logger, response, message='Invalid enrolment code used')
```
gavinleeedwards: The 2 `Invalid enrolment code used` messages shouldn't really be logged as errors

shahi645: ahh yes nice! 👍

PR #372, opened 2018-07-24
Fix invalid use of logger

### LINKED ISSUE #371: Invalid use of logger
A recent pull request (#369) has introduced some invalid uses of logger.

1. In iac_controller.py [line 25](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/controllers/iac_controller.py#L25) and [line 32](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/controllers/iac_controller.py#L32 ) `logger.info` is passed to `ApiError` rather than `logger`. `log_level` should have been passed as an argument instead: 

```python
raise ApiError(logger.info, response, log_level='info', message='Invalid enrolment code used')
```

2. In forgot_password.py [line 41](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/views/passwords/forgot_password.py#L41) and [line 45](https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/views/passwords/forgot_password.py#L45) `log_level` is passed as a parameter to `exc.logger`, leading to `log_level` and `level` being output in the log line. This is because `exc.logger` is a proxy to whatever log method was chosen when the exception was raised, meaning that `log_level` becomes just another property in the log output.

shahi645:
# Motivation and Context
A recent change has introduced some invalid use of logging, concerning the log level

# What has changed
log level changed to correctly output log level

# How to test?
Raise logging found:
 - In iac_controller.py line 25 and line 32 
 - In forgot_password.py line 41 and line 45

fixes #371

### DIFF
--- a/frontstage/controllers/iac_controller.py
+++ b/frontstage/controllers/iac_controller.py
@@ -22,14 +22,14 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
         if response.status_code == 404:
             logger.info('IAC not found')
             if validate:
-                raise ApiError(logger.info, response, message='Invalid enrolment code used')
+                raise ApiError(logger, response, log_level='info', message='Invalid enrolment code used')
             return
         # 401s may include error context in the JSON response
         elif response.status_code != 401:
             raise ApiError(logger, response, message='Failed to retrieve IAC')
 
     if validate and not response.json().get('active', False):
-        raise ApiError(logger.info, response, message='Invalid enrolment code used')
+        raise ApiError(logger, response, log_level='info', message='Invalid enrolment code used')
 
     logger.info('Successfully retrieved IAC')
     return response.json()
--- a/frontstage/controllers/iac_controller.py
+++ b/frontstage/controllers/iac_controller.py
@@ -22,14 +22,14 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
         if response.status_code == 404:
             logger.info('IAC not found')
             if validate:
-                raise ApiError(logger, response, log_level='info', message='Invalid enrolment code used')
+                raise ApiError(logger, response, message='Invalid enrolment code used')
             return
         # 401s may include error context in the JSON response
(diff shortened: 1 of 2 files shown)


frontstage/controllers/iac_controller.py
```diff
@@ -22,14 +22,14 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
         if response.status_code == 404:
             logger.info('IAC not found')
             if validate:
-                raise ApiError(logger.info, response, message='Invalid enrolment code used')
+                raise ApiError(logger, response, log_level='info', message='Invalid enrolment code used')
```
benjefferies: This seems wrong. It feels like log_level shouldn't be a part of the exception. It should be a new type of Exception (that could extend APiError)

jcox-dev: @benjefferies perhaps but not relevant to this PR. It was refactored this way some time ago when we deprecated frontstage-api.

shahi645: @benjefferies @jcox-dev Improvements team have raised a ticket to refactor logging and specifically this 👍

shahi645: this was just to fix a bug introduced

ajmaddaford: The log level isn't being used, when the ApiError is handled in error_handlers.py it doesn't use the logger passed via the exception, it just uses uses `logger.error`. I think we should be discussing why we generate a 500 error when an enrolment code is not found or is not active.

shahi645: That's a valid point @ajmaddaford, in terms of this piece of work, how should I look to rectify this.

ajmaddaford: I think we should continue to log at error as we're returning a 500.

shahi645:
@jcox-dev @ajmaddaford Ready for re-review 😉

PR #373, opened 2018-07-26
Defect - Forgot password has incorrect message when user submits an unregistered email address

MIslam92:
Current defect where the forgot password page has an incorrect message displayed to the user when they submit an unregistered email address

Card: https://trello.com/c/M8KCuVwm/235-ext-defect-us012-forgot-password-has-incorrect-message-when-user-submits-an-unregistered-email-address-m

### How to review
Check that the correct content is displayed on the password reset page as well as validation when you enter a non valid email address as well on the forgot password screen.

### DIFF
--- a/frontstage/templates/passwords/forgot-password.check-email.html
+++ b/frontstage/templates/passwords/forgot-password.check-email.html
@@ -5,10 +5,9 @@
 
 {% block main %}
 
-    <h1 class="saturn">Check your email</h1>
+    <h1 class="saturn">Password reset request sent</h1>
 
-    <p>If your account is verified, you will receive an email at the registered email address.</p>
-    <p>Please follow the link in the email to reset your password.</p>
+    <p>If {{ email }} is registered to an account, you'll receive an email with instructions to reset your password</p>
 
     <div class="details js-details" data-guidance-label="Help with email" data-guidance="Help with email" data-hide-label="Help with email" data-show-label="Help with email">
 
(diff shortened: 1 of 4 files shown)


frontstage/views/passwords/forgot_password.py
```diff
@@ -51,19 +45,13 @@ def post_forgot_password():
             if exc.status_code == 404:
                 logger.error('Requesting password change for email registered'
                              ' on OAuth2 server but not in party service')
-                template_data = {"error": {"type": {"Email address is not registered"}}}
-                return render_template('passwords/forgot-password.html', form=form, data=template_data)
+                return render_template('passwords/forgot-password.html', form=form)
```
jcox-dev: Should this also go to `passwords/forgot-password.check-email.html` as above? Seems strange to do it for one and then present no information to the user when Party check fails.

frontstage/templates/passwords/forgot-password.check-email.html
```diff
@@ -5,10 +5,9 @@
 
 {% block main %}
 
-    <h1 class="saturn">Check your email</h1>
+    <h1 class="saturn">Password reset request sent</h1>
 
-    <p>If your account is verified, you will receive an email at the registered email address.</p>
-    <p>Please follow the link in the email to reset your password.</p>
+    <p>If {{ email_address }} is registered to an account, you'll receive an email with instructions to reset your password</p>
```
shahi645: This needs to be `email` as thats what you've called it when rendering this page, or you can change the kwarg when its rendered

shahi645: I also think calling it `email_address` is better than just `email` small thing though

PR #377, opened 2018-08-06
Remove bi cases/generate todo list not using bi cases

gemmai95:
# Motivation and Context

Needed for removal of BI cases. The survey list is now created by other means that does not require BI cases.

# What has changed

Added new functionality to get survey list details from B cases instead of BI. Removed secure messaging requiring case id. 

Quite a bit of refactoring has been done around the case controller and tests. Removed unneeded code and added completed by phone to have same functionality as complete.

Tests have been written to cover all changes.

# How to test?

Tested with current docker dev that worked as expected and acceptance tests.

# Links

Trello: https://trello.com/c/PLkePSlW


frontstage/templates/surveys/surveys-todo.html
```diff
@@ -8,10 +8,10 @@
 
 {% include 'partials/tab_list.html' %}
 
-{% if just_added_case_id %}
+{% if added_survey %}
 <div class="panel panel--simple panel--success survey--success col-5@m u-mb-m ">
     <div class="panel__body">
-        <div>Your <a id="NEW_SURVEY_NOTIF" href="#ADDED_{{ just_added_case_id }}">new survey</a> has been added</div>
+        <div>Your new survey has been added</div>
```
MIslam92: Any reason we decided to remove the ability to when you add a new survey having a link on it so it jumps to where on the list that new survey was added and highlighted?

gemmai95: It was only highlighting one instance and multiple may have been added. Now it will highlight all but it can't jump to multiple places on the page. I am going to speak to Leah about this to see what she wants to do.

frontstage/common/mappers.py
```diff
@@ -0,0 +1,27 @@
...
+            "date": date_time.strftime('%d %b %Y'),
+            "month": date_time.strftime('%m'),
+            "time": date_time.strftime('%H:%M GMT'),
+            "is_in_future": date_time > iso8601.parse_date(datetime.now().isoformat())
+        }
+    return formatted_events
+
+
+def convert_date(date):
+    input_date_format = 'YYYY-MM-DDThh:mm:ss'
```
MIslam92: This `convert_date` is not covered in the tests

frontstage/controllers/case_controller.py
```diff
@@ -1,6 +1,6 @@
 import logging
+import itertools
```
andrewmil: imports should be alphabetical

frontstage/controllers/case_controller.py
```diff
@@ -64,42 +32,27 @@ def calculate_case_status(case, collection_instrument_type):
...
+                 business_party_id=business_party_id, survey_short_name=survey_short_name)
+    match = False
 
+    party = party_controller.get_respondent_party_by_id(party_id)
+    for association in party['associations']:
+        if association['partyId'] == business_party_id:
+            survey = survey_controller.get_survey_by_short_name(survey_short_name)
+            for enrolment in association['enrolments']:
+                if enrolment['surveyId'] == survey['id']:
+                    match = True
```
andrewmil: It's a bit difficult to understand what's happening here. Could this be broken down in a similar way to the survey list code with list comprehensions/generators

frontstage/controllers/case_controller.py
```diff
@@ -197,19 +145,19 @@ def get_cases_by_party_id(party_id, case_events=False):
...
 
 
-def get_eq_url(case_id, party_id):
+def get_eq_url(case_id, party_id, business_party_id, survey_short_name):
     logger.debug('Attempting to generate EQ URL', case_id=case_id, party_id=party_id)
 
     case = get_case_by_case_id(case_id)
 
-    if case['caseGroup']['caseGroupStatus'] == 'COMPLETE':
+    if case['caseGroup']['caseGroupStatus'] == 'COMPLETE' or case['caseGroup']['caseGroupStatus'] == 'COMPLETEDBYPHONE':
```
andrewmil: `if case['caseGroup']['caseGroupStatus'] in ('COMPLETE', 'COMPLETEDBYPHONE'):`
saves a bit of code

frontstage/controllers/case_controller.py
```diff
@@ -197,19 +145,19 @@ def get_cases_by_party_id(party_id, case_events=False):
...
         logger.info('The case group status is complete, opening an EQ is forbidden',
                     case_id=case_id, party_id=party_id)
         abort(403)
 
-    check_case_permissions(party_id, case['partyId'], case_id=case_id)
+    check_case_permissions(party_id, case['partyId'], business_party_id, survey_short_name)
 
-    payload = eq_payload.EqPayload().create_payload(case)
+    payload = eq_payload.EqPayload().create_payload(case, party_id, business_party_id, survey_short_name)
```
andrewmil: Should we just import the `EqPayload` class rather than the whole `eq_payload` file?

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +210,61 @@ def validate_case_category(category):
...
+def get_cases_for_list_type_by_party_id(party_id, list_type='todo'):
+    logger.debug('Get cases for party for list', party_id=party_id, list_type=list_type)
+
+    business_cases = get_cases_by_party_id(party_id)
+    cases = filter_cases_by_case_group(business_cases)
+    history_statuses = ['COMPLETE', 'COMPLETEDBYPHONE', 'NOLONGERREQUIRED']
+    if list_type == 'history':
+        filtered_cases = [business_case
+                          for business_case in cases
+                          if business_case.get('caseGroup', {}).get('caseGroupStatus') in history_statuses]
```
andrewmil: We're doing the safe `get` here. Is there a need though? I'd imagine every case coming through here will have `caseGroup` and `caseGroupStatus`. So we could just have:
`business_case['caseGroup']['caseGroupStatus']`

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +210,61 @@ def validate_case_category(category):
...
+                if enrolment['business_party']['id'] == case['partyId']
+                and enrolment['collection_exercise']['id'] == case['caseGroup']['collectionExerciseId']), None)
+    if case:
+        collection_instrument = collection_instrument_controller.get_collection_instrument(case['collectionInstrumentId'])
+        return {**enrolment,
+                "case_id": case['id'],
+                "collection_instrument": collection_instrument,
+                "status": calculate_case_status(case['caseGroup']['caseGroupStatus'], collection_instrument['type'])
+                }
+    else:
```
andrewmil: We don't need this `else: return`. The function will return None by default if it doesn't hit the `if` code

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +210,61 @@ def validate_case_category(category):
...
+        return
+
+
+def get_enrolments_with_cases(enrolments, tag):
+    business_ids = {enrolment['business_party']['id']
+                    for enrolment in enrolments}
+    cases = [get_cases_for_list_type_by_party_id(business_id, tag)
+             for business_id in business_ids]
+    flattened_case_list = [case for sublist in cases for case in sublist]
```
andrewmil: We do this flattened thing in a few places. We could make it a utility function maybe?
Something like:
```
def flatten_list(list):
    return [item for sublist in list for item in sublist]
```

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -25,12 +26,17 @@ def get_collection_exercise(collection_exercise_id):
...
     logger.debug('Successfully retrieved collection exercise', collection_exercise_id=collection_exercise_id)
-    return response.json()
+    collection_exercise = response.json()
+
+    if collection_exercise['events']:
+        collection_exercise['events'] = convert_events_to_new_format(collection_exercise['events'])
 
+    return collection_exercise
 
-def get_collection_exercise_event(collection_exercise_id, tag):
```
andrewmil: :+1: removing this saves a lot of calls

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -40,26 +46,56 @@ def get_collection_exercise_event(collection_exercise_id, tag):
...
+    collection_exercises = response.json()
+
+    for collection_exercise in collection_exercises:
+        if collection_exercise['events']:
+            collection_exercise['events'] = convert_events_to_new_format(collection_exercise['events'])
+
+    return collection_exercises
+
+
+def enrolment_with_collection_exercise(enrolment, collection_exercises):
```
andrewmil: Might want to name this function `link_enrolment_with_collection_exercise` to distinguish it better from `get_enrolments_with_collection_exercises`

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +175,25 @@ def verify_token(token):
...
+def get_enrolment_details(business_party_id, enrolments):
+    business_details = get_party_by_business_id(business_party_id)
+    enrolments_with_details = [{"business_party": business_details, "enrolment_details": enrolment}
+                               for enrolment in enrolments
+                               if enrolment['enrolmentStatus'] == 'ENABLED']
+
+    return enrolments_with_details
+
+
+def get_survey_list_details_for_party(party_id, tag):
```
andrewmil: Tiny thing might not be relevant. Since this function calls the one above it I usually would prefer to have this one on top. Same with the similar thing in collection exercise service

frontstage/controllers/survey_controller.py
```diff
@@ -30,26 +27,30 @@ def get_survey(survey_id):
...
+        raise ApiError(logger, response,
+                       log_level='warning' if response.status_code == 404 else 'exception',
+                       message='Failed to retrieve survey by its short name',
+                       survey_short_name=survey_short_name)
+
+    logger.debug('Successfully retrieved survey by its short name', survey_short_name=survey_short_name)
+    return response.json()
+
+
+def enrolment_with_survey(enrolment, surveys):
```
andrewmil: I think it would be better to call this `link_enrolment_with_survey`

frontstage/views/secure_messaging/create_message.py
```diff
@@ -16,26 +16,25 @@
...
-        sent_message = send_message(party_id, case_id, survey, ru_ref)
+        sent_message = send_message(party_id, survey, ru_ref)
         thread_url = url_for("secure_message_bp.view_conversation",
                              thread_id=sent_message['thread_id']) + "#latest-message"
         flash(Markup('Message sent. <a href={}>View Message</a>'.format(thread_url)))
         return redirect(url_for('secure_message_bp.view_conversation_list'))
 
     else:
         return render_template('secure-messages/secure-messages-view.html',
-                               ru_ref=ru_ref, survey=survey, case_id=case_id,
```
andrewmil: Do we not need this `case_id` in the secure message view anymore?

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +175,25 @@ def verify_token(token):
...
+    logger.debug("Get party enrolments", party_id=party_id)
+
+    respondent = get_respondent_party_by_id(party_id)
+    enrolments = [get_enrolment_details(association['partyId'], association['enrolments'])
+                  for association in respondent['associations']]
+    flattened_enrolments = [enrolment for sublist in enrolments for enrolment in sublist]
```
andrewmil: This might be cleaner if we have a `get_respondent_enrolments(respondent)` function that runs these two lines.

Then this function would read nicely without any list comprehensions
```
respondent = get_respondent_party_by_id(party_id)
enrolments = get_respondent_enrolments(respondent)
enrolments_with_surveys = survey_controller.get_surveys_with_enrolments(flattened_enrolments)
enrolments_with_ces = collection_exercise_controller.get_enrolments_with_collection_exercises(enrolments_with_surveys)
enrolments_with_cases = case_controller.get_enrolments_with_cases(enrolments_with_ces, tag)
```

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +175,25 @@ def verify_token(token):
...
+    enrolments_with_surveys = survey_controller.get_surveys_with_enrolments(flattened_enrolments)
+    enrolments_with_ces = collection_exercise_controller.get_enrolments_with_collection_exercises(enrolments_with_surveys)
+    enrolments_with_cases = case_controller.get_enrolments_with_cases(enrolments_with_ces, tag)
+    return enrolments_with_cases
```
andrewmil: Maybe want a debug level success log here

frontstage/views/secure_messaging/create_message.py
gemmai95: No Secure message no longer sends a case event this is all it was used for.

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +210,61 @@ def validate_case_category(category):
...
+    if list_type == 'history':
+        filtered_cases = [business_case
+                          for business_case in cases
+                          if business_case.get('caseGroup', {}).get('caseGroupStatus') in history_statuses]
+    else:
+        filtered_cases = [business_case
+                          for business_case in cases
+                          if business_case.get('caseGroup', {}).get('caseGroupStatus') not in history_statuses]
+
+    logger.debug("Sucessfully retrieved cases for party survey list", party_id=party_id, list_type=list_type)
```
MIslam92: `Successfully` 🙃

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +210,61 @@ def validate_case_category(category):
...
+    business_ids = {enrolment['business_party']['id']
+                    for enrolment in enrolments}
+    cases = [get_cases_for_list_type_by_party_id(business_id, tag)
+             for business_id in business_ids]
+    flattened_case_list = [case for sublist in cases for case in sublist]
+    enrolments_with_cases = [link_enrolment_to_cases(enrolment, flattened_case_list)
+                             for enrolment in enrolments]
+    filtered_enrolments_with_cases = [enrolment for enrolment in enrolments_with_cases
+                                      if enrolment is not None]
+    return filtered_enrolments_with_cases
```
MIslam92: Also, could we add a debug success log here to potentially?

Pipfile
```diff
@@ -23,8 +28,13 @@ arrow = "==0.12.0"
 "iso8601" = "==0.1.12"
 sdc-cryptography = "==0.2.1"
 greenlet = "==0.4.13"
+"flake8" = "*"
+responses = "*"
```
insacuri: flake8 and responses should be dev packages as they're only involved in testing

frontstage/common/mappers.py
```diff
@@ -0,0 +1,17 @@
+from datetime import datetime
+
+import iso8601
+
+
+def convert_events_to_new_format(events):
+    formatted_events = {}
+    for event in events:
+        date_time = iso8601.parse_date(event['timestamp'])
```
insacuri: Need to wrap this bit in a try/except block I think.  If the timestamp isn't correct for whatever reason, it'll throw a ParseError that needs to be handled.

Might be worth testing this with a duff timestamp and see what happens?

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +176,40 @@ def verify_token(token):
         raise ApiError(logger, response, log_level=log_level, message='Failed to verify token', token=token)
 
     logger.debug('Successfully verified token')
+
+
+def get_enrolment_details(business_party_id, enrolments):
```
jcox-dev: Doesn't appear to be any unit test coverage of these: 

- get_enrolment_details
- get_respondent_enrolments
- get_survey_list_details_for_party

frontstage/controllers/survey_controller.py
```diff
@@ -30,26 +27,35 @@ def get_survey(survey_id):
...
+                       log_level='warning' if response.status_code == 404 else 'exception',
+                       message='Failed to retrieve survey by its short name',
+                       survey_short_name=survey_short_name)
+
+    logger.debug('Successfully retrieved survey by its short name', survey_short_name=survey_short_name)
+    return response.json()
+
+
+def get_enrolment_with_survey(enrolment, surveys):
+    logger.debug("Attempting to retrieve survey for enrolment", business_party=enrolment['business_party']['id'])
```
jcox-dev: Missing coverage for these new functions:

- get_enrolment_with_survey
- get_surveys_with_enrolments

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -40,26 +46,42 @@ def get_collection_exercise_event(collection_exercise_id, tag):
...
+    for collection_exercise in collection_exercises:
+        if collection_exercise['events']:
+            collection_exercise['events'] = convert_events_to_new_format(collection_exercise['events'])
+
+    return collection_exercises
+
+
+def get_live_collection_exercises_for_survey(survey_id):
+    collection_exercises = get_collection_exercises_for_survey(survey_id)
+    live_collection_exercises = [collection_exercise
```
andrewmil: No need to assign this to a variable if we're returning it on the next line

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +175,41 @@ def verify_token(token):
...
+
+
+def get_survey_list_details_for_party(party_id, tag):
+
+    for enrolment in get_respondent_enrolments(party_id):
+        business_id = enrolment['business_id']
+        survey_id = enrolment['survey_id']
+
+        live_collection_exercises = collection_exercise_controller.get_live_collection_exercises_for_survey(survey_id)
+        collection_exercises_by_id = dict((ce['id'], ce) for ce in live_collection_exercises)
```
andrewmil: :+1: This makes later steps a lot cleaner

frontstage/controllers/survey_controller.py
```diff
@@ -1,13 +1,10 @@
 import logging
-from datetime import datetime, timezone
 
 import requests
```
andrewmil: If we're alphabetising import this should be below `flask`

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +212,35 @@ def validate_case_category(category):
...
+
+def filter_cases_by_case_group(cases):
+    logger.debug("Attempting to remove multiple cases for one case group")
+    for key, group in itertools.groupby(cases, key=lambda x: x['caseGroup']['id']):
+        group = list(group)
+        if len(group) > 1:
+            sorted_group = sorted(group, key=lambda k: k['createdDateTime'], reverse=True)
+            for i in sorted_group:
+                if sorted_group.index(i) != 0:
+                    cases.remove(i)
```
andrewmil: This can be simplified using generators I think
```
grouped_cases = (list(group)
                 for _, group in itertools.groupby(cases, key=lambda x: x['caseGroup']['id']))
return (sorted(group, key=lambda k: k['createdDateTime'], reverse=True)[0]
        for group in grouped_cases)
```

frontstage/controllers/survey_controller.py
gemmai95: Why would this be below flask?

jcox-dev: I think we should probably aim for:

```python
import standard_package  # stdlib level

import a_package  # third party level
import package
from another_package import module
from bigger_package.library import tool

from application import controller  # project level
```

frontstage/controllers/party_controller.py
```diff
@@ -154,3 +175,53 @@ def verify_token(token):
...
+                'collection_exercise': collection_exercises_by_id[case['caseGroup']['collectionExerciseId']]
+            }
+
+
+def is_respondent_enrolled(party_id, business_party_id, survey_short_name):
+    for enrolment in get_respondent_enrolments(party_id):
+        survey = survey_controller.get_survey_by_short_name(survey_short_name)
+        if enrolment['business_id'] == business_party_id and enrolment['survey_id'] == survey['id']:
+            return True
+    return
```
andrewmil: Don't need this `return` here as function will automatically `return None` if it reaches the end of the function

frontstage/controllers/case_controller.py
```diff
@@ -154,27 +89,23 @@ def get_case_categories():
...
+def get_case_data(case_id, party_id, business_party_id, survey_short_name):
     logger.debug('Attempting to retrieve detailed case data', case_id=case_id, party_id=party_id)
 
     # Check if respondent has permission to see case data
     case = get_case_by_case_id(case_id)
-    check_case_permissions(party_id, case['partyId'], case_id=case_id)
-
-    full_case_data = build_full_case_data(case)
+    if not party_controller.is_respondent_enrolled(party_id, business_party_id, survey_short_name):
+        raise NoSurveyPermission(party_id, case_id)
```
andrewmil: Untested Exception

frontstage/controllers/case_controller.py
```diff
@@ -197,19 +128,20 @@ def get_cases_by_party_id(party_id, case_events=False):
...
 
-def get_eq_url(case_id, party_id):
+def get_eq_url(case_id, party_id, business_party_id, survey_short_name):
     logger.debug('Attempting to generate EQ URL', case_id=case_id, party_id=party_id)
 
     case = get_case_by_case_id(case_id)
 
-    if case['caseGroup']['caseGroupStatus'] == 'COMPLETE':
+    if not party_controller.is_respondent_enrolled(party_id, business_party_id, survey_short_name):
+        raise NoSurveyPermission(party_id, case_id)
```
andrewmil: Untested Exception

frontstage/controllers/party_controller.py
```diff
@@ -204,12 +204,24 @@ def get_survey_list_details_for_party(party_id, tag):
...
+                'survey_short_name': survey['shortName'],
+                'survey_ref': survey['surveyRef'],
                 'business_party': get_party_by_business_id(business_id),
                 'collection_exercise': collection_exercises_by_id[case['caseGroup']['collectionExerciseId']]
             }
+
+
+def is_respondent_enrolled(party_id, business_party_id, survey_short_name):
+    for enrolment in get_respondent_enrolments(party_id):
+        survey = survey_controller.get_survey_by_short_name(survey_short_name)
```
jcox-dev: Can this be moved outside the `for` loop if we're only ever going to get the same survey each iteration?

frontstage/controllers/case_controller.py
```diff
@@ -261,3 +194,30 @@ def validate_case_category(category):
...
+        filtered_cases = [business_case
+                          for business_case in cases
+                          if business_case['caseGroup']['caseGroupStatus'] not in history_statuses]
+
+    logger.debug("Successfully retrieved cases for party survey list", party_id=party_id, list_type=list_type)
+    return filtered_cases
+
+
+def filter_cases_by_case_group(cases):
+    logger.debug("Attempting to remove multiple cases for one case group")
```
jcox-dev: There's a bit of magic going on here that would be good to document in a docstring/comment as it isn't straight forward to scan.

frontstage/common/mappers.py
```diff
@@ -0,0 +1,21 @@
...
+from iso8601 import parse_date, ParseError
+
+
+def convert_events_to_new_format(events):
+    formatted_events = {}
+    for event in events:
+        try:
+            date_time = parse_date(event['timestamp'])
+        except ParseError:
+            raise ParseError
```
shahi645: Do we want a message to go with this?

frontstage/common/mappers.py
```diff
@@ -0,0 +1,21 @@
...
+
+        formatted_events[event['tag']] = {
+            "day": date_time.strftime('%A'),
+            "date": date_time.strftime('%d %b %Y'),
+            "month": date_time.strftime('%m'),
+            "time": date_time.strftime('%H:%M GMT'),
```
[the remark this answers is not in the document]
gemmai95: Don't think so. I will have a look at removing these.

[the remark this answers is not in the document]
tej99: kind find a usage for month but day and time both get used

PR #380, opened 2018-08-13
Forgotten password redirect url

gavinleeedwards:
# Motivation and Context
There are 2 issues that have been fixed:

- Issue with the redirect URL -  if user is registered they are redirected to `passwords/forgot-password/check-email` else to `passwords/forgot-password`.  This is still an issue as someone could potentially figure out if that email is registered or not

- If a registered email was entered, the email address wasn't being passed through via the redirect so wasn't being played back to the user on the `passwords/forgot-password/check-email` page

# What has changed
- Regardless fo whether the email address is registered or not, the user will always be redirected to `passwords/forgot-password/check-email`
- The email address is passed via the URL in an encoded string using URLSafeSerializer (https://pythonhosted.org/itsdangerous/).  This will prevent users manipulating the URL to inject anything dodgy onto the page and will disguise the email in the browser history.

# How to test?
- All tests should pass
- Attempt a password reset using a registered email and a non-registered email.  You should be directed to the same page for both and the email address should be played back for both i.e. `If <email> is registered to an account, you'll receive an email....`

# Links

https://trello.com/c/sWmRfnoD/292-forgotten-password-redirect-url-s
https://pythonhosted.org/itsdangerous/


frontstage/views/passwords/forgot_password.py
```diff
@@ -14,6 +16,8 @@
 
 BAD_AUTH_ERROR = 'Unauthorized user credentials'
 
+s = URLSafeSerializer(app.config['JWT_SECRET'])
```
ajmaddaford: This should use `SECRET_KEY` rather than `JWT_SECRET`

frontstage/views/passwords/forgot_password.py
```diff
@@ -45,15 +51,22 @@ def post_forgot_password():
...
 
 @passwords_bp.route('/forgot-password/check-email', methods=['GET'])
 def forgot_password_check_email():
-    return render_template('passwords/forgot-password.check-email.html')
+    encoded_email = request.args.get('email', None)
+
+    try:
+        email = s.loads(encoded_email)
+    except BadSignature:
+        raise InvalidURLSignature('Unable to decode email from URL')
```
ajmaddaford: I don't think we need to raise a custom exception, we could just deal with the error here.

frontstage/views/passwords/forgot_password.py
```diff
@@ -45,15 +51,22 @@ def post_forgot_password():
             if exc.status_code == 404:
                 logger.error('Requesting password change for email registered'
```
shahi645: @gavinleeedwards worth checking, but is this an actual error, maybe be like a warning log?

shahi645: also, if it is an `error` would we want to redirect?

gavinleeedwards: @shahi645 Good shout.  At this point it would be an actual error because the user should be in party.  This exception is also hit from a normal APIError - if party was down - so the log message wasn't covering everything.  As it's an error on our side, the user shouldn't be told that the email has been sent (they may have entered their email fine) so it will display the 500 page instead.

frontstage/templates/passwords/forgot-password.check-email.html
```diff
@@ -20,7 +20,7 @@ <h1 class="saturn">Password reset request sent</h1>
               <li>Check your spam folder</li>
               <li>Check that you've entered your email address correctly </li>
               <li>Ensure that you've registered an account using this email address </li>
-              <br>If you're still having problems, call us on 0300 1234 931 or send an <br/>email to <a href="mailto:surveys.ons@gov.uk">surveys.ons@gov.uk</a>.</p>
+              <br>If you're still having problems, call us on 0300 1234 931 or send an <br/>email to <a href="mailto:surveys@.ons.gov.uk">surveys@ons.gov.uk</a>.</p>
```
shahi645: you've got a `.` after the `@` you can remove

gavinleeedwards: All done :+1:

frontstage/views/passwords/forgot_password.py
```diff
@@ -43,17 +49,24 @@ def post_forgot_password():
...
 @passwords_bp.route('/forgot-password/check-email', methods=['GET'])
 def forgot_password_check_email():
-    return render_template('passwords/forgot-password.check-email.html')
+    encoded_email = request.args.get('email', None)
+
+    try:
+        email = s.loads(encoded_email)
+    except BadSignature:
+        logger.info('Unable to decode email from URL')
+        return render_template('errors/404-error.html'), 400
```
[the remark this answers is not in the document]
ajmaddaford: We should also log this as `error` rather than `info`

frontstage/views/passwords/forgot_password.py
```diff
@@ -43,17 +49,28 @@ def post_forgot_password():
...
 
 
 @passwords_bp.route('/forgot-password/check-email', methods=['GET'])
 def forgot_password_check_email():
-    return render_template('passwords/forgot-password.check-email.html')
+    encoded_email = request.args.get('email', None)
+
+    if encoded_email is None:
+        logger.error('No email parameter supplied')
+        return redirect('passwords/forgot-password')
```
ajmaddaford: Use `url_for` and the blueprint here to avoid duplicating the url path

frontstage/views/passwords/forgot_password.py
```diff
@@ -43,17 +49,28 @@ def post_forgot_password():
...
+
+    try:
+        email = url_safe_serializer.loads(encoded_email)
+    except BadSignature:
+        logger.error('Unable to decode email from URL', encoded_email=encoded_email)
+        return render_template('errors/404-error.html'), 404
```
ajmaddaford: Use `abort(404)` here to avoid duplicating the reference to the 404-error template

frontstage/views/passwords/forgot_password.py
```diff
@@ -43,17 +49,28 @@ def post_forgot_password():
             party_controller.reset_password_request(email)
         except ApiError as exc:
             if exc.status_code == 404:
-                logger.error('Requesting password change for email registered'
-                             ' on OAuth2 server but not in party service')
-                return render_template('passwords/forgot-password.check-email.html', email=email)
+                logger.error('Failed to retrieve details from party service')
+                return render_template('errors/404-error.html')
```
ajmaddaford: I think this should be a 500 error, as if this happens it will be because something has gone wrong on our side

shahi645: i agree

gavinleeedwards: This exception handling has been removed altogether now as it's already being dealt with by party_controller.

ajmaddaford:
We should have a test for when the user exists in the oauth server but not party. https://github.com/ONSdigital/ras-frontstage/blob/70a560eda2a3b9503e569ae1777c83764f81a58f/tests/app/test_passwords.py#L69 looks like it should test this, but it expects a 200 rather than a 500.

PR #390, opened 2018-08-29
Notify user to verify their email

gavinleeedwards:
# Motivation and Context
Currently, if an external user attempts to log in before they've verified their account, they are taken to a page and told that they must verify before continuing.  If they cannot find their original verification email from when they signed up then they have to contact us to request another verification email, as new verification emails can only be sent from response-operations.
This PR allows external users to request another verification email if they sign in with an unverified account.

# What has changed
- Amended existing template to allow external users to request another verification email
- Added a new template to inform users that the verification email has been sent
- Added functionality to send another verification email via the party service endpoint

# How to test?
- Create a new account but do not verify it
- Log in with the new account - you should be told that the account is locked and have the option to resend the verification email
- Click the link to resend the verification email.  A new mail should be sent (check the party service logs for the new verification URL)

# Links
Trello card --> https://trello.com/c/5LKDQyBI/282-us205-ext-notifying-a-user-to-verify-their-email-address-when-they-sign-in-to-an-unverified-account-m


frontstage/templates/sign-in/sign-in.account-not-verified.html
```diff
@@ -1,12 +1,14 @@
...
-{% block page_title %}Trouble signing in? - ONS Business Surveys{% endblock %}
+{% block page_title %}Verify your email - ONS Business Surveys{% endblock %}
 
 {% block main %}
-    <h1 class="saturn">Trouble signing in?</h1>
+    <h1 class="saturn">Verify your email</h1>
 
-    <p>Please follow the link in the verification email we sent you.</p>
-    <p>If you haven't received an email please call us on 0300 1234 931 or email <a href="mailto:surveys@ons.gov.uk">surveys@ons.gov.uk</a></p>
+    <p>When you created your account we sent you an email to verify your email address.</p>
```
shahi645: This matches content, but i think it doesn't make sense. Shouldn't it be:

```
When you created your account we sent you an email to verify your account.
```

frontstage/views/sign_in/sign_in.py
```diff
@@ -78,3 +79,10 @@ def login():
         'account_activated': account_activated
     }
     return render_template('sign-in/sign-in.html', form=form, data=template_data)
+
+
+@sign_in_bp.route('/resend_verification/<party_id>', methods=['GET'])
+def resend_verification(party_id):
+    party_controller.resend_verification_email(party_id)
+    logger.info("Re-sent verification email.", party_id=party_id)
```
shahi645: single quotes please 😄

frontstage/controllers/party_controller.py
```diff
@@ -109,6 +109,20 @@ def get_respondent_by_email(email):
     return response.json()
 
 
+def resend_verification_email(party_id):
+    logger.debug('Re-sending verification email', party_id=party_id)
```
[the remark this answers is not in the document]
NickGrantONS: IMHO 'info' means "useful information when the service is operating normally" and 'debug' means "extra information which might help with debugging, but is otherwise not useful when service is robust and trusted". I'd say that any happy-path logging is debug, not info. Just my two cents tho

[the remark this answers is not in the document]
shahi645: That sounds like a compelling argument, I'm happy to go with this. I always get confused 😕

frontstage/templates/sign-in/sign-in.account-not-verified.html
jcox-dev: Either makes sense in my mind, @shahi645. And both forms can been found elsewhere 👍

shahi645: Yeah i noticed that, as long as we are consistent i suppose

PR #392, opened 2018-09-03
Fix missing jwcrypto hashes

tej99:
# Motivation and Context

Pipenv lock did not add jwcrypto hashes into the Piplock.file.
Re doing it after updating pipenv and clearing cache.

# What has changed

Explicitly specify jwcrpto package in Pipfile

# How to test?

Tests should pass

# Links

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,12 +1,17 @@
 [[source]]
+
 name = "pypi"
 verify_ssl = true
 url = "https://pypi.python.org/simple"
 
+
 [requires]
+
 python_version = "3.6"
 
+
 [packages]
+
 cfenv = "*"
 flask = "*"
 flask-wtf = "*"
@@ -26,7 +31,9 @@ dateutils = "*"
 requestsdefaulter = "*"
 flask-zipkin = "*"
 
+
 [dev-packages]
+
 codecov = "*"
 coverage = "*"
 "flake8" = "*"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -25,8 +30,11 @@ greenlet = "*"
 dateutils = "*"
 requestsdefaulter = "*"
 flask-zipkin = "*"
+jwcrypto = "*"
```
AdamHawtin: You shouldn't need to add jwcrypto as a top level dependency if it wasn't before

tej99: Yes I agree but... running `pipenv lock` without it results in 
``` 
"jwcrypto": {
    "hashes": [],
    "version": "==0.5"
```
And I want to avoid pinning the top level dependency to a version that produces the correct hashes.

tej99: Found the problem: https://github.com/ONSdigital/sdc-cryptography/blob/450f022983ae4912d002964ee2abb41b81fa8f2e/requirements.txt#L2

Luckily it's a repo we manage...

Pipfile.lock
```diff
@@ -266,7 +262,10 @@
             "version": "==2.10"
         },
         "jwcrypto": {
-            "hashes": [],
+            "hashes": [
+                "sha256:3158ea0ccb70407ec441e76d67aeb827ff85adb142291c1d184f8b8bd4dbdb93",
```
AdamHawtin: :thumbsup:

PR #396, opened 2018-09-11
269 ext user account unlock

shahi645:
# Motivation and Context
We want to display a template to the respondent when the account is locked and to call party to notify the respondent via gov notify
# What has changed
- Added a party controller to call endpoint in party-service which calls notify gateway
- Added a new template

# How to test?
Run with [Django](https://github.com/ONSdigital/django-oauth2-test/pull/46) and [party](https://github.com/ONSdigital/ras-party/tree/269-ext-user-account-unlock) once account is locked you should see a new template. Then check Party logs to get the url to reset your password and unlock your account.

# Links
[Trello](https://trello.com/c/6IEaOnA9/269-us204-ext-user-account-locked-l)


frontstage/templates/sign-in/sign-in.account-locked.html
```diff
@@ -0,0 +1,37 @@
+{% import 'partials/section.html' as section %}
+{% extends "layouts/_twocol.html" %}
+
+{% block page_title %}Trouble signing in? - ONS Business Surveys{% endblock %}
+
+{% block main %}
+    <h1 class="saturn">Check your email</h1>
+
+    <p>You've tried to sign in few times with the wrong details.</p>
+    <p>If your email is registered to an account, we will send an email explaining how to sign in.</p>
```
MIslam92: This based of the content should be passing in your actual email address the user tried registering with

shahi645: no we dont want to pass email about, plus this is an anonymous page, we dont want to let the people know which email it is

shahi645: ill double check with someone

frontstage/controllers/party_controller.py
```diff
@@ -267,3 +267,18 @@ def change_respondent_status(respondent_id, status):
         raise ApiError(logger, response)
 
     logger.info('Successfully changed account status', respondent_id=respondent_id, status=status)
+
+
+def notify_account_locked(username):
+    logger.debug('Sending notification email for password reset')
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/respondents/notify-respondent'
```
tej99: This pr probably shouldn't be merged in until the party service changes have been made and merged.

frontstage/views/sign_in/sign_in.py
```diff
@@ -47,12 +47,13 @@ def login():
...
         except OAuth2Error as exc:
             error_message = exc.oauth2_error
-            if BAD_AUTH_ERROR in error_message:
-                return render_template('sign-in/sign-in.html', form=form, data={"error": {"type": "failed"}})
-            elif USER_ACCOUNT_LOCKED in error_message:
+            if USER_ACCOUNT_LOCKED in error_message:
                 logger.info('User account is locked on the OAuth2 server')
                 if party_json['status'] == 'ACTIVE' or party_json['status'] == 'CREATED':
                     change_respondent_status(respondent_id=party_id, status='SUSPENDED')
+                notify_account_locked(username=username)
```
tej99: Do you want to send an email every time a user tries to log in with a locked account? Probably should have this within the `if party_json['status'] == 'ACTIVE'` block so it only done when the account is set to suspended.

Possibly even have it as logic in party (when setting status to suspended send out email to user) so you don't have to call party twice and have it as part of a single transaction.

tests/app/controllers/test_party_controller.py
```diff
@@ -6,13 +6,13 @@
 from config import TestingConfig
 from frontstage import app
 from frontstage.controllers import party_controller
-from frontstage.controllers.party_controller import change_respondent_status
```
gemmai95: Do we need this import if we are importing party_controller above?

frontstage/controllers/party_controller.py
```diff
@@ -251,19 +251,22 @@ def is_respondent_enrolled(party_id, business_party_id, survey_short_name, retur
             return True
 
 
-def change_respondent_status(respondent_id, status):
-    logger.debug('Changing account status', respondent_id=respondent_id, status=status)
+def notify_party_and_respondent_account_locked(respondent_id, email_address, status=None):
```
tej99: The name before fits the function better, status is passed in so respondent is not always going to be set to 'account_locked' also depending on what status is passed through notify will/will not be called.

shahi645:
Yep @gavinleeedwards

PR #400, opened 2018-09-14
Allow a verification link to be resent if using an expired one

gavinleeedwards:
# Motivation and Context
If a respondent attempts to use an expired verification token (older than 80 hours) then they have to phone us and request to be sent another link.  This change allows respondents to resend the verification link to their email.

# What has changed
- Amended the page that is displayed when users attempt to us an expired link, giving them the option to send another.
- Call to a new endpoint in party to use the expired token to send another mail.
- Added tests.

# How to test?
- Run against branch`206-resend-verification-if-using-expired-link` in ras-party.  Ensure that `EMAIL_TOKEN_EXPIRY` is set to 0 in party so that the links immediately expire when manually testing.
- Using the new link to resend the verification email should generate a new verification link - check in party logs.
- All tests should pass.

# Links
- https://trello.com/c/fqc35BzO

### DIFF
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -145,6 +145,18 @@ def resend_verification_email(party_id):
     logger.debug('Successfully re-sent verification email', party_id=party_id)
 
 
+def resend_verification_email_expired_token(token):
+    logger.debug('Re-sending verification email', token=token)
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/resend-verification-email-expired-token/{token}'
+    response = requests.get(url, auth=app.config['PARTY_AUTH'])
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception('Re-sending of verification email for expired token failed', token=token)
+        raise ApiError(response)
+
+
 def reset_password_request(username):
     logger.debug('Attempting to send reset password request to party service')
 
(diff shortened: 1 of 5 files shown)


frontstage/controllers/party_controller.py
```diff
@@ -145,6 +145,18 @@ def resend_verification_email(party_id):
     logger.debug('Successfully re-sent verification email', party_id=party_id)
 
 
+def resend_verification_email_expired_token(token):
+    logger.debug('Re-sending verification email', token=token)
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/resend-verification-email-expired-token/{token}'
+    response = requests.get(url, auth=app.config['PARTY_AUTH'])
+
```
AndrewTorrance: This should be a Post since its not idempotent.

frontstage/controllers/party_controller.py
```diff
@@ -145,6 +145,18 @@ def resend_verification_email(party_id):
     logger.debug('Successfully re-sent verification email', party_id=party_id)
 
 
+def resend_verification_email_expired_token(token):
+    logger.debug('Re-sending verification email', token=token)
```
shahi645: @gavinleeedwards Just a note on this and a lot of the controllers in this file, they are all quite similar, might be worth adding a tech debt ticket to look to refactor and move common bits to private functions. Just a thought

andrewmil: Could be a good idea @shahi645 but worth noting that we used to have a similar requests wrapper which did that. 
It had the exception handling/basic auth stuff handled inside it. In general though we found that it was just an extra thing to worry about as we had to make modifications to it every time a new scenario came up for the request (new ways of handling errors/different types of request/sending different types of data)

PR #401, opened 2018-09-17
Get only the live collection excercises for survey

NickGrantONS:
# Motivation and Context
The Collection Exercise service has been enhanced to be able to return only collexes which are in LIVE state, so that they don't have to be received and filtered by Response-Ops and Frontstage.

# What has changed
Updated method `get_collection_exercises_for_survey` so that it has an optional parameter to only return the live collexes and updated `get_live_collection_exercises_for_survey` to use that method instead of doing the filtering.

# How to test?
Load frontstage and check that only live collection exercises can be viewed.

# Links
Trello: https://trello.com/c/NFKGLLkV/370-logic-to-be-removed-in-affected-services-ie-frontstage-rops-m

### DIFF
--- a/frontstage/controllers/collection_exercise_controller.py
+++ b/frontstage/controllers/collection_exercise_controller.py
@@ -53,10 +53,13 @@ def get_collection_exercise_events(collection_exercise_id):
     return response.json()
 
 
-def get_collection_exercises_for_survey(survey_id):
+def get_collection_exercises_for_survey(survey_id, live_only=None):
     logger.debug('Retrieving collection exercises for survey', survey_id=survey_id)
 
-    url = f"{app.config['COLLECTION_EXERCISE_URL']}/collectionexercises/survey/{survey_id}"
+    if live_only is True:
+        url = f"{app.config['COLLECTION_EXERCISE_URL']}/collectionexercises/survey/{survey_id}?liveOnly=true"
+    else:
+        url = f"{app.config['COLLECTION_EXERCISE_URL']}/collectionexercises/survey/{survey_id}"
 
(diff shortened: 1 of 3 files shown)


frontstage/controllers/collection_exercise_controller.py
```diff
@@ -53,10 +53,13 @@ def get_collection_exercise_events(collection_exercise_id):
     return response.json()
 
 
-def get_collection_exercises_for_survey(survey_id):
+def get_collection_exercises_for_survey(survey_id, live_only=None):
     logger.debug('Retrieving collection exercises for survey', survey_id=survey_id)
 
-    url = f"{app.config['COLLECTION_EXERCISE_URL']}/collectionexercises/survey/{survey_id}"
+    if live_only is True:
```
shahi645: can just say `if live_only:`

frontstage/controllers/collection_exercise_controller.py
```diff
@@ -79,11 +82,7 @@ def get_collection_exercises_for_survey(survey_id):
 
 
 def get_live_collection_exercises_for_survey(survey_id):
-    collection_exercises = get_collection_exercises_for_survey(survey_id)
-    return [collection_exercise
-            for collection_exercise in collection_exercises
-            if collection_exercise['state'] == 'LIVE'
-            and not collection_exercise['events']['go_live']['is_in_future']]
+    return get_collection_exercises_for_survey(survey_id, True)
```
shahi645: 👍

frontstage/controllers/collection_exercise_controller.py
gemmai95: He wants to make sure it only goes to that one if the parameter passed is True as if you did that it would use it if anything was passed.

PR #404, opened 2018-09-19
Logout RAS RM from EQ

MIslam92:
# Motivation and Context
EQ have implemented the logout facility so redirect to our logout page, but they need us to provide the url to redirect to. Although we already pass the location of the todo list we also need to do this as well with another variable we can pass to logout to the relevant page.

# What has changed
Adding in a new variable to pass in the payload to EQ to logout from EQ & RAS to the signed out page.
A new env has also been introduced to Jenkins for this as well.

# How to test?
Ensure AT pass etc.

# Links
Trello Card: https://trello.com/c/6Bfd2jlB/382-logout-ras-rm-from-eq-m

### DIFF
--- a/config.py
+++ b/config.py
@@ -31,6 +31,7 @@ class Config(object):
     ZIPKIN_SAMPLE_RATE = int(os.getenv("ZIPKIN_SAMPLE_RATE", 0))
 
     ACCOUNT_SERVICE_URL = os.getenv('ACCOUNT_SERVICE_URL')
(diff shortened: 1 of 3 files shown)


dcdarrell9:
🎨 Just a comment, couldn't we have a logout_bp where the logout functionality would be instead of adding it onto the end of of sign-in? `sign-in/logout` just seems a bit weird. Feel free to ignore :man_shrugging:

PR #406, opened 2018-09-24
207 resend password reset link if expired

gavinleeedwards:
# Motivation and Context
When a respondent requests a change of password, they are emailed a link that expires after 80 hours.  If they try to use the link after the 80 hours then they are redirected to a page that tells them to phone us for another link.  This PR allows the respondent to generate a new link which is then sent to the email registered to their account.

# What has changed

New and amended pages to allow a user to resend a password reset link if the one they attempt to use has expired. 

# How to test?
- All unit tests should pass
- Set the EMAIL_TOKEN_EXPIRY in party config to -1 so that tokens expire as soon as they're created (otherwise you'll be waiting 80 hours for one to expire)
- Run this branch alongside `207-resend-password-reset-link-if-expired` in ras-party
- In frontstage, request a password reset for an active account - check the party logs to find it
- Follow the link to reset the password.  The link should be expired and you should be redirected to a page and given the option to resend a password reset link
- Click the link to send another and check in the party logs that another link has been generated
- You should finally be redirected to the 'Check your email page'
- The new link that's been generated in party should allow you to reset your password as normal (if you change the EMAIL_TOKEN_EXPIRY back)

# Links
- https://trello.com/c/1FXbQiA2

### DIFF
--- a/frontstage/templates/passwords/password-expired.html
+++ b/frontstage/templates/passwords/password-expired.html
@@ -9,10 +9,7 @@
 
 <div>
     <p>
-        You can either <a href="/passwords/forgot-password">request another password reset</a>.
-    </p>
-    <p>
-        or <a href="/sign-in">sign in</a> with your existing password.
+        You can <a href="{{ url_for('passwords_bp.resend_password_email_expired_token', token=token) }}">request another password reset email</a> or call us on <a href="tel:03001234931">0300 1234 931</a>.
     </p>
 </div>
 
--- a/frontstage/templates/passwords/password-expired.html
+++ b/frontstage/templates/passwords/password-expired.html
@@ -9,7 +9,8 @@
 
 <div>
     <p>
-        You can <a href="{{ url_for('passwords_bp.resend_password_email_expired_token', token=token) }}">request another password reset email</a> or call us on <a href="tel:03001234931">0300 1234 931</a>.
+        You can <a href="{{ url_for('passwords_bp.resend_password_email_expired_token', token=token) }}">request another password reset email</a>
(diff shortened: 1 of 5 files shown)


frontstage/templates/passwords/password-expired.html
```diff
@@ -9,10 +9,8 @@ <h1 class="saturn">Your link has expired</h1>
 
 <div>
     <p>
-        You can either <a href="/passwords/forgot-password">request another password reset</a>.
-    </p>
-    <p>
-        or <a href="/sign-in">sign in</a> with your existing password.
+        You can <a href="{{ url_for('passwords_bp.resend_password_email_expired_token', token=token) }}">request another password reset email</a><br/>
```
shahi645: Remove line break, its not in the design

gavinleeedwards: You're right, it isn't but if the break isn't there then half of the telephone number wraps around to the next line which doesn't look right.  Agreed to move the break to move just the telephone number to the next line.

frontstage/views/passwords/reset_password.py
```diff
@@ -67,3 +67,10 @@ def post_reset_password(token):
 @passwords_bp.route('/reset-password/confirmation', methods=['GET'])
 def reset_password_confirmation():
     return render_template('passwords/reset-password.confirmation.html')
+
+
+@passwords_bp.route('/resend-password-email-expired-token/<token>', methods=['GET'])
```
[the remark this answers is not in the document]
gavinleeedwards: Changed to /reset-password/check-email in line with similar existing endpoints.

PR #407, opened 2018-09-26
Remove unnecessary auth call

tej99:
# Motivation and Context

A call to authentication service was introduced as a way to stop locked users to reset their password in [PR230](https://github.com/ONSdigital/ras-frontstage/pull/230).

The requirements around this has changed so locked users can now reset their password and unlock
their account (changed logic in django-oauth-test [PR46](https://github.com/ONSdigital/django-oauth2-test/pull/46) ). Which now makes the call to the django-oauth-test service redundant.

# What has changed

Removed the call to django-oauth-test for send reset password email

# How to test?

Ran AT and check you can still reset your password for both locked and unlocked account.

# Links

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ install:
   - pipenv install --dev --deploy
 
 script:
+  - make lint
   - APP_SETTINGS=TestingConfig pipenv run pytest --cov=frontstage --cov-report xml --ignore=node_modules
   - pipenv run coverage report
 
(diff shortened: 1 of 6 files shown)


shahi645:
Good to approve this, just waiting on @tej99 to fix some linting errors and remove dead code

.travis.yml
```diff
@@ -16,6 +16,7 @@ install:
   - pipenv install --dev --deploy
 
 script:
+  - make lint
```
shahi645: 🌮

PR #412, opened 2018-10-10
Only display access survey button for eQ when:

shahi645:
# Motivation and Context
Only display access survey button for eQ when: 
 - notstarted, 
 - inprogress,
 - reopened

# What has changed
Above

# How to test?
Change casegroup status and see if button disappears

### DIFF
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -8,6 +8,8 @@ from frontstage.controllers import case_controller, collection_exercise_controll
     survey_controller
 from frontstage.exceptions.exceptions import ApiError, UserDoesNotExist
 
+CLOSED_STATE = ['COMPLETE', 'COMPLETEDBYPHONE', 'NOLONGERREQUIRED']
+
 logger = wrap_logger(logging.getLogger(__name__))
 
 
@@ -245,6 +247,9 @@ def get_survey_list_details_for_party(party_id, tag, business_party_id, survey_i
             )
             collection_exercise = collection_exercises_by_id[case['caseGroup']['collectionExerciseId']]
             added_survey = True if business_party_id == business_party['id'] and survey_id == survey['id'] else None
+            display_access_button = should_display_access_button(case['caseGroup']['caseGroupStatus'],
+                                                                 collection_instrument['type'])
+
             yield {
 
                 'case_id': case['id'],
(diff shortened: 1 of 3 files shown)


frontstage/controllers/party_controller.py
```diff
@@ -262,10 +267,17 @@ def get_survey_list_details_for_party(party_id, tag, business_party_id, survey_i
                 'period': collection_exercise['userDescription'],
                 'submit_by': collection_exercise['events']['return_by']['date'],
                 'collection_exercise_ref': collection_exercise['exerciseRef'],
-                'added_survey': added_survey
+                'added_survey': added_survey,
+                'display_access_button': display_access_button
             }
 
 
+def should_display_access_button(status, ci_type):
```
jcox-dev: Might make more sense to change to:

```python
def is_closed(status, ci_type):
    return ci_type == 'EQ' or status in CLOSED_STATE
```

frontstage/templates/surveys/surveys.html
```diff
@@ -49,7 +49,7 @@
             <span id="status-{{ loop.index }}" name="status">{{ survey.status }}</span>
         </div>
         <div class="grid__col col-2@m">
-            {% if not survey.status == 'Complete' or not survey.collection_instrument_type == 'EQ' %}
+            {% if not survey.display_access_button %}
```
jcox-dev: This is currently only showing the access link if it *shouldn't* be displayed.

Probably worth changing to:
```python
{% if not survey.is_closed %}
...
```
to avoid confusion.

tests/app/controllers/test_party_controller.py
```diff
@@ -208,3 +210,12 @@ def test_get_survey_list_details_for_party(self, get_respondent_enrolments, get_
             self.assertTrue(survey_details['survey_short_name'] is not None)
             self.assertTrue(survey_details['business_party_id'] is not None)
             self.assertTrue(survey_details['collection_exercise_ref'] is not None)
+
+    def test_should_not_display_access_button_for_eQ(self):
+        self.assertFalse(should_display_access_button(status='COMPLETE', ci_type='EQ'))
```
jcox-dev: We could have test fixture-like lists of tuple combinations here instead of three separate tests:

```python
...
    Combination = namedtuple("Combination", ["status", "ci_type", "expected"])
    combinations = [
        Combination('NOTSTARTED', 'SEFT', True),
        Combination('COMPLETE', 'SEFT', False), 
        ...  # etc.
    ]
    for c in combinations:
        self.assertEqual(should_display_access_button(c.status, c.ci_type), c.expected)
    ...
```

shahi645: This is a lot better actually 🌮

frontstage/controllers/party_controller.py
tej99: is_closed is probably not the best name for this function because a seft survey "is closed" but it shouldn't return True because Seft surveys can still be accessed when the survey has been completed. Also that function would return True for all EQ surveys, whether they are closed or not.

frontstage/templates/surveys/surveys.html
jcox-dev: Spoke to @shahi645 and makes sense to keep `.display_button` but should be changed to `{% if survey.display... %}`

PR #413, opened 2018-10-18
Adjust log level for iac not found

gemmai95:
# Motivation and Context

Stop spamming splunk logs and causing alerts.

# What has changed

Change iac code not active or found is handled.

# How to test?

# Links

Trello: https://trello.com/c/DylkAFhC

### DIFF
--- a/frontstage/controllers/iac_controller.py
+++ b/frontstage/controllers/iac_controller.py
@@ -11,7 +11,7 @@ from frontstage.exceptions.exceptions import ApiError
 logger = wrap_logger(logging.getLogger(__name__))
 
 
-def get_iac_from_enrolment(enrolment_code, validate=False):
+def get_iac_from_enrolment(enrolment_code):
     logger.debug('Attempting to retrieve IAC')
     url = f"{app.config['IAC_URL']}/iacs/{enrolment_code}"
     response = requests.get(url, auth=app.config['IAC_AUTH'])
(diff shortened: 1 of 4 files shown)


frontstage/controllers/iac_controller.py
```diff
@@ -21,18 +21,16 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
     except requests.exceptions.HTTPError:
         if response.status_code == 404:
             logger.info('IAC not found')
-            if validate:
-                raise ApiError(logger, response, message='Invalid enrolment code used')
             return
```
shahi645: Theres no `return` here, you dont need this, or you need a `pass`/`continue`?

frontstage/controllers/iac_controller.py
```diff
@@ -21,18 +21,16 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
     except requests.exceptions.HTTPError:
         if response.status_code == 404:
             logger.info('IAC not found')
```
shahi645: probably wan't to log status code from API?

frontstage/controllers/iac_controller.py
```diff
@@ -21,18 +21,16 @@ def get_iac_from_enrolment(enrolment_code, validate=False):
...
-                raise ApiError(logger, response, message='Invalid enrolment code used')
             return
         # 401s may include error context in the JSON response
         elif response.status_code != 401:
             raise ApiError(logger, response, message='Failed to retrieve IAC')
 
-    if validate and not response.json().get('active', False):
-        raise ApiError(logger, response, message='Invalid enrolment code used')
+    if response.json().get('active', False):
+        logger.info("Invalid IAC used")
```
shahi645: status code?

jcox-dev: Does this need to return False/None or something other than the JSON response below to show it's inactive?

frontstage/controllers/iac_controller.py
gemmai95: I don't really get what you mean there is a return right there.

frontstage/controllers/iac_controller.py
gemmai95: The status code won't be useful here as it will be 200

shahi645: not from the Api?

frontstage/controllers/iac_controller.py
shahi645: i mean you're better off using a `pass`

jcox-dev: You want to `return` early instead of returning the `response.json()` at the end of the function.

shahi645: i think i've got my python mixed up a little, just seems weird you're returning nothing here

jcox-dev: Yep yep. This `return` is useful because it says "I don't want this function to do anything more".

frontstage/controllers/iac_controller.py
gemmai95: Yeah the api returns 200 as it successfully got the iac code it's just not active

PR #414, opened 2018-10-23
Improve logging for alerts

gemmai95:
# Motivation and Context

Adds extra context to logs to help with alerting

# What has changed

# How to test?

# Links

Trello: https://trello.com/c/bdY6GFqq

### DIFF
--- a/frontstage/controllers/case_controller.py
+++ b/frontstage/controllers/case_controller.py
@@ -156,7 +156,7 @@ def get_eq_url(case_id, party_id, business_party_id, survey_short_name):
                     category=category,
                     description=f"Instrument {ci_id} launched by {party_id} for case {case_id}")
 
-    logger.debug('Successfully generated EQ URL', case_id=case_id, ci_id=ci_id, party_id=party_id,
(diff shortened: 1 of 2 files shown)


frontstage/controllers/case_controller.py
```diff
@@ -157,7 +157,7 @@ def get_eq_url(case_id, party_id, business_party_id, survey_short_name):
                     description=f"Instrument {ci_id} launched by {party_id} for case {case_id}")
 
     logger.debug('Successfully generated EQ URL', case_id=case_id, ci_id=ci_id, party_id=party_id,
-                 business_party_id=business_party_id, survey_short_name=survey_short_name)
+                 business_party_id=business_party_id, survey_short_name=survey_short_name, tx_id=payload['tx_id'])
```
gavinleeedwards: Would you be able to either log tx_id somewhere else as well or change the logging level of this line? Since this is logging at debug level, we still won't have any RAS production logs that link tx_id with party_id, survey etc.

jcox-dev: +1 to bumping this to INFO.

PR #419, opened 2018-11-20
Use smaller base layer for Docker image.

rjweeks:
# Motivation and Context
Reduce size of Docker image from 1.08Gb to 643Mb using smaller python:3.6-slim base layer.


rjweeks:
[TBL358 - review base docker images](https://trello.com/c/x0swxIs4/358-tbl358-review-base-docker-images)

PR #420, opened 2018-12-13
Sends to specific internal user instead of group, and defaults dev co…

AndrewTorrance:
# Motivation and Context
Internal users could not see the name of a user who had sent a reply to a message. It was showing Survey team. So in order to get context they needed to open up the thread which then marked it as read leading to messages not being replied to .
This change is to show the name of the last internal person who sent a message on that thread. So that internal users do not need to open up messages if they know another internal person is dealing with it .
 In order to attain this Ras-frontsgae should not display the internal user name to the respondent , but retain the internal user so that it can be set in the msg_to field at the point of sending the secure_message to the secure message service

# What has changed
This change only required a chance to ras-frontstage and the acceptance tests. Ras-frontstage now 
sends a reply back to the internal person who sent the message being replied to. if the respondent sends a message on a different subject it still gets sent to the survey team.

# How to test?
Manual testing: 
Run frontstage and response ops  . Send a new message from frontstage . Send a reply from response-op-ui . Validate that the user name is displayed in response ops in the conversation list view. And that ONS Survey team shown in the conversation list on frontstage. Reply from frontstage about 3 times. ( in order to prove later that frontstage sends the details of the last known internal user in a conversation and not just looking at the previous message in conversation)

Now go to response-ops ui and log in as a new user ( may have to get details from post gres , and auto standalone test users have numbers as first and last name ) . The to field should be the original internal user .Reply to the message from this new internal user . Validate the from is this new internal user .Go to Fronstage validate that they still dont see the name of the internal user. Reply . 

The automatic acceptance tests validate that the secure message is sent with the id of the internal user if known.Unit tests validate that the correct internal user id . (If a respondent starts a new conversation they send to'GROUP' as the to, if they reply it should be the user id of the internal person who sent the original)

# Links
https://trello.com/c/PGjxHXYA

### DIFF
--- a/config.py
+++ b/config.py
@@ -118,9 +118,9 @@ class DevelopmentConfig(Config):
     IAC_PASSWORD = os.getenv('IAC_PASSWORD', 'secret')
     IAC_AUTH = (IAC_USERNAME, IAC_PASSWORD)
 
-    OAUTH_URL = os.getenv('OAUTH_URL', 'http://localhost:8040')
-    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID', 'ons@ons.gov')
-    OAUTH_CLIENT_SECRET = os.getenv('OAUTH_CLIENT_SECRET', 'password')
+    OAUTH_URL = os.getenv('OAUTH_URL', 'http://localhost:8041')
+    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID', 'admin')
+    OAUTH_CLIENT_SECRET = os.getenv('OAUTH_CLIENT_SECRET', 'secret')
     OAUTH_BASIC_AUTH = (OAUTH_CLIENT_ID, OAUTH_CLIENT_SECRET)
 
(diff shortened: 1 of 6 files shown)


config.py
```diff
@@ -118,9 +118,9 @@ class DevelopmentConfig(Config):
     IAC_PASSWORD = os.getenv('IAC_PASSWORD', 'secret')
     IAC_AUTH = (IAC_USERNAME, IAC_PASSWORD)
 
-    OAUTH_URL = os.getenv('OAUTH_URL', 'http://localhost:8040')
-    OAUTH_CLIENT_ID = os.getenv('OAUTH_CLIENT_ID', 'ons@ons.gov')
-    OAUTH_CLIENT_SECRET = os.getenv('OAUTH_CLIENT_SECRET', 'password')
+    OAUTH_URL = os.getenv('OAUTH_URL', 'http://localhost:8041')
```
mfcrocker: Are these changes intended to be merged?

PR #423, opened 2018-12-19
Convert UTC unit tests

mfcrocker:
This just adds a unit test for the work done in https://github.com/ONSdigital/ras-frontstage/pull/422


mfcrocker:
@owencjones The sad path is already covered by test_generate_eq_url_incorrect_date_format. The bug fix just adds functionality when the date string provided isn't in the timezone that the server is, and this unit test tests that.

PR #424, opened 2018-12-20
582 redirect user to page requested after sign in

ryangrundy7:
# Motivation and Context
When a user clicks a link from an email, it would come up with an `JWTValidation` saying the user must sign in. What we want to happen is though, for the user to click the link and be directed to the sign in page. When they sign in they then get brought to the requested page they wanted.

# What has changed
- User now gets redirected to sign in page when attempting to visit a page which requires a login.
- Split up login method as it was too complex

# How to test?
- Send a message from internal to external. When trying to visit that conversation when not signed in, you should be brought to the sign in page. When signed in, you should be redirected to the conversation.
# Links
[Trello](https://trello.com/c/nLsI8fjn)

### DIFF
--- a/config.py
+++ b/config.py
@@ -147,4 +147,3 @@ class TestingConfig(DevelopmentConfig):
     ACCOUNT_SERVICE_LOG_OUT_URL = 'http://frontstage-url/sign-in/logout'
     EQ_URL = 'https://eq-test/session?token='
     JSON_SECRET_KEYS = open("./tests/test_data/jwt-test-keys/test_key.json").read()
-    SERVER_NAME = 'localhost:8082'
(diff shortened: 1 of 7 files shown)


config.py
```diff
@@ -147,3 +147,4 @@ class TestingConfig(DevelopmentConfig):
     ACCOUNT_SERVICE_LOG_OUT_URL = 'http://frontstage-url/sign-in/logout'
     EQ_URL = 'https://eq-test/session?token='
     JSON_SECRET_KEYS = open("./tests/test_data/jwt-test-keys/test_key.json").read()
+    SERVER_NAME = 'localhost:8082'
```
NickGrantONS: is this used by anything?

PR #426, opened 2019-01-24
Fix todo page to be more performant

ryangrundy7:
# Motivation and Context
Currently for some respondents on prod, they're unable to load the to-do page because it is hitting the timeout limit. To increase the performance of the to-do page, threads have been added to make multiple calls at once instead of sequentially.
# What has changed
- Added threads to the to-do page to make multiple calls at once and improve the performance

# How to test?
- As a respondent, enroll onto a load of surveys with different collection exercises so you fill up your to-do page on frontstage. The page should be quick to load. From my own testing, its been just over a second.
- Run the acceptance tests
- Just make sure that nothings broken on frontstage, as far as I can see, nothing has changed

# Links
[Trello](https://trello.com/c/aKL0leZQ)


frontstage/controllers/party_controller.py
```diff
@@ -229,32 +230,112 @@ def get_respondent_enrolments(party_id):
                 }
 
 
+class CachingDataThreadCreation(threading.Thread):
```
AndrewTorrance: I am not sure of the advantage of using a class to launch the thread as opposed to launching directly. Both will work here .Also this is now a general purpose thread wrapper. So its not tied to anything to do with caching . Shouldn't it be in a file of its own in in the common folder with a more generic class name and a Docstring to reflect its use as a generic thread wrapper

frontstage/controllers/party_controller.py
```diff
@@ -229,32 +230,112 @@ def get_respondent_enrolments(party_id):
...
+        threads.append(CachingDataThreadCreation(get_survey, cache_data, app.config, survey_id))
+        threads.append(CachingDataThreadCreation(get_collex, cache_data, app.config, survey_id))
+
+    for business_id in business_ids:
+        threads.append(CachingDataThreadCreation(get_case, cache_data, app.config, business_id, tag))
+        threads.append(CachingDataThreadCreation(get_party, cache_data, app.config, business_id))
+
+    for thread in threads:
+        thread.start()
+    for thread in threads:
```
AndrewTorrance: A comment here why join is done in a loop after start may be useful.

frontstage/controllers/party_controller.py
```diff
@@ -229,32 +230,112 @@ def get_respondent_enrolments(party_id):
...
+    for collection_instrument_id in collection_instrument_ids:
+        threads.append(CachingDataThreadCreation(get_collection_instrument, cache_data, app.config,
+                                                 collection_instrument_id))
+
+    for thread in threads:
+        thread.start()
+    for thread in threads:
+        thread.join()
+
+# This method has gone through a rewrite in an attempt to make the to-do page more performant. Before the rewrite, it
```
AndrewTorrance: When I read this I thought it was a comment about caching_data_for_collection_instrument. When its about get_survey_list_details_for_party . So I think it should be a docstring on get_survey_list_details_for_party ? I am not sure if the comment should reference old code , it should just say what the new code does

frontstage/controllers/party_controller.py
```diff
@@ -229,32 +230,112 @@ def get_respondent_enrolments(party_id):
...
+
+    cache_data = {'surveys': dict(),
+                  'businesses': dict(),
+                  'collexes': dict(),
+                  'cases': dict(),
+                  'instrument': dict()}
+
+    # These two will call the services to get responses and cache the data for later use.
+
+    caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag)
```
AndrewTorrance: Both caching_xxx_survey_xx and caching_ci_xxx do a start and join. Therefore we get survey data first then we get ci data . Why not have them both start and then do a join afterwards so these run asynchronously ? (unless the data from one is needed for the other ? But I dont THINK it does)

frontstage/common/eq_payload.py
```diff
@@ -46,7 +46,7 @@ def create_payload(self, case, party_id, business_party_id, survey):
         collex_event_dates = self._get_collex_event_dates(collex_id)
 
         # Party
-        party = party_controller.get_party_by_business_id(business_party_id, collection_exercise_id=collex_id)
+        party = party_controller.get_party_by_business_id(current_app.config, business_party_id, collection_exercise_id=collex_id)
```
AndrewTorrance: One for the team , but should we pass the entire config or just the values we need ?

frontstage/controllers/party_controller.py
ryangrundy7: It's a generic class that can take in any function with any amount of arguments so it makes sense to move it to the common folder. I'll change the name to ThreadWrapper 👍

frontstage/controllers/party_controller.py
ryangrundy7: We need this as two functions because before we can get the collection instruments, we need the cases to be cached to get the collection instrument ids from them. It might be a good idea to change the name of the first function but I kept it as survey_list because it's getting the important responses for the todo page.

frontstage/controllers/party_controller.py
ryangrundy7: Seems like a good idea to make it a docstring instead. Once this goes in everyone is gonna forget the old way of doing it.

frontstage/common/eq_payload.py
NickGrantONS: Fewer parameters = better

frontstage/controllers/party_controller.py
```diff
@@ -229,32 +230,116 @@ def get_respondent_enrolments(party_id):
                 }
 
 
+def set_enrolment_data(enrolment_data):
+    surveys_ids = set()
+    business_ids = set()
+    for enrolment in enrolment_data:
```
NickGrantONS: Maybe needs a comment here to explain that we're figuring out the unique set of surveys and parties, so that we don't get the data more than once

ryangrundy7: Sounds good 👍

frontstage/common/eq_payload.py
ryangrundy7: I spoke to Torrance and John about it, because we're only passing in two config values we agreed to just to pass them in instead of the entire object

PR #429, opened 2019-02-07
Add ability to fetch cases without IAC codes

ryangrundy7:
# Motivation and Context
To try and make the to-do page even quicker, an IAC boolean is passed to the case service to say if we want it added to the case or not. Passing in false will skip a DB call which should make things quicker.
# What has changed
- Pass in IAC boolean to case service

# How to test?
- Run with [Case service PR](https://github.com/ONSdigital/rm-case-service/pull/105)
- Run with Acceptance tests
- Open up the todo page and make sure everything is working such as accessing survey and completing it.
# Links
[Trello](https://trello.com/c/5xsPP92s)

### DIFF
--- a/config.py
+++ b/config.py
@@ -9,7 +9,7 @@ from distutils.util import strtobool
 class Config(object):
     DEBUG = False
     TESTING = False
-    VERSION = '1.11.0'
+    VERSION = '1.9.0'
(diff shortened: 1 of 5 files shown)


insacuri:
Code looks reasonable.
Is there no extra testing that needs to be done around this? or is it covered by the acceptance tests already?

PR #431, opened 2019-02-21
Already added survey message fix

ChristopherMShelton:
# Motivation and Context
Didn't work before, now it does.

# What has changed
Added true/false unit tests for the check to see if a user is already enrolled
Made a change to the check which removed unexpected behaviour (preventing people from enrolling who should be able to.)

# How to test?
1) Run the unit tests for add_survey_submit.py
2) Try and enrol to a survey on a users account that they are already enrolled to using a new enrolment code (should fail and show Already added survey' message).
3) Try and enrol to a survey that the user is not enrolled on (should succeed and show survey added message).
# Links
https://trello.com/c/gfNFRVsR/569-569-add-validation-for-respondents-adding-another-survey-to-their-enrolment

### DIFF
--- a/frontstage/views/surveys/add_survey_submit.py
+++ b/frontstage/views/surveys/add_survey_submit.py
@@ -40,8 +40,8 @@ def add_survey_submit(session):
 
         already_enrolled = None
         if is_business_enrolled(info['associations'], case['caseGroup']['surveyId']):
-            logger.error('User tried to enrol onto a survey they are already enrolled on',
-                         case_id=case_id, party_id=party_id)
+            logger.info('User tried to enrol onto a survey they are already enrolled on',
+                        case_id=case_id, party_id=party_id)
             already_enrolled = True
         else:
             case_controller.post_case_event(case_id,
--- a/frontstage/views/surveys/add_survey_submit.py
+++ b/frontstage/views/surveys/add_survey_submit.py
@@ -58,17 +58,17 @@ def add_survey_submit(session):
 
     logger.info('Successfully retrieved data for confirm add organisation/survey page',
                 case_id=case_id, party_id=party_id)
-    return redirect(url_for('surveys_bp.get_survey_list', _anchor=(business_party_id, added_survey_id), _external=True, business_party_id=business_party_id,
+    return redirect(url_for('surveys_bp.get_survey_list', _anchor=(business_party_id, added_survey_id), _external=True,
+                            business_party_id=business_party_id,
                             survey_id=added_survey_id, tag='todo', already_enrolled=already_enrolled))
 
 
 def is_respondent_and_business_enrolled(associations, survey_id, party_id):
     """
-    This function checks that the business AND the respondent are enrolled in a survey. We first check if the respondent
-    is enrolled, then if not, we check that the business is enrolled on the survey. A business can have multiple
-    respondents enrolled for a survey, and a respondent can be enrolled for multiple surveys. If we only checked if the
-    business was enrolled, then there could only ever be one respondent that could be able to answer the survey, which
-    isn't the desired functionality.
+    This function checks whether or not a respondent's associated party is associated with the survey. If yes,
+    we then check if the survey_id of the survey we are trying to add is found within the enrolments of the respondent.
+    We do this because a respondent can be responding on behalf of several businesses, and a business can have several
+    people enrolled on the same survey.
 
     :param associations: List of respondents and their enrolled surveys
     :param survey_id: id of the added survey
@@ -81,3 +81,4 @@ def is_respondent_and_business_enrolled(associations, survey_id, party_id):
                 if survey['surveyId'] == survey_id:
                     return True
     return False
+
--- a/frontstage/views/surveys/add_survey_submit.py
+++ b/frontstage/views/surveys/add_survey_submit.py
@@ -39,7 +39,7 @@ def add_survey_submit(session):
         info = party_controller.get_party_by_business_id(business_party_id, collection_exercise_id)
 
         already_enrolled = None
-        if is_business_enrolled(info['associations'], case['caseGroup']['surveyId']):
+        if is_business_enrolled(info['associations'], case['caseGroup']['surveyId'], party_id):
             logger.info('User tried to enrol onto a survey they are already enrolled on',
                         case_id=case_id, party_id=party_id)
             already_enrolled = True
@@ -62,7 +62,19 @@ def add_survey_submit(session):
                             survey_id=added_survey_id, tag='todo', already_enrolled=already_enrolled))
 
 
-def is_business_enrolled(associations, survey_id):
-    return any(survey for info in associations
-               for survey in info['enrolments']
-               if survey['surveyId'] == survey_id)
+def is_business_enrolled(associations, survey_id, party_id):
+    """
+    Makes two checks as a business can be enrolled  on a survey but not the respondent, which would causes a single
+    check to break if another respondent from the business had previously enrolled on that survey
+
+    :param associations: List of respondents and their enrolled surveys
+    :param survey_id: id of the added survey
+    :param party_id: id of the respondent
+    :return: True if respondent and the business is enrolled on the survey and false otherwise
+    """
+    for info in associations:
+        if info['partyId'] == party_id:
+            for survey in info['enrolments']:
+                if survey['surveyId'] == survey_id:
+                    return True
+    return False
--- a/frontstage/views/surveys/add_survey_submit.py
+++ b/frontstage/views/surveys/add_survey_submit.py
@@ -1,3 +1,4 @@
+
 import logging
 
 from flask import redirect, request, url_for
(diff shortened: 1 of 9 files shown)


frontstage/views/surveys/add_survey_submit.py
```diff
@@ -49,4 +59,22 @@ def add_survey_submit(session):
...
+    Makes two checks as a business can be enrolled  on a survey but not the respondent, which would causes a single
+    check to break if another respondent from the business had previously enrolled on that survey
+
+    :param associations: List of respondents and their enrolled surveys
+    :param survey_id: id of the added survey
+    :param party_id: id of the respondent
+    :return: True if respondent and the business is enrolled on the survey and false otherwise
+    """
+    for info in associations:
+        if info['partyId'] == party_id:
```
NickGrantONS: a more Pythonic way of doing this might be to put the if part into the for loop above

frontstage/views/surveys/add_survey_submit.py
```diff
@@ -49,4 +59,22 @@ def add_survey_submit(session):
...
+            for survey in info['enrolments']:
+                if survey['surveyId'] == survey_id:
```
NickGrantONS: a more Pythonic way of doing this might be to put the if part into the for loop above

frontstage/views/surveys/add_survey_submit.py
```diff
@@ -49,4 +59,22 @@ def add_survey_submit(session):
...
+    """
+    Makes two checks as a business can be enrolled  on a survey but not the respondent, which would causes a single
+    check to break if another respondent from the business had previously enrolled on that survey
+
+    :param associations: List of respondents and their enrolled surveys
+    :param survey_id: id of the added survey
+    :param party_id: id of the respondent
+    :return: True if respondent and the business is enrolled on the survey and false otherwise
+    """
+    for info in associations:
```
insacuri: Looking at it, it should be `for association in associations`.  Info is too generic

frontstage/views/surveys/add_survey_submit.py
insacuri: Do you have an example of what it would look like?  I might just be slow this morning but I can't think of nicer way to do this

frontstage/views/surveys/surveys_list.py
```diff
@@ -28,7 +29,9 @@ def get_survey_list(session, tag):
     if tag == 'todo':
         response = make_response(render_template('surveys/surveys-todo.html',
                                                  sorted_surveys_list=sorted_survey_list,
-                                                 added_survey=True if business_id and survey_id else None))
+                                                 added_survey=True if business_id and survey_id and not already_enrolled else None,
```
insacuri: This line is starting to get a bit long now.  Either put on the line above `added_survey = True if business_id and survey_id and not already_enrolled else None` (or something to that effect), or you could maybe roll this into a function? (something like `added_survey = is_survey_added()`on this line or something like that).
I mention rolling this into a function because this added survey is taking multiple bits of data to come to its answer, but it's not clear at a glance what each bit means.  That could also be me not understanding fully what's going on

tests/app/views/surveys/test_add_survey_submit.py
```diff
@@ -66,3 +91,21 @@ def test_add_survey_submit_fail(self, decrypt_enrolment_code, get_iac_from_enrol
 
         self.assertEqual(response.status_code, 500)
         self.assertTrue('Error 500 - Server error'.encode() in response.data)
+
+    def test_already_enrolled(self, ):
```
insacuri: Function definition should be `def test_already_enrolled(self):`

tests/app/views/surveys/test_add_survey_submit.py
```diff
@@ -66,3 +91,21 @@ def test_add_survey_submit_fail(self, decrypt_enrolment_code, get_iac_from_enrol
...
+        with open('tests/test_data/party/business_party.json') as j:
```
insacuri: j should be literally anything else.  Something like `business_party_json` would be a longer name, but you would actually know roughly what you were referring to when looking through the code.
Single letter variables should be very sparsely used (usually for loop counters)

tests/app/views/surveys/test_add_survey_submit.py
```diff
@@ -66,3 +91,21 @@ def test_add_survey_submit_fail(self, decrypt_enrolment_code, get_iac_from_enrol
...
+            data = json.load(j)
+        survey = 'cb8accda-6118-4d3b-85a3-149e28960c54'
+
+        party_id = 'f956e8ae-6e0f-4414-b0cf-a07c1aa3e37b'
+
+        self.assertTrue(is_business_enrolled(data['associations'], survey, party_id))
+
+    def test_not_already_enrolled(self, ):
```
insacuri: Same as above with function parameters

frontstage/views/surveys/add_survey_submit.py
ChristopherMShelton: Yea the way I tried didn't work and the other ways I've found to shorten it make it pretty unreadable

frontstage/views/surveys/add_survey_submit.py
```diff
@@ -49,4 +59,22 @@ def add_survey_submit(session):
     logger.info('Successfully retrieved data for confirm add organisation/survey page',
                 case_id=case_id, party_id=party_id)
     return redirect(url_for('surveys_bp.get_survey_list', _anchor=(business_party_id, added_survey_id), _external=True, business_party_id=business_party_id,
-                            survey_id=added_survey_id, tag='todo'))
+                            survey_id=added_survey_id, tag='todo', already_enrolled=already_enrolled))
+
+
+def is_business_enrolled(associations, survey_id, party_id):
+    """
+    Makes two checks as a business can be enrolled  on a survey but not the respondent, which would causes a single
```
insacuri: I don't think this docstring properly describes what is going on.  I think it should be something like this... but I don't know, if you could get 3rd person to make sure it makes sense then that would be helpful as this what this one function does is really important and the Why of this function is fairly important to explain to future people who come along to maintain this in the future.
`This function checks that the business AND the respondent are enrolled in a survey.  We first check if the respondent is enrolled, then if not, we check that the business is enrolled on the survey.  A business can have multiple respondents enrolled for a survey, and a respondent can be enrolled for multiple surveys.
If we only checked if the business was enrolled, then there could only ever be one respondent that could be able to answer the survey, which isn't the desired functionality.`

frontstage/views/surveys/add_survey_submit.py
```diff
@@ -49,4 +59,22 @@ def add_survey_submit(session):
     logger.info('Successfully retrieved data for confirm add organisation/survey page',
                 case_id=case_id, party_id=party_id)
     return redirect(url_for('surveys_bp.get_survey_list', _anchor=(business_party_id, added_survey_id), _external=True, business_party_id=business_party_id,
-                            survey_id=added_survey_id, tag='todo'))
+                            survey_id=added_survey_id, tag='todo', already_enrolled=already_enrolled))
+
+
+def is_business_enrolled(associations, survey_id, party_id):
```
insacuri: This function name should probably be `is_respondent_and_business_enrolled` as it's really checking 2 things.

frontstage/views/surveys/add_survey_submit.py
owencjones:
```Ensures that
* Individuals are a associated with the business
* Business is enrolled in a survey```

?

Comment (insacuri) [reply frontstage/views/surveys/add_survey_submit.py]: There's another thing to consider for this.  There are technically 2 'errors' that occur with this function.  A user could not be associated with the business, and the business could not be enrolled with the survey.  Currently, if either of those are false then we just go 'nah, you can't be enrolled on this survey'.  Maybe not now, but we should be more helpful with our error messages.

File: frontstage/views/surveys/add_survey_submit.py [hunk 10]
```diff
@@ -49,4 +59,25 @@ def add_survey_submit(session):
...
+
+    :param associations: List of respondents and their enrolled surveys
+    :param survey_id: id of the added survey
+    :param party_id: id of the respondent
+    :return: True if respondent and the business is enrolled on the survey and false otherwise
+    """
+    for association in associations:
+        if association['partyId'] == party_id:
+            for survey in association['enrolments']:
+                if survey['surveyId'] == survey_id:
```
NickGrantONS:
```suggestion
          for association in (for association in associations if association['partyId'] == party_id):
            for survey in (for survey in association['enrolments'] if survey['surveyId'] == survey_id):
```

insacuri: I think maybe trying something like this could be the best way to write it.  
```for association in assocations:
  if is_respondent_associated_with_busienss(association, party_id):
    if is_business_enrolled_in_survey(var1, var2)```

Having 2 separate functions that explain in english what they're doing  will be more useful then seeing nested for loops I think.  Then we can have this function tie everything together, along with any explanations that are needed.
This might look awful when implemented?  I don't know, but I think it'd help reduce the complexity of what we're doing here.

NickGrantONS: your suggestion is better than mine

owencjones: I like this approach, not least because splitting functions leads to easier testing, which in turn leads to more accurate testing.

PR #434, opened 2019-03-14
Update packages to fix security issue

insacuri:
# Motivation and Context
There is a security issue against this repo for pyyaml.  On the 13th March, they released 5.1, which means we can finally fix this issue.  In addition to this, a number of other packages have been updated.


insacuri:
This raises a larger issue with redis being updated.  Closing until we figure out what to do

PR #435, opened 2019-03-20
restricted enrolments as seen in the UI to those for started collex

AndrewTorrance:
# Motivation and Context
Ras frontstage was erroring in the case where a user was enrolled on a collection exercise but that collection exercise had not started .  

# What has changed
The error was due to using enrolment survey id as an index into collection exercises , but collection exercises were restricted to those that had started . 
Changed the generator to not consider enrolments where the collex had not yet started . 

# How to test?
Run unit tests . 

# Links
https://trello.com/c/W59SAyNd

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ services:
     - redis-server
 
 install:
-  - pip install pipenv==8.3.2
+  - pip install pipenv
   - pipenv install --dev --deploy
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 sudo: required
-dist: trusty
 python: '3.6'
 
 # before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,10 @@ script:
   - APP_SETTINGS=TestingConfig pipenv run pytest --cov=frontstage --cov-report xml --ignore=node_modules
   - pipenv run coverage report
 
+env:
+  global:
+    - PIPENV_IGNORE_VIRTUALENVS=1
+
 after_success:
   - pipenv run codecov
 
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -234,7 +234,14 @@ def get_respondent_enrolments(party_id):
 
 
 def get_respondent_enrolments_for_known_collex(enrolment_data, cache_collex):
-    """remove any enrolments that are for surveys not in already started collex."""
+    """ Needed because enrolment_data includes not started enrolments ,
+    but cache_collex only contains started collex. Hence indexing collex by enrolment[survey] causes a 500
+
+    :param enrolment_data: A list of enrolments.
+    :param cache_collex: A list of not started collection exercises.
+    :return: list of enrolments corresponding to the known collection exercises
+    """
+
     enrolments = []
     for enrolment in enrolment_data:
         if enrolment['survey_id'] in cache_collex:
@@ -329,7 +336,6 @@ def get_survey_list_details_for_party(party_id, tag, business_party_id, survey_i
                   'instrument': dict()}
 
     # These two will call the services to get responses and cache the data for later use.
-
     caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag)
     caching_data_for_collection_instrument(cache_data)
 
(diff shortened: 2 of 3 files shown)


tests/app/controllers/test_party_controller.py
```diff
@@ -194,6 +194,21 @@ def test_lock_respondent_account_fail(self):
...
+        if we have a corresponding collex."""
+
+        collex = {"survey1": "collex1", "survey3": "collex3"}
+
+        enrolment_data = [{"survey_id": "survey1", "enrolment_data": "enrolment1"},
+                          {"survey_id": "survey2", "enrolment_data": "enrolment2"},
+                          {"survey_id": "survey3", "enrolment_data": "enrolment3"}]
+
+        result = get_respondent_enrolments_for_known_collex(enrolment_data, collex)
+        self.assertEquals(len(result), 2)
```
ryangrundy7: Change this to `assertEqual` to get rid of warning 🎨

frontstage/controllers/party_controller.py
```diff
@@ -304,8 +315,7 @@ def get_survey_list_details_for_party(party_id, tag, business_party_id, survey_i
     :survey_id: This is the surveys uuid
 
     """
-    enrolment_data = get_respondent_enrolments(party_id)
-
+    enrolment_data = list(get_respondent_enrolments(party_id))
```
insacuri: `get_respondent_enrolments` returns either a dictionary or None, is that correct?

insacuri: `get_respondent_enrolments` yields a generator (still don't get why we bothered with this) which is iterable already, so you shouldn't have to make it a list I think.
Is the adding of `list()` there to handle the occiasion when the respondent has no enrolments?

frontstage/controllers/party_controller.py
```diff
@@ -233,6 +233,17 @@ def get_respondent_enrolments(party_id):
                 }
 
 
+def get_respondent_enrolments_for_known_collex(enrolment_data, cache_collex):
+    """remove any enrolments that are for surveys not in collex
```
insacuri: This docstring is missing some punctuation or something.  I don't really understand it

.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 sudo: required
-dist: trusty
```
insacuri: Does this need to be removed?  or was it unnecessary to begin with?

frontstage/controllers/party_controller.py
```diff
@@ -324,8 +334,7 @@ def get_survey_list_details_for_party(party_id, tag, business_party_id, survey_i
     caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag)
     caching_data_for_collection_instrument(cache_data)
 
-    for enrolment in get_respondent_enrolments(party_id):
```
insacuri: Before this was calling the same function twice within a few lines of each other?  Ew.  Your change makes this much better

frontstage/controllers/party_controller.py
AndrewTorrance: It was made a list as it was getting a called twice. Under the hood it calls out to party . So making it a list means that we dont have so many http requests which are costly.

frontstage/controllers/party_controller.py
AndrewTorrance: There was another line in the docstring. Updated for clarity

.travis.yml
AndrewTorrance: Travis was erroring due to the ongoing numpy issue. This was an attempt to resolve by making travis.yml more like party service  that was not having these issues

frontstage/controllers/party_controller.py
```diff
@@ -233,6 +233,15 @@ def get_respondent_enrolments(party_id):
                 }
 
 
+def get_respondent_enrolments_for_known_collex(enrolment_data, cache_collex):
+    """remove any enrolments that are for surveys not in already started collex."""
```
insacuri: Sorry to be picky, but the Why here is important.  Mentioning the problem we're fixing will clue future people as to why it's in here

frontstage/controllers/party_controller.py
insacuri: Totally fair.  Could you do something for me and raise a card to look into changing that bit to not return a generator, but a normal list.  The return value for this function is fairly small so I don't understand why we were using yield to begin with.

PR #439, opened 2019-04-08
Display suitable error messages to respondents us223

ChristopherMShelton:
# Motivation and Context
Replaced error templates with new versions that do not present as errors to the user, removing technical 'jargon'.

# What has changed
Changed error templates
Added flash message to sign in page for when user is redirected after signing out

# How to test?
Run unit tests
Change timeout from 3600 to something like 30, after being timed out you will be redirected to the sign in page and an info panel should appear.

# Links
https://trello.com/c/EiwEvwud/811-display-suitable-error-messages-to-respondents-5


tests/app/test_registration.py
```diff
@@ -96,7 +96,7 @@ def test_enter_enrolment_code_fail(self, mock_object):
         response = self.app.post('/register/create-account', data={'enrolment_code': enrolment_code}, follow_redirects=True)
 
         self.assertEqual(response.status_code, 500)
-        self.assertTrue('Server error'.encode() in response.data)
+        self.assertTrue('An error has occurred'.encode() in response.data)
 
     # ============== CONFIRM ORG AND SURVEY ===============
```
owencjones: Not your fault, but again wondering why we don't lint tests

tests/app/test_registration.py
```diff
@@ -252,7 +252,7 @@ def test_create_account_get_page_with_inactive_enrolment_code(self, mock_object)
                                 follow_redirects=True)
 
         self.assertEqual(response.status_code, 500)
-        self.assertTrue('Server error'.encode() in response.data)
+        self.assertTrue('An error has occurred'.encode() in response.data)
```
owencjones: At this stage, you've referred to `'An error has occurred'.encode()` so often that I think declaring a variable of it earlier in test setup would be worthwhile to make for ease of changing it.

ChristopherMShelton: Added to config, just a method that returns a string now but a string file would be better in future

PR #442, opened 2019-04-10
Remove html5 attribs from forms

owencjones:
# Motivation and Context
The use of HTML5 input attributes like `required` prevents the user experience being in line with that desired.

# What has changed
* Wrote custom validators that extend WTForm validators and removes html5 attribs
* Imported these validators in form models, instead of the WTForm validators
* Wrote tests

# How to test?
* Run through process of signing up as a respondent and confirm that in source code, no input has `required` attribute (can do this using inspect element)
* Run through other forms in frontstage, do the same check
* Acceptance tests

# Links
https://trello.com/c/lPtTcThh

### DIFF
--- a/frontstage/common/validators.py
+++ b/frontstage/common/validators.py
@@ -13,7 +13,7 @@ class InputRequired(IR):
     Example sage: element = StringField('name', [ FrontstageInputRequired(message="Error message to show") ]
     """
     def __init__(self, message=None):
-        super(IR, self).__init__()
+        super().__init__()
         self.field_flags = remove_required(self.field_flags)
 
 
@@ -22,7 +22,7 @@ class DataRequired(DR):
     WTForms validator replacement that requires data, but doesn't add 'required' attribute to rendered html
     """
     def __init__(self, message=None):
-        super(DR, self).__init__()
+        super().__init__()
         self.field_flags = remove_required(self.field_flags)
 
 
--- a/frontstage/common/validators.py
+++ b/frontstage/common/validators.py
@@ -1,0 +1,34 @@
+from wtforms.validators import InputRequired as IR
+from wtforms.validators import DataRequired as DR
+
+"""
+A collection of validators for use with WTForms forms
+"""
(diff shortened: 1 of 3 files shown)


frontstage/common/validators.py
```diff
@@ -0,0 +1,34 @@
...
+    """
+    def __init__(self, message=None):
+        super(DR, self).__init__()
+        self.field_flags = remove_required(self.field_flags)
+
+
+def remove_required(flags):
+    if 'required' in flags:
+        n = flags.index('required')
+        return flags[: n] + flags[n + 1:]
```
mfcrocker: Will this work if `required` is the last item in `flags`?

owencjones: Yeah, if you have a `List` and slice it with an out or range index as the `end` position, it corrects to max index instead.  This even works with an empty `List`.

See example below from Python REPL:
```
>>> l = []
>>> l2 = l[:3]
>>> l2
[]```

Two `List`s, (both alike in dignity) one empty, and one created from the out-of-bounds reference to a slice of the other.  Just returns what it can.

To address this concern further, I've added two extra tests that confirm that it is removed if it's either in the middle of, or at the end of, the list.

Happy to mark this resolved?

Comment (mfcrocker) [reply frontstage/common/validators.py]: Awesome :)

File: tests/test_validators.py [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+        output = remove_required(test_tuple)
+
+        self.assertFalse('required' in output)
+
+    def test_should_remove_a_required_member_if_its_in_the_middle_of_the_list(self):
+        test_tuple = ('otheritem', 'required', 'other', 'somethingelse')
+        output = remove_required(test_tuple)
+
+        self.assertFalse('required' in output)
+
```
ryangrundy7: Theres whitespace on this line which is causing linting error

owencjones: Resolved.

PR #444, opened 2019-04-23
Extract strings

owencjones:
# Motivation and Context
Strings within files that are to be shown to the user can now be wrapped in a `gettext` style function that allows them to be retrieved from a strings file, if an override for the string exists.

This means that UX decisions to alter strings can be made by non-developers, without having to allow them to edit code.

`gettext` itself wasn't used because of it's reliance on the binary compiled format, and extended tooling, which is simple for developers but not for UX team members.  This version is less efficient, but stores all values in a JSON file, and initialises them in memory at bootstrapping of the app.  With the available memory, we can easily accomodate this.

As a bonus, the `gettext` drop in replacement can support string sets for locales, so would make translation of frontstage easier if it becomes necessary in future, which seems like a possibility.

I have also written a small command line tool that will extract the strings from a python file using a simple regex.  This will allow developers to easily create the strings files necessary.

This has been framed in the context of i18n, but is initially intended to simplify alteration of string messages presented to users - locale based options are just a quick-win feature I was able to add.

# What has changed
* Added a class module that can be included in a python file to load strings list for that file
* Added tests for this
* Wrapped strings in the `models.py` file, which defines forms used in the app, for ease of use
* Changed some strings in the first instance, to comply better with GDS
* Added CLI script to find and format wrapped strings in files

# How to test?
1. Start ras-frontstage, with all necessary dependencies also running
2. Ensure forms work
3. Take a string value from frontstage and redefine it in its strings file (`frontstage/i18n/form_messages.json`)
4. Re-start frontstage and confirm that string changes appropriately in UI
5. Usual tests (unit & acceptance) to confirm nothing has broken
6. Optionally, test the CLI util against `models.py`, but bear in mind that it's very much a beta feature - not to be heavily relied upon, only to offer help.

# Links
https://trello.com/c/qCcZrkz5


get_strings.py
```diff
@@ -0,0 +1,62 @@
+from json import dumps
```
mfcrocker: Is the repo root the best location for this utility?

frontstage/i18n/translations.py
```diff
@@ -0,0 +1,49 @@
...
+
+        try:
+            if not translation_file.endswith('.json'):
+                translation_file += '.json'
+
+            file_name = join(self.translation_path, translation_file)
+            with open(file_name) as f:
+                self.translations = load(f)
+        except FileNotFoundError as exc:
+            logger.error(f'Instantiating translation with file "{file_name}"')
```
rhiGriff: We shouldn't be using f strings in logs it should be key value pairs

gavinleeedwards: Also, would you be able to change this error message please to make it clear that the file isn't there, just to make it easier for anyone seeing the error in logs.

owencjones: @rhiGriff Can you provide an example - naturally I have no access to the page you linked.  Any chance of a snippet to demonstrate what you might write?

I'm relatively sure that if all logging needs to be KV of some sort, that Frontstage basically needs all log lines rewritten, so might need a ticket for that too.

rhiGriff: I did some work on it before but yeah it will need looking at again.

Here is an example:

 logger.info('User account is locked on the OAuth2 server', party_id=party_id)

https://github.com/ONSdigital/ras-frontstage/blob/91d1ee655f037b6c817fcedd4f5f99279b5ca140/frontstage/views/sign_in/sign_in.py#L37

get_strings.py
owencjones: It was more that I didn't really know of a better place.  Could create a folder for it if need be

frontstage/models.py
```diff
@@ -57,30 +63,30 @@ def validate_phone_number(form, field):
             input_number = phonenumbers.parse(field.data, 'GB')  # Tell the parser we are looking for a GB number
 
             if not phonenumbers.is_possible_number(input_number):
-                raise ValidationError('This should be a valid phone number between 9 and 15 digits')
+                raise ValidationError(_('This should be a valid phone number between 9 and 15 digits'))
 
             if not phonenumbers.is_valid_number(input_number):
-                raise ValidationError('Please use a valid UK number e.g. 01632 496 0018.')
+                raise ValidationError(_('Please use a valid UK number e.g. 01632 496 0018.'))
```
necrophonic: I thought all "please"'s where being removed?

owencjones: Allowing non techs to do this is exactly what this ticket achieves - We haven't been given any chosen strings yet.

necrophonic: Ah perfect 👍

PR #446, opened 2019-04-25
Added email re-entry field

berroar:
# Motivation and Context
On request from the business area we need to add a second email box on the create user page to check that the email being entered is correct.

# What has changed
Added an additional email field in register/enter-your-details.html. 
Updated the email field validation to ensure that the input entered into the original email and re-typed email fields match. 
Added a new unit test to check that the new field displays the correct error message when an email address is re-typed incorrectly.

# How to test?
Run the unit and acceptance tests.

# Links
https://trello.com/c/Gfbhw78e/567-us567-confirm-email-address-when-creating-a-new-account-5


config.py
```diff
@@ -36,7 +36,7 @@ class Config(object):
     JSON_SECRET_KEYS = os.getenv('JSON_SECRET_KEYS')
 
     REDIS_HOST = os.getenv('REDIS_HOST', 'localhost')
-    REDIS_PORT = os.getenv('REDIS_PORT', 6379)
+    REDIS_PORT = os.getenv('REDIS_PORT', 7379)
```
insacuri: I can't remember why, but I think this change can't be checked in as something else uses it?  Could be wrong though

berroar: I had to change it to run the tests and forgot to revert back to the original. Will revert to the original.

frontstage/models.py
```diff
@@ -34,7 +34,13 @@ class RegistrationForm(FlaskForm):
                                 validators=[InputRequired('Email address is required'),
                                             Email(message='Invalid email address'),
                                             Length(max=254,
-                                                   message='Your email must be less than 254 characters')])
+                                                   message='Your email must be less than 254 characters'),
+                                            EqualTo('email_address_confirm',
+                                                     message=app.config['EMAIL_MATCH_ERROR_TEXT'])
```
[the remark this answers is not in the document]
berroar: I've added the new string extraction methods :)

PR #447, opened 2019-04-25
403 message thread not found status page

ChristopherMShelton:
# Motivation and Context
Added a template for when a user follows a direct link to a secure message for an account other than the one they are logged in to.

This is so that users with multiple respondent accounts aren't confused as to why they cant access a message.

# What has changed
Added page template
Created new exception which results in a page render
Implemented unit tests for new functionality

# How to test?
Run frontstage unittests
Can manually test by sending a message to a respondent, grabbing the link from frontstage and trying to access it after logging into another account.

# Links
https://trello.com/c/iWD6OxY9/824-403-message-thread-not-found-status-page-3

### DIFF
--- a/frontstage/controllers/conversation_controller.py
+++ b/frontstage/controllers/conversation_controller.py
@@ -10,7 +10,7 @@ from structlog import wrap_logger
 from urllib3 import Retry
 
 from frontstage.common.session import SessionHandler
-from frontstage.exceptions.exceptions import ApiError, AuthorizationTokenMissing, NoMessagesError
+from frontstage.exceptions.exceptions import ApiError, AuthorizationTokenMissing, NoMessagesError, IncorrectAccountAccessError
 
 
 logger = wrap_logger(logging.getLogger(__name__))
@@ -34,11 +34,14 @@ def get_conversation(thread_id):
         response = session.get(url, headers=headers)
         try:
             response.raise_for_status()
-        except HTTPError:
-            raise ApiError(logger, response,
-                           log_level='error',
-                           message='Thread retrieval failed',
-                           thread_id=thread_id)
+        except HTTPError as exception:
+            if exception.response.status_code == 404:
+                raise IncorrectAccountAccessError()
+            else:
+                raise ApiError(logger, response,
+                               log_level='error',
+                               message='Thread retrieval failed',
+                               thread_id=thread_id)
 
     logger.debug('Thread retrieval successful', thread_id=thread_id)
 
--- a/frontstage/controllers/conversation_controller.py
+++ b/frontstage/controllers/conversation_controller.py
@@ -35,8 +35,8 @@ def get_conversation(thread_id):
         try:
             response.raise_for_status()
         except HTTPError as exception:
-            if exception.response.status_code == 404:
-                raise IncorrectAccountAccessError()
+            if exception.response.status_code == 403:
+                raise IncorrectAccountAccessError(message='Access not granted for thread', thread_id=thread_id)
             else:
                 raise ApiError(logger, response,
                                log_level='error',
--- a/frontstage/error_handlers.py
+++ b/frontstage/error_handlers.py
@@ -8,7 +8,7 @@ from werkzeug.utils import redirect
 
 from frontstage import app
 from frontstage.common.session import SessionHandler
-from frontstage.exceptions.exceptions import ApiError, InvalidEqPayLoad, JWTValidationError
+from frontstage.exceptions.exceptions import ApiError, InvalidEqPayLoad, JWTValidationError, IncorrectAccountAccessError
 
 logger = wrap_logger(logging.getLogger(__name__))
 
@@ -65,3 +65,9 @@ def server_error(error):
 def eq_error(error):
     logger.error('Failed to generate EQ URL', error=error.message, url=request.url, status_code=500)
     return render_template('errors/500-error.html'), 500
+
+
+@app.errorhandler(IncorrectAccountAccessError)
+def secure_message_forbidden_error(error):
+    logger.error('Attempt to access secure message without correct session permission')
+    return render_template('errors/403-incorrect-account-error.html')
--- a/frontstage/error_handlers.py
+++ b/frontstage/error_handlers.py
@@ -69,5 +69,5 @@ def eq_error(error):
 
 @app.errorhandler(IncorrectAccountAccessError)
 def secure_message_forbidden_error(error):
-    logger.error('Attempt to access secure message without correct session permission')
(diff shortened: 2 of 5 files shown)


frontstage/error_handlers.py
```diff
@@ -65,3 +65,9 @@ def server_error(error):
 def eq_error(error):
     logger.error('Failed to generate EQ URL', error=error.message, url=request.url, status_code=500)
     return render_template('errors/500-error.html'), 500
+
+
+@app.errorhandler(IncorrectAccountAccessError)
+def secure_message_forbidden_error(error):
+    logger.error('Attempt to access secure message without correct session permission')
```
gavinleeedwards: I think this should be info level rather than error - it's a user error rather than one we need to know about.

frontstage/templates/errors/403-incorrect-account-error.html
```diff
@@ -0,0 +1,15 @@
+{% import 'partials/section.html' as section %}
+{% extends "layouts/_twocol.html" %}
+
+{% block page_title %}ONS Business Surveys{% endblock %}
+
+{% block main %}
+    <h1 class="saturn">Access to this page is forbidden</h1>
+    <p>The page you are trying to view is not for this account.</p>
+	<p>If you have another ONS Business Surveys account, check that you are signed in to the right one.</p>
+	<p>If the problem continues, contact the survey enquiries helpline fo further support</p>
```
gavinleeedwards: Typo - `fo further support`

frontstage/error_handlers.py
gavinleeedwards: Also, could you add more context to the error message please (userid, thread_id, whatever you have), as the message isn't much help on its own

frontstage/controllers/conversation_controller.py
```diff
@@ -34,11 +34,14 @@ def get_conversation(thread_id):
         response = session.get(url, headers=headers)
         try:
             response.raise_for_status()
-        except HTTPError:
-            raise ApiError(logger, response,
-                           log_level='error',
-                           message='Thread retrieval failed',
-                           thread_id=thread_id)
+        except HTTPError as exception:
+            if exception.response.status_code == 404:
```
mfcrocker: I think this should be 403

frontstage/exceptions/exceptions.py
```diff
@@ -93,3 +93,7 @@ class UserDoesNotExist(Exception):
     def __init__(self, message):
         super().__init__()
         self.message = message
+
+
+class IncorrectAccountAccessError(Exception):
+        pass
```
mfcrocker: Why isn't this implemented?

tests/app/test_secure_message.py
```diff
@@ -166,3 +167,16 @@ def test_create_message_post_no_ru_ref(self):
                                  data=self.message_form, headers=self.headers, follow_redirects=True)
 
         self.assertEqual(response.status_code, 400)
+
+    @requests_mock.mock()
+    def test_get_thread_wrong_account(self, mock_request):
+        mock_request.get(url_get_thread, status_code=404, json={'messages': [conversation_json], 'is_closed': False})
```
mfcrocker: Same again on 403 v 404

frontstage/exceptions/exceptions.py
ChristopherMShelton: It just doesn't need to do anything when it happens other than return the template, which is done elsewhere

frontstage/controllers/conversation_controller.py
owencjones: It _should be_, but I think Chris right that it's actually not.

The service, IIRC, throws back 'Not found' so that fuzzing the server to find real values won't produce different results, but I may be wrong.

@ChristopherMShelton to confirm?

frontstage/exceptions/exceptions.py
owencjones: @mfcrocker Essentially the same as a abstract class - the value is in identifying the error, that said...

@ChristopherMShelton Shouldn't we be adding something to this to log the occurence?

frontstage/templates/errors/403-incorrect-account-error.html
```diff
@@ -0,0 +1,15 @@
+{% import 'partials/section.html' as section %}
```
owencjones: Is this used?

frontstage/controllers/conversation_controller.py
insacuri: I'm with Matt, this should be a 403.  If the service returns 404 if there is a message/thread but they don't have access, then it should be fixed in secure-message, worked around here.

frontstage/exceptions/exceptions.py
insacuri: We've done this in a bunch of other places in other repos.  I'm fine with it not having logging here, but in the places that we ARE throwing this exception (just the 1 place at the minute), we should be logging an error

EDIT: Basically what owen said ^^

frontstage/controllers/conversation_controller.py
insacuri: I think the bit of code that needs to change can be found https://github.com/ONSdigital/ras-secure-message/blob/master/secure_message/repository/retriever.py#L190

At the minute, the thread_id AND the user_id need to match, otherwise it returns nothing in the form of a 404 which isn't true at all.  The API should give back an honest answer, and frontstage should then make the choice if they want to hide the existence of an unauthorised bit of data with a 404 (which we're choosing not to do with this PR!).

The code for this will look something like:
```python
result = SecureMessage.query.join(Conversation) \
                .join(Status) \
                .filter(SecureMessage.thread_id == thread_id) 

# First check if a thread exists           
if not result.all():
    logger.error('Thread does not exist', thread_id=thread_id, user_id=user.user_uuid)
    raise NotFound(description=f"Conversation with thread_id '{thread_id}' does not exist")

result = result.filter(Status.actor == user.user_uuid) \
                .order_by(SecureMessage.id.desc())

# Then we check if the user matches.  If the filter returns nothing, then it's for someone else.
if not result.all():
    logger.error('Thread but exists, but belongs to a different respondent', thread_id=thread_id, user_id=user.user_uuid)
    raise NotAuthorised(description=f"User {user_id} not authorised to view conversation with thread_id '{thread_id}'")
```

Don't copy that verbatim as the logging is awful, but it's to show what I mean.  You'd need to define a 'notAuthorised' exception if it doesn't exist already and have it return a 403 when it's raised.

ChristopherMShelton:
Broken behavioural test is addressed in secure message PR (status code is changed)

PR #461, opened 2019-07-10
appended '- 400' to the 400 error page title for distinction from 500 in analytics

ChristopherMShelton:
# Motivation and Context
So that we can track 400 and 500 errors independently in google analytics

# What has changed
added - 400 to the 400 template title, this is what our user researcher deemed most appropriate.

# How to test?
hit the error and look at the title

# Links
https://trello.com/c/JllhbBP2/1069-400-and-500-page-both-using-the-same-title-1

### DIFF
--- a/frontstage/templates/errors/400-error.html
+++ b/frontstage/templates/errors/400-error.html
@@ -1,7 +1,7 @@
 {% import 'partials/section.html' as section %}
 {% extends "layouts/_twocol.html" %}
 
-{% block page_title %}An error has occurred - ONS Business Surveys{% endblock %}
+{% block page_title %}An error has occurred - ONS Business Surveys - 400){% endblock %}
 
 {% block main %}
         <div class="grid__col col-8@m">
--- a/frontstage/templates/errors/500-error.html
+++ b/frontstage/templates/errors/500-error.html
@@ -1,7 +1,7 @@
 {% import 'partials/section.html' as section %}
 {% extends "layouts/_twocol.html" %}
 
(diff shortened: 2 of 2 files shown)


insacuri:
Are we not doing the same for the 500 page?

ChristopherMShelton:
> Are we not doing the same for the 500 page?

yea probably makes sense to be honest, otherwise analytics will be convoluted historically from this - this way both start getting tracked fresh

ChristopherMShelton:
> Cannot approve as tests are not passing locally

this isn't a change request

frontstage/templates/errors/400-error.html
```diff
@@ -1,7 +1,7 @@
 {% import 'partials/section.html' as section %}
 {% extends "layouts/_twocol.html" %}
 
-{% block page_title %}An error has occurred - ONS Business Surveys{% endblock %}
+{% block page_title %}An error has occurred - ONS Business Surveys - 400){% endblock %}
```
necrophonic: I'd be potentially inclined to say `4xx` (and `5xx` below) as these'll pop up for a variety of those error types.

Is there anyway the actual code can be pushed into the template? Might be useful to have breakdowns of the _actual_ error codes in analytics?

ChristopherMShelton: Told not to have any technical jargon in the templates, this is just so we can see which pages are being hit and how often in the analytics. Also fairly sure once it arrives at frontstage they are in fact all 500's or 400's (except specifically 404 which has a different template)

necrophonic: Fair enough 👍

PR #462, opened 2019-07-15
Fix broken /thread links in emails

insacuri:
# Motivation and Context
The remove v2 endpoint change for secure message had a side effect that old emails stopped working correctly as /thread was changed to /threads

# What has changed
/thread/<thread_id> and /threads/<thread_id> now hit the same endpoint, whereas before hitting /thread/<thread_id> would cause an error as that endpoint doesn't exist anymore.

# How to test?
- Unit tests pass
- Ensure that thread in fronstage can be accessed using both /thread and /threads urls (just change it manually after accessing it the first time)

# Links

### DIFF
--- a/frontstage/views/secure_messaging/message_get.py
+++ b/frontstage/views/secure_messaging/message_get.py
@@ -15,9 +15,16 @@ from frontstage.views.secure_messaging import secure_message_bp
 logger = wrap_logger(logging.getLogger(__name__))
 
 
+@secure_message_bp.route('/thread/<thread_id>', methods=['GET', 'POST'])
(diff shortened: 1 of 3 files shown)


frontstage/views/secure_messaging/message_get.py
```diff
@@ -15,9 +15,16 @@
 logger = wrap_logger(logging.getLogger(__name__))
 
 
+@secure_message_bp.route('/thread/<thread_id>', methods=['GET', 'POST'])
 @secure_message_bp.route('/threads/<thread_id>', methods=['GET', 'POST'])
```
necrophonic: Can you map two annotations or should they be in the same one?

PR #463, opened 2019-07-18
Change planet font classes to design system versions

insacuri:
# Motivation and Context
The pattern library use of planets to determine text size is deprecated in favour of utility font size classes.  Updates the version of the pattern library to the most recent, updates the font classes and fixes some javascript that broke as a result.

# What has changed
- Updated the version of the pattern library used to 2.0.0.
- Updated the planet classes to utility font sizes.  
- Fixed the guidance section that broke as part of the update.

# How to test?
Look at as many pages as possible and ensure it looks the same as the current ras-frontstage.  Also ensure no functionality is broken as a result.

# Links
https://trello.com/c/C74y7xaL/1074-change-frontstage-to-use-the-utility-classes-in-the-html-instead-of-the-planets-8

### DIFF
--- a/frontstage/templates/contact-us.html
+++ b/frontstage/templates/contact-us.html
@@ -5,15 +5,15 @@
 
 {% block main %}
 
-    <h1 class="saturn">Contact us</h1>
+    <h1 class="u-fs-l">Contact us</h1>
     
     <h2 class="neptune">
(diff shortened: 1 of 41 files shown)


insacuri:
Failing as the package safety part of the test isn't working currently

PR #473, opened 2019-08-20
All done

owencjones:
# Motivation and Context
Make it possible to display a customise message on all pages to advise users of system issues that may effect them

# What has changed
- System now checks for a Redis value to set a message
- If value present, passed as global variable to all layouts
- If passed to layout, shown as a warning panel at the top of page

# How to test?
- Confirm tests run
- Review code content
- Run frontstage and confirm that you can set a key `AVAILABILITY_MESSAGE` that will show a message on pages on the site (ensure you set on the database in use)

# Links
- https://trello.com/c/LrZe68zI

### DIFF
--- a/frontstage/__init__.py
+++ b/frontstage/__init__.py
@@ -1,6 +1,5 @@
 import redis
 from flask_wtf.csrf import CSRFProtect
-
 from frontstage.create_app import create_app_object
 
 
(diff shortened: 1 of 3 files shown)


insacuri:
Can you put a more sensible title for the review?  We'll come and look at the list in the future and not know what this PR did

frontstage/__init__.py
```diff
@@ -10,6 +9,13 @@
                           db=app.config['REDIS_DB'])
 csrf = CSRFProtect(app)
 
+
+@app.context_processor
+def inject_availability_message():
+    if len(redis.keys('AVAILABILITY_MESSAGE')) == 1:
+        return dict(availability_message=redis.get('AVAILABILITY_MESSAGE').decode('utf-8'))
```
insacuri: Minor, but instead of dict() it's neater to use curly brackets

owencjones: Done.

PR #488, opened 2019-11-19
Reorganise code to check auth before party

insacuri:
# Motivation and Context
Currently, we're open to being DDoSed as our sign in checks party first before the auth service to see if the user exists.  If someone were to repeatedly hit us, they could take down party with too much traffic for party to handle.
 The auth service is only really used for account management whereas party gets used for much more then that

# What has changed
- Reorganised code to check the auth service for account existance/errors before checking party to get the id for the session token.
- Updated a few error messages to make some very unlikely events easier to diagnose if they happened (where the user exists in auth but somehow not in party).

# How to test?
- Unit tests pass
- Login should work in the same way.  Notice that if you repeatedly hit frontstage with a bad login, party doesn't get any traffic.

# Links
https://trello.com/c/0FOgUr66


frontstage/views/sign_in/sign_in.py
```diff
@@ -111,3 +118,13 @@ def resend_verification_expired_token(token):
     party_controller.resend_verification_email_expired_token(token)
     logger.info('Re-sent verification email for expired token.', token=token)
     return render_template('sign-in/sign-in.verification-email-sent.html')
+
+
+def obfuscate_email(email):
```
jterryons: Is there a unit test for this?

insacuri: Good spot.  There isn't and is a perfect candidate for one!

PR #495, opened 2019-11-27
Filter ended collection exercises out of lists

insacuri:
# Motivation and Context
Currently we get the live collection exercises for the user for the todo and history lists.  These never go away so the list gets longer and longer.  This PR filters the surveys from the live surveys list if the end date (which is a date far in the future from the return by date) to make the list be full of relevant surveys.

Note:  The collection exercise service has the concept of not-started and live surveys only.  In the future, if it ever becomes a list of 3 (not-started, live-but-not-ended and ended) then the filter done here can be removed and we can change the call to collection exercise with the data we care about.

# What has changed
- Added code to filter out the ended surveys from the list.

# How to test?
- Unit and acceptance tests pass.
- Spin the service up (with seeded data from accecptance tests), change the >= to <= in the filter function to see the lists change.  You can also comment the check out to see everything instead of the filtered lists.

# Links
https://trello.com/c/vfbt44rn
https://trello.com/c/OWir1kSL

### DIFF
--- a/tests/app/controllers/test_party_controller.py
+++ b/tests/app/controllers/test_party_controller.py
@@ -1,3 +1,5 @@
+
+import json
 import unittest
 from collections import namedtuple
 from unittest.mock import patch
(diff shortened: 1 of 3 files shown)


tests/app/controllers/test_party_controller.py
```diff
@@ -275,3 +278,25 @@ def test_display_button(self):
         ]
         for combination in combinations:
             self.assertEqual(display_button(combination.status, combination.ci_type), combination.expected)
+
+    def test_filter_ended_collection_exercises(self):
+        """Tests the functionality of the 'filter_ended_collection_exercises' function"""
+        with open('tests/test_data/party/collection_exercises.json') as business_json_data:
+            data = json.load(business_json_data)
+
+        # Data has 2 collection exercises in the far future, nothing should be removed
```
jterryons: perhaps a bit nitpicky but couldnt we mock datetime.now to return a static date? wont have to worry about the test failing in the fututre then

PR #496, opened 2019-11-28
Added a temp script to disable button after submit for 1 second

boxadesign:
# Motivation and Context
Issue with forms being submitted multiple times due to multiple button clicks.
# What has changed
Added a script that disables a button for 1 second.
NOTE - this is a temp fix until we refactor to use design system components.

# How to test?
Check all forms submit ok. Test for double submissions.

### DIFF
--- a/frontstage/static/js/script.js
+++ b/frontstage/static/js/script.js
@@ -2,12 +2,6 @@
 
 function _instanceof(left, right) { if (right != null && typeof Symbol !== "undefined" && right[Symbol.hasInstance]) { return !!right[Symbol.hasInstance](left); } else { return left instanceof right; } }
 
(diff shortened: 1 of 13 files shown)


boxadesign:
> Does this only work on html buttons? There is a confirm 'button' (it's actually a link...) that is known to break things if it's double clicked

Do you know which link it is?

insacuri:
https://github.com/ONSdigital/ras-frontstage/blob/master/frontstage/templates/surveys/surveys-confirm-organisation.html#L14 that's the one I know about, but I'd be surprised if that was the only one

insacuri:
The double click protection on buttons is good to go.  I'll hold off on approval until the protection on links is in.

PR #503, opened 2020-01-07
Remove frontstage emails through party

lauraedwards1:
# Motivation and Context

All the email communications on Frontstage needed for things like reset password etc went through Party, including for things as simple as verification link generation/checking.

# What has changed

# How to test?

# Links


config.py
```diff
@@ -13,6 +13,9 @@ class Config(object):
     PORT = os.getenv('PORT', 8082)
     MAX_UPLOAD_LENGTH = os.getenv('MAX_UPLOAD_LENGTH', 20 * 1024 * 1024)
 
+    EMAIL_TOKEN_SALT = os.getenv('EMAIL_TOKEN_SALT', 'aardvark')
+    EMAIL_TOKEN_EXPIRY = int(os.getenv('EMAIL_TOKEN_EXPIRY', '306000'))
```
mfcrocker: This is 85 hours - why?

frontstage/controllers/notify_controller.py
```diff
@@ -0,0 +1,69 @@
...
+        self.notify_account_locked = config['RAS_NOTIFY_ACCOUNT_LOCKED_TEMPLATE']
+
+    def _send_message(self, email, template_id, personalisation=None, reference=None):
+        """
+        Send message to gov.uk notify wrapper
+        :param email: email address of recipient
+        :param template_id: the template id on gov.uk notify to use
+        :param personalisation: placeholder values in the template
+        :param reference: reference to be generated if not using Notify's id
+        :rtype: 201 if success
```
insacuri: I know you didn't write this but it should be `:returns: 201 if success`

frontstage/controllers/validators.py
```diff
@@ -0,0 +1,56 @@
+import itertools
```
insacuri: This whole thing isn't used.  It can be deleted (and thank god, it's an over engineered way to do validation)

requests_wrapper.py
```diff
@@ -0,0 +1,43 @@
+import requests
```
insacuri: I'm almost certain this isn't useful.  We have a whole extra class because someone didn't want to have to deal with timeouts and setting auth? 
I know you didn't write it but can you change the notify gateway to use the timeout and auth args and remove this class?

frontstage/controllers/notify_controller.py
```diff
@@ -0,0 +1,69 @@
...
+            response = Requests.post(url, json=notification)
+
+            logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
+
+        except Exception as e:
+            ref = reference if reference else 'reference_unknown'
+            raise exceptions.RasNotifyError("There was a problem sending a notification to Notify-Gateway "
+                                            "to GOV.UK Notify", error=e, reference=ref)
+
+    def request_to_notify(self, email, template_name, personalisation=None, reference=None):
```
insacuri: I know you copied this, but I think this could be written better.

In _get_template_id, the templates should be moved to the __init__ function as the list of templates is a perfect candidate to be an attribute for this object.

Next the _get_template_id function can be removed.  The line that calls it can be replaced with
```python
try:
    template_id = self.templates[template_name]
except KeyError:
    raise KeyError('Template does not exist')
```

frontstage/controllers/notify_controller.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def _send_message(self, email, template_id, personalisation=None, reference=None):
+        """
+        Send message to gov.uk notify wrapper
+        :param email: email address of recipient
+        :param template_id: the template id on gov.uk notify to use
+        :param personalisation: placeholder values in the template
+        :param reference: reference to be generated if not using Notify's id
+        :rtype: 201 if success
+        """
```
insacuri: This should also have a `:raises KeyError:  Raised when the template name provided doesn't exist`

frontstage/controllers/notify_controller.py
```diff
@@ -0,0 +1,69 @@
...
+
+        if not self.config['SEND_EMAIL_TO_GOV_NOTIFY']:
```
insacuri: Can you do 3 changes to this.
- Add `self.send_email_to_notify = self.config['SEND_EMAIL_TO_GOV_NOTIFY']` to __init__
- Remove `self.config = config` from __init__
- Replace this line with `if not self.send_email_to_notify:`

This object gets the config passed to it, then it's saved as an attribute for seemingly no reason.  This'll just make it tidier and discourages people from being lazy with this object by doing `self.config['thing']` if they want something

frontstage/controllers/notify_controller.py
insacuri: Also add `:raises RasNotifyError: Raised when there is an error sending a message to the gov notify service`

tests/test_data/mocks.py
```diff
@@ -0,0 +1,89 @@
+import json
```
insacuri: Is this file used by anything?  I can't see what uses it

config.py
```diff
@@ -142,10 +155,16 @@ class TestingConfig(DevelopmentConfig):
     TESTING = True
     DEVELOPMENT = False
     WTF_CSRF_ENABLED = False
+    SEND_EMAIL_TO_GOV_NOTIFY = True
```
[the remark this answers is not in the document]
lauraedwards1: tests need it

PR #507, opened 2020-01-21
Append the protocol and host to reset link

insacuri:
# Motivation and Context
Before, the password reset link would be `/passwords/reset-password/<token>`
This PR fixes so that it becomes `http://localhost:8082/passwords/reset-password/<token>`

# What has changed

# How to test?

# Links

### DIFF
--- a/frontstage/views/passwords/reset_password.py
+++ b/frontstage/views/passwords/reset_password.py
@@ -96,7 +96,11 @@ def request_password_change(email):
     logger.info("Requesting password change", party_id=party_id)
 
     token = verification.generate_email_token(email)
-    verification_url = url_for('passwords_bp.post_reset_password', token=token)
+
+    # url_root comes with a trailing slash that we don't want.
+    url_root = request.url_root[:-1]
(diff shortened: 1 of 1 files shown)


frontstage/views/passwords/reset_password.py
```diff
@@ -96,7 +96,11 @@ def request_password_change(email):
     logger.info("Requesting password change", party_id=party_id)
 
     token = verification.generate_email_token(email)
-    verification_url = url_for('passwords_bp.post_reset_password', token=token)
+
+    # url_root comes with a trailing slash that we don't want.
+    url_root = request.url_root[:-1]
```
jterryons: is it a guarantee that it will always have a trailing slash? Is there any scenario where we might not get a trailing slash?

insacuri: Good question, there isn't a gaurantee (and the documentation in flask is actually wrong presently on this).  I'll put a check to not remove it if the last char isn't a slash

insacuri: Fixed

PR #552, opened 2020-04-07
Fix for the broken CSRF

warrenbailey:
# Motivation and Context

Reordering the CSRF setup and adding a configurable time out for the token

# What has changed

# How to test?

# Links

### DIFF
--- a/_infra/helm/frontstage/values.yaml
+++ b/_infra/helm/frontstage/values.yaml
@@ -43,7 +43,7 @@ email:
 
 scheme: http
 csrf:
-  protected: "True"
+  protected: True
   timeLimit: 7200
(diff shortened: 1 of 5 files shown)


_infra/helm/frontstage/values.yaml
```diff
@@ -32,7 +32,7 @@ ingress:
   certNameSurveys: surveys-cert
 
 analytics:
-  enabled: false
+  enabled: False
```
jterryons, line 35: not sure if this needs changing, i think i did this in error

warrenbailey: I've done that by mistake, good spot

PR #560, opened 2020-04-14
Local test fix

jterryons:
CSRF and Secure cookies break chrome tests, firefox doesnt have the same issues. This is due to chrome following their own standards and not RFC. By controlling secure cookies and csrf by the WTF_CSRF_ENABLED flag we can test the basic functionality of the frontend with acceptance tests


frontstage/views/sign_in/sign_in.py
```diff
@@ -34,7 +35,9 @@ def login():
     form.username.data = form.username.data.strip()
     account_activated = request.args.get('account_activated', None)
 
-    if request.method == 'POST' and form.validate():
+    secure = Config.WTF_CSRF_ENABLED
```
warrenbailey, line 38: This should be loaded from the app config.

secure = app.config['WTF_CSRF_ENABLED']

PR #562, opened 2020-04-16
force https and allow http traffic at the loadbalancer

jterryons:
http traffic will always give us the google 404 error, which will confuse external users.

By forcing https in talisman we can ensure that we are always redirected to https. Our HSTS headers will ensure all all subsequent redirects happen at the browser level preventing any man in the middle attacks.

The httpsOnly annotation needs to be removed in order to allow http traffic at the loadbalancer

### DIFF
--- a/config.py
+++ b/config.py
@@ -157,7 +157,6 @@ class DevelopmentConfig(Config):
     SURVEY_AUTH = (SURVEY_USERNAME, SURVEY_PASSWORD)
     WTF_CSRF_TIME_LIMIT = int(os.getenv('WTF_CSRF_TIME_LIMIT', '3200'))
     SECURE_APP = bool(strtobool(os.getenv('SECURE_APP', "False")))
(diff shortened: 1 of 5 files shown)


config.py
```diff
@@ -97,7 +97,8 @@ class Config(object):
     RAS_NOTIFY_ACCOUNT_LOCKED_TEMPLATE = os.getenv('RAS_NOTIFY_ACCOUNT_LOCKED_TEMPLATE', 'account_locked_id')
     SEND_EMAIL_TO_GOV_NOTIFY = os.getenv('SEND_EMAIL_TO_GOV_NOTIFY', False)
     REQUESTS_POST_TIMEOUT = os.getenv('REQUESTS_POST_TIMEOUT', 20)
-    WTF_CSRF_ENABLED = bool(strtobool(os.getenv('WTF_CSRF_ENABLED', "True")))
+    SECURE_APP = bool(strtobool(os.getenv('SECURE_APP', "True")))
```
warrenbailey, line 100: We probably want SECURE_APP set to False in the Development and Testing Config objects

PR #566, opened 2020-04-22
updated the logging in the notify controller

bkhan2:
# Motivation and Context
updating logging when notify notification doesn't get sent to be clearer

# What has changed
added URL and Status Code logging.

### DIFF
--- a/frontstage/controllers/notify_controller.py
+++ b/frontstage/controllers/notify_controller.py
@@ -5,9 +5,7 @@ import structlog
 import requests
 
 from frontstage.exceptions import exceptions
-from fronstage.exceptions.exceptions import RasNotifyError
 from flask import current_app as app
-from requests.exceptions import HTTPError
 
 logger = structlog.wrap_logger(logging.getLogger(__name__))
 
@@ -59,8 +57,8 @@ class NotifyGateway:
         except Exception as e:
             ref = reference if reference else 'reference_unknown'
             resp = response if response else {
-                'status_code' : 'Unknown',
-                'text' : 'Exception thrown before request was made'
+                'status_code': 'Unknown',
+                'text': 'Exception thrown before request was made'
             }
             raise exceptions.RasNotifyError("There was a problem sending a notification "
                                             "via Notify-Gateway to GOV.UK Notify.",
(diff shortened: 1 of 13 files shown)


frontstage/controllers/notify_controller.py
```diff
@@ -41,26 +41,27 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+        auth = app.config['SECURITY_USER_NAME'], app.config['SECURITY_USER_PASSWORD']
+        response = requests.post(url, json=notification, auth=auth,
+                                 timeout=int(app.config['REQUESTS_POST_TIMEOUT']))
+        status_code = response.status_code
 
+        if status_code == 201:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
-
-        except Exception as e:
+        else:
```
jterryons, line 60: Silimar comment to the other pr. we shouldbe be raising an error on codes 200 -399.

mfcrocker: As other PR, disagree with this

frontstage/controllers/notify_controller.py
```diff
@@ -41,26 +41,27 @@ def _send_message(self, email, template_id, personalisation=None, reference=None
...
+                                 timeout=int(app.config['REQUESTS_POST_TIMEOUT']))
+        status_code = response.status_code
 
+        if 201 <= status_code <= 399:
             logger.info('Notification id sent via Notify-Gateway to GOV.UK Notify.', id=response.json()["id"])
-
-        except Exception as e:
+        else:
             ref = reference if reference else 'reference_unknown'
             raise exceptions.RasNotifyError("There was a problem sending a notification to Notify-Gateway "
```
mfcrocker, line 62: You don't need to pass these variables in as fstrings, `RasNotifyError` takes kwargs
```suggestion
            raise RasNotifyError("There was a problem sending a notification to Notify-Gateway.", url=url, status_code=status_code, message=response.text, reference=ref)
```

jterryons:
the code can only be tested in a suite of integration tests. None of the code is isolated for testing making it very difficult to trust any of the changes for so little gain. its not worth the effort right now, we have bigger issues to deal with.

PR #582, opened 2020-06-01
add session logic for unread messages

jterryons:
Add an unread message count to the current session.
Controllers that render a new page will get the unread message count from the sessio with the exception of the signin page and the view conversation route. These calls must hit the api directly.

When the secure-message api returns a count, the session will be updated with the new information and will be timed for 10 minutes, after 10 minutes the count will refresh with another query to the secure message api.

Changed SessionHandler to Session. The session object will has several factory methods and convenience functions. Session logic has been taken out of the controllers. The session object can be persisted by calling `session.save()`. Renamed the `update_session` function to `refresh` as it was only renewing the expiry time of the jwt, refesh is more readable.

Changed the jwt_authorisation decorator to return a session object rather than a jwt dict (which was just the session object anyway). The Oauth token we retrieve on signin was never persisted to the session, instead our jwt.timestamp function was discarding it so ive removed the logic that puts the oauth token into a temporary dict as it serves no purpose if not persisted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,8 +19,8 @@ docker-test: unit-tests integration-tests
 check:
 	pipenv check
 
-unit-tests: check lint
+unit-tests: check
 	APP_SETTINGS=TestingConfig pipenv run pytest $(UNIT_TESTS) --cov frontstage --cov-report term-missing	
 
-integration-tests: check lint
+integration-tests: check
 	APP_SETTINGS=TestingConfig pipenv run pytest $(INTEGRATION_TESTS) --cov frontstage --cov-report term-missing	
--- a/Makefile
+++ b/Makefile
@@ -7,9 +7,6 @@ build-docker:
 build-kubernetes:
 	docker build -f _infra/docker/Dockerfile .
 
-lint:
-	pipenv run flake8 ./frontstage ./tests
-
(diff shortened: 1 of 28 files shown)


frontstage/views/sign_in/sign_in.py
```diff
@@ -42,7 +42,7 @@ def login():  # noqa: C901
         bound_logger = logger.bind(email=obfuscate_email(username))
         bound_logger.info("Attempting to find user in auth service")
         try:
-            oauth2_token = oauth_controller.sign_in(username, password)
+            oauth_controller.sign_in(username, password)
```
warrenbailey, line 45: I think we should remove references to oauth, its confusing as the auth server doesn't use oauth at all

Makefile
```diff
@@ -7,9 +7,6 @@ build-docker:
 build-kubernetes:
 	docker build -f _infra/docker/Dockerfile .
 
-lint:
```
warrenbailey, line 10: as above?

warrenbailey:
Tested and confirmed working

warrenbailey:
Raising another ticket to remove references to oauth2 and to tidy up the sign in code.

PR #589, opened 2020-06-05
Enable overseas telephone numbers in Frontstage account creation

jdfox:
# Motivation and Context
Updated UI for use of international numbers

# What has changed
Changed web forms slightly; existing validation works for international phone numbers. 

# How to test?
**local:**  
- `make unit-tests`
- `make integration-tests`

**dev**:
- in a dev cluster, spin up the frontstage pod with the image `overseas-phone-nums`
- run acceptance tests
- create an account with an unused enrolment code from response-ops-ui
- check that on the page "Enter your account details", the form's phone number section matches the updated section in the Powerpoint
- Sign up with an international number, e.g. +44 7777 123456

# Links
https://trello.com/c/331UX3Uw


frontstage/templates/register/register.enter-your-details.html
```diff
@@ -126,6 +126,7 @@ <h2 class="section__title neptune u-mt-m">Phone number</h2>
                     {{ form.phone_number.label(class_='label') }}
                     {{ form.phone_number(class_='input input--text input-type__input') }}
                 </div>
+                <span style="font-size: 80%">For international numbers include the country code</span>
```
mfcrocker, line 129: Is there a CSS class we could use for this? I don't like us putting in in-line styles.

jdfox: Agreed, thanks. I've added a class ".field-description", which is bound to be useful for other things.

PR #599, opened 2020-06-15
adding functionality to reject 0 byte files

bkhan2:
# Motivation and Context
to reject 0 byte files

# What has changed
added functions to check upload file size
added a test to test the upload of 0 byte files

# How to test?
spin up frontstage
get a survey through rops
upload a 0 byte file
- it should reject the file and there should be an option to browse for a new file


frontstage/controllers/collection_instrument_controller.py
```diff
@@ -86,3 +87,12 @@ def upload_collection_instrument(upload_file, case_id, party_id):
         raise ApiError(logger, response)
 
     logger.info('Successfully uploaded collection instrument', case_id=case_id, party_id=party_id)
+
+
+def check_collection_instrument_size(upload_file, party_id, case_id):
+    logger.info('Checking the file size', party_id=party_id, case_id=case_id)
+    upload_file.save('/tmp/file')
```
insacuri, line 94: We shouldn't be saving anything to the file system.  You'll need to figure out another way to check for the size.

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -86,3 +86,14 @@ def upload_collection_instrument(upload_file, case_id, party_id):
         raise ApiError(logger, response)
 
     logger.info('Successfully uploaded collection instrument', case_id=case_id, party_id=party_id)
+
+
+def check_collection_instrument_size(upload_file, party_id, case_id):
```
insacuri, line 91: This function name should probably be `is_collection_exercise_too_small`, because that's what it's doing.  It also makes it instantly understandable by just looking at the function name.

Also, why are we passing in the party_id and case_id if all we're doing is logging what they are.  Agreed we should be adding the context to the logs where possible, but I don't think passing the values to the function exclusively for logging is correct

Also add a docstring!

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -86,3 +86,14 @@ def upload_collection_instrument(upload_file, case_id, party_id):
...
+    logger.info('Checking the file size', party_id=party_id, case_id=case_id)
+    upload_file_readable = upload_file.read()
+    upload_file_size = len(upload_file_readable)
+    print('file size is: ', upload_file_size)
```
insacuri, line 95: Remove this.  I assume it was for debugging?

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -86,3 +86,14 @@ def upload_collection_instrument(upload_file, case_id, party_id):
...
+
+    if upload_file_size > 0:
+        logger.info('File size is correct', party_id=party_id, case_id=case_id)
+        return True
```
insacuri, line 99: So what happens if it is too big?  You don't `return False` and you don't raise an exception.  Currently if the file is 0 bytes then it implicitly does a `return None`.
Having a function with 2 return types (in this case bool or None) is bad and will bite you down the line.
I think having this return False is good enough (especially if you do the function rename I mention earlier)

frontstage/views/surveys/upload_survey.py
```diff
@@ -34,37 +34,50 @@ def upload_survey(session):
...
-        elif "File too large" in ex.error_message:
-            error_info = 'size'
-        else:
-            logger.error('Unexpected error message returned from collection instrument service',
-                         status=ex.status_code,
-                         error_message=ex.error_message,
-                         party_id=party_id,
-                         case_id=case_id)
-            error_info = "unexpected"
+    # Checking if the file size is great than 5999 bytes
```
insacuri, line 48: The comment and error message check that it's 6kb+ but the function just checks for it being more than 0.
Also, how did we arrive to 6kb?  It sounds about right and I assume you just made an empty excel file and looked at the size of it?

frontstage/controllers/collection_instrument_controller.py
```diff
@@ -86,3 +86,14 @@ def upload_collection_instrument(upload_file, case_id, party_id):
         raise ApiError(logger, response)
 
     logger.info('Successfully uploaded collection instrument', case_id=case_id, party_id=party_id)
+
+
+def check_collection_instrument_size(upload_file, party_id, case_id):
+    logger.info('Checking the file size', party_id=party_id, case_id=case_id)
+    upload_file_readable = upload_file.read()
+    upload_file_size = len(upload_file_readable)
```
insacuri, line 94:
```suggestion
    upload_file_size = len(upload_file.read())
```

You only use the variable once, so might as well flatten it

frontstage/views/surveys/upload_survey.py
```diff
@@ -34,37 +34,50 @@ def upload_survey(session):
     party_controller.is_respondent_enrolled(party_id, business_party_id, survey_short_name)
 
     # Get the uploaded file
-    upload_file = request.files['file']
-    upload_filename = upload_file.filename
+
+    # The variable name has '_raw' so that it can be called in the file size checking function,
```
insacuri, line 38: Comment probably isn't worth it.  2-3 lines later is a variable called `upload_file` so it's pretty clear why it's called this without the comment

frontstage/views/surveys/upload_survey.py
```diff
@@ -34,37 +34,50 @@ def upload_survey(session):
...
-            error_info = 'size'
-        else:
-            logger.error('Unexpected error message returned from collection instrument service',
-                         status=ex.status_code,
-                         error_message=ex.error_message,
-                         party_id=party_id,
-                         case_id=case_id)
-            error_info = "unexpected"
+    # Checking if the file size is great than 5999 bytes
+    if collection_instrument_controller.check_collection_instrument_size(upload_file_raw, party_id, case_id):
```
insacuri, line 49: The logic in this bit is really weird now so I think it should be rewritten (though maybe as part of another PR after this one so too much isn't going on at once)
The validation for the file is really weird.  So it looks like previously ALL the validation was done in the collection-instrument service.  We hand it whatever crap the user gives to us, we pass it to the service and it tells us if something is wrong.  Now we do a bit of validation on our side, they do the rest of the validation on their side. 

Now that I write this.  I think the collection instrument service should also get the same treatment?  If something else wanted to post a collection instrument, there'd be no safeguards to make sure that it's not a 0 byte file, even though we know that 0 byte files are wrong

insacuri: Though after talking to JT, what's been done here is fine as is.  If we want to improve it later on we'll just do it down the line.  Just need to fix the other comments I have and it should be good to go.
I tested it in my sandbox environment (mostly to play with the github actions) and it works well.

tests/integration/views/surveys/test_upload_survey.py
```diff
@@ -110,3 +122,9 @@ def test_upload_survey_content_too_long(self):
 
         self.assertEqual(response.status_code, 302)
         self.assertTrue('/surveys/upload-failed'.encode() in response.data)
+
+    def test_is_collection_instrument_too_small(self):
```
insacuri, line 126: Reading this test, I realise the logic is the wrong way around.  I would expect a functionc alled `is_collection_instrument_too_small` to return True if it IS too small, and False if it IS NOT too small.  

The way it's done currently is the opposite of that.

PR #601, opened 2020-06-23
Remove the oauth2 code 

jdfox:
# Motivation and Context
Clean up old oauth2 stuff, and non-oauth2 stuff that's labelled oauth2

# What has changed
Eight files changed, with auth controller and other functions renamed, and  /api/v1/tokens endpoint trimmed

# How to test?
**Local**:
- `make unit-tests`
- `make integration-tests`

**Sandbox**:
- spin up cluster with frontstage image `rm-oauth2`, and test

# Links
https://trello.com/c/q33sUuxc


tests/integration/test_sign_in.py
```diff
@@ -31,7 +31,7 @@ class TestSignIn(unittest.TestCase):
 
     def setUp(self):
         self.app = app.test_client()
-        self.oauth_token = {
+        self.auth_token = {
             "id": 1,
             "access_token": "8c77e013-d8dc-472c-b4d3-d4fbe21f80e7",
```
insacuri, line 36: This bit needs to be cleaned up to better reflect what you actually get these days.  The old django-oauth2-test would give you an access token (and I think expiry?).  These days the 'token' is just an empty dict I think (but double check before you start deleting stuff!)

frontstage/controllers/auth_controller.py
```diff
@@ -28,21 +28,17 @@ def sign_in(username, password):
 
     try:
         response.raise_for_status()
-        # This if statement will only be true if the 'oauth' service is the ras-rm-auth-service.  This is only
-        # meant to be a temporary measure until the switch is made and the new auth service is used.  Once
-        # the oauth2 service is switched off, we should look to tidy up the api of the auth service (As the service
-        # doesn't need a /tokens endpoint if there are no tokens) and tidy up this services interaction with it.
         if response.status_code == 204:
```
insacuri, line 31: This bit of code needs to change.  Looking specifically for a 204 was there to handle the changeover.  
After the 'successfully signed in' log line, you need to delete `return response.json()` and replace it with`return {}`
You also need to delete this if block.  so the code should look  something like 
```python
try:
    response.raise_for_status()
    <error code>

bound_logger.info('Successfully signed in')
return {}
```

frontstage/controllers/auth_controller.py
```diff
@@ -28,21 +28,17 @@ def sign_in(username, password):
...
             return {}
     except requests.exceptions.HTTPError:
         if response.status_code == 401:
```
insacuri, line 34: Find out why we have 2 separate blocks for http errors.  I can't remember why we'd want handle a 401 differently from anything else as we generally don't tell people if they've just typed in the wrong password vs not having an account

frontstage/logger_config.py
```diff
@@ -20,9 +20,9 @@ def add_service(logger, method_name, event_dict):  # pylint: disable=unused-argu
         return event_dict
 
     logging.basicConfig(stream=sys.stdout, level=log_level, format=logger_format)
-    oauth_log = logging.getLogger("requests_oauthlib")
-    oauth_log.addHandler(logging.NullHandler())
-    oauth_log.propagate = False
+    auth_log = logging.getLogger("requests_authlib")
```
insacuri, line 23: This whole bit might be able to go.  requests_oauthlib is the name of a python library that was removed recently when we had a bit of a tidyup of the libraries.  I think this block was handling the logging for that library which isn't here anymore.

jdfox: fixed

frontstage/controllers/auth_controller.py
jdfox: Good idea, thanks. Done.

frontstage/controllers/auth_controller.py
jdfox: Done.

frontstage/controllers/auth_controller.py
```diff
@@ -0,0 +1,38 @@
...
+    headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+    }
+    response = requests.post(url, headers=headers, auth=app.config['BASIC_AUTH'], data=data)
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        auth_error = response.json().get('detail', '')
```
insacuri, line 32: Looking at the ras-rm-auth-service, I realise I was wrong about this bit.  We need to handle the 401 error (to log a warning for a failed login attempt), but we also need the non-401 error incase something more serious is happening when we try to connect to the auth service. 
How we have it here, if the service was returning 500s, we'd still just be logging warnings when something was actually on fire.

tests/integration/test_sign_in.py
jdfox: Auth hands back a response object here, as a dict, and we need to avoid confusion with JWTs. Removed the expired OAUTH token from the tests ;  renamed remaining ".auth_token" to ".auth_response"; set the initial ".auth_response" to an empty dict.

PR #605, opened 2020-07-01
display 'file size too small' error page

bkhan2:
# Motivation and Context
adding ability for frontstage to let user know the file is too small and to reupload a non-zero byte file.

# What has changed
added code into upload_survey views


frontstage/views/surveys/upload_survey_failed.py
```diff
@@ -33,6 +33,9 @@ def upload_failed(session):
     elif error_info == "size":
         error_info = {'header': "Error uploading - file size too large",
                       'body': 'The spreadsheet must be smaller than 20MB in size'}
+    elif error_info == "sizeSmall":
+        error_info = {'header': "Error uploading - file size too small",
+                      'body': "The spreadsheet must be larger than 6KB in size"}
```
mfcrocker, line 38: Is there any particular reason we've gone for 6KB here?

bkhan2: the template that is given is 6kb, so i used that. however it is worth noting that the actual limit is 1 byte, just so frontstage rejects 0 byte files. Anything above zero can go through.

warrenbailey: I think it should be 1kb, if thats what the actual limit is

PR #633, opened 2020-11-24
Fix CSP error for GTagMgr in base

jdfox:
# Motivation and Context
CSP script error in Tag Manager code, found during investigations with DST, complains about missing nonce (or alternatives) 

# What has changed
Added missing nonce to script tag in base html template

# How to test?
Deploy to dev, run acceptance tests, check for CSP error

# Links
- https://trello.com/c/etrcdyrU

### DIFF
--- a/frontstage/views/sign_in/sign_in.py
+++ b/frontstage/views/sign_in/sign_in.py
@@ -92,8 +92,7 @@ def login():  # noqa: C901
                             value=session.session_key,
                             expires=session.get_expires_in(),
                             secure=True,
-                            httponly=True,
-                            SameSite=None; Secure)
+                            httponly=True)
         count = conversation_controller.get_message_count_from_api(session)
         session.set_unread_message_total(count)
         bound_logger.info('Successfully created session', session_key=session.session_key)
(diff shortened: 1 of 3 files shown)


frontstage/views/sign_in/sign_in.py
```diff
@@ -91,8 +91,8 @@ def login():  # noqa: C901
         response.set_cookie('authorization',
                             value=session.session_key,
                             expires=session.get_expires_in(),
-                            secure=secure,
-                            httponly=secure)
+                            secure=True,
```
warrenbailey, line 94: I think this is configurable for a reason??

The secure variable is set by the WTF_CSRF_ENABLED environment variable and it won't make a difference in preprod/prod as it's enabled, so I'm not sure why you're doing this?

warrenbailey (changes requested):

frontstage/views/sign_in/sign_in.py
jdfox: You're right: reverting.

warrenbailey (approved):

PR #660, opened 2021-01-21
Uac business details

jdfox:
# Motivation and Context
Adding survey name and business name to survey message threads.

# What has changed
- added survey and business vars to conversation view template
- defined survey var from query to survey controller
- defined business var from existing conversation data
- updated existing test of thread, with mocking for survey name query
- also combined Makefile's unit and integration test targets into optional combined `make test` target

# How to test?
`make test`, deploy frontstage branch image to dev cluster and smoke test

# Links
- https://trello.com/c/oO3qMudG

# Screenshots (if appropriate):

PR #661, opened 2021-01-22
Improved testing and error handling

jdfox:
# Motivation and Context
Improvements to error handling and testing for messages in Frontstage, followup to PR 660

# What has changed
- renamed message test data file for clarity, along with referencing vars
- Wrapped new survey name lookup in try-except, as suggested by Adam M.
- added pytest config to reduce GC descriptor deprecation warning clutter

# How to test?
- `make test`, smoke test in dev cluster

# Links
- https://github.com/ONSdigital/ras-frontstage/pull/660
- https://trello.com/c/oO3qMudG


frontstage/views/secure_messaging/message_get.py
```diff
@@ -48,8 +49,13 @@ def view_conversation(session, thread_id):
             return redirect(url_for('secure_message_bp.view_conversation_list'))
 
     unread_message_count = {'unread_message_count': get_message_count_from_api(session)}
-    survey_name = get_survey(app.config['SURVEY_URL'], app.config['BASIC_AUTH'],
-                             refined_conversation[-1]['survey_id']).get('longName')
+    survey_name = None
```
warrenbailey, line 52: would be good to have a unit test which checks this is rendered as "No survey" in this scenario.

tests/integration/test_secure_message.py
```diff
@@ -5,7 +5,7 @@
 
 from frontstage import app
 from frontstage.exceptions.exceptions import IncorrectAccountAccessError
-from tests.integration.mocked_services import (conversation_json, conversation_list_json,
```
warrenbailey, line 8: Why have we renamed conversation to message? I thought message was singular and conversation was multiple messages?

pytest.ini
```diff
@@ -0,0 +1,10 @@
+[pytest]
+filterwarnings =
+    ignore:Call to deprecated create function Descriptor
```
warrenbailey, line 3: I can see these are coming from the google api, i wonder if we can exclude that from the check rather than adding these here

warrenbailey (changes requested):

frontstage/views/secure_messaging/message_get.py
```diff
@@ -48,8 +49,13 @@ def view_conversation(session, thread_id):
...
+    try:
+        survey_name = get_survey(app.config['SURVEY_URL'], app.config['BASIC_AUTH'],
+                                 refined_conversation[-1]['survey_id']).get('longName')
+    except ApiError as exc:
+        logger.error('Failed to retrieve survey name', status_code=exc.status_code)
```
insacuri, line 57: The logging should be info level as the error logging has happened already in the controller.
Additionally, the message here isn't all that useful.  Maybe something like `Failed to get survey name, setting to None`?

frontstage/views/secure_messaging/message_get.py
```diff
@@ -48,8 +49,13 @@ def view_conversation(session, thread_id):
...
+
     business_name = conversation['messages'][-1]['@business_details']['name']
```
insacuri, line 59: I can't remember, but we're certain you'll always find a `@business_details -> name` in every message?  I think it throws a keyError if not which gives an ugly 500error screen so if there are any doubts then it needs to be handled as well

insacuri (changes requested):
One small change, but I'm happy with the rest

tests/integration/test_secure_message.py
insacuri: It's because the conversation_json was a misnomer.  The contents of `conversation_json` has just a single message in a conversation.  A lot of the things that use it tend to wrap it in a list (as a conversation is a list of messages) so it's clearer what is it with this name change

warrenbailey: ah ok that's fair though then

pytest.ini
jdfox: It's a temporary problem arising from protobuf not having a Python 3.9 wheel yet. Removing pytest.ini: anyone testing frontstage can put these lines into their setup.cfg if they wish, but it's only a minor annoyance and will likely be fixed soon.

frontstage/views/secure_messaging/message_get.py
jdfox: Done, thanks.

frontstage/views/secure_messaging/message_get.py
jdfox: Better to try-except then, especially if there's any chance of the error in future messages. Done, thanks.

frontstage/views/secure_messaging/message_get.py
jdfox: Done, thanks.

tests/unit/test_message_get.py
```diff
@@ -24,3 +38,11 @@ def test_get_msg_to_returns_newest_internal_user_if_known(self):
...
 
         self.assertEqual(to, ['user2'])
+
+    @requests_mock.mock()
+    def test_get_msg_without_survey_name(self, mock_request):
+        mock_request.get(url_get_thread, json={'messages': [message_json], 'is_closed': False})
+        mock_request.get(url_get_conversation_count, json={'total': 0})
+        mock_request.get(url_get_survey_long_name, json={"longName": "None",})
+        response = self.app.get("secure-message/threads/9e3465c0-9172-4974-a7d1-3a01592d1594")
+        self.assertTrue('None'.encode() in response.data)
```
insacuri, line 48: This is a bit of a weak test, 'None' could easily appear anywhere in the HTML (in the business name for example).  Can you amend this to make it more likely that it's definitely looking at the survey name part?  It's okay to put a bigger block of HTML in if you need to

warrenbailey: Also I thought the template would render "No Business" or "No Survey" rather than None? Should we be checking for that?

frontstage/views/secure_messaging/message_get.py
jdfox: Done, thanks

tests/unit/test_message_get.py
jdfox: Quite right: fixed.

warrenbailey (approved):

PR #670, opened 2021-02-15
Changing account to my account and adding validation for email

akssinhaamit:
Business has asked for Account to be changed to My Account against `https://trello.com/c/rKk4XdJA/364-s27-user-account-creation-change-own-email-address` also a email validation for existing email in use has been added with this PR.


mfcrocker (changes requested):
This is fine code-wise, but the validation piece is a problem from a user enumeration perspective

akssinhaamit:
This will be reworked on another companion trello card.

PR #672, opened 2021-02-16
Feature flag to turn off account email changes

akssinhaamit:
Account change will require further business discussions to come up with the further trello card. For the time being to make it not a blocker - introducing feature flag, so that the normal work can continue.

EDIT (from @insacuri):  I've changed this PR to instead only disable the email functionality while leaving the other details intact.  The feature flag mostly works, but I had to comment out a little bit to get it to work.  Getting the email change functionality is literally the next thing I'm going to be working on so whilst it's a bit ugly, it shouldn't be in there for too long.


insacuri (approved):

jdfox:
One small nitpick: could the comments about changes planned to be made once dependencies are completed (e.g. line 207 in models.py) be prefixed with "TODO:"? Otherwise, LGTM, if it's been tested.

jdfox (approved):

PR #676, opened 2021-02-22
User Self Serve - Help answering a survey question

akssinhaamit:
[Trello](https://trello.com/c/gFbVdknh/459-s28-user-self-serve-help-answering-a-survey-question)


frontstage/views/surveys/help/surveys_help.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+@surveys_bp.route('/help/<short_name>/<business_id>/<option>/<sub_option>/send-message', methods=['GET'])
+@jwt_authorization(request)
+def get_send_help_message_page(session, short_name, business_id, option, sub_option):
+    """Gets the send message page once the option and sub option is selected"""
+    if 'errors' in request.args:
+        errors = request.args['errors']
+        flash(errors)
+    subject, text_one, text_two = get_subject_and_breadcrumbs_title(sub_option, f'surveys/help/{short_name}/{option}')
```
insacuri, line 134: Is it possible to use more descriptive words then text_one and text_two?  What are they for?  It's not clear at a glance?

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -0,0 +1,197 @@
...
+
+    logger.info('Secure message sent successfully',
+                message_id=response['msg_id'], party_id=party_id, business_id=business_id)
+    return response
+
+
+def get_subject_and_breadcrumbs_title(option, uri):
+    """Gets the subject line for the secure message, the title of the breadcrumbs for sub options """
+    if option == 'do-not-have-specific-figures':
+        return 'I don’t have specific figures for a response', \
```
insacuri, line 191: If the subject and text_two are always the same and the text_one is basically hard coded.  Can you do something like:
```python
text_one = 'Help completing this survey'
if option == 'do-not-have-specific-figures':
    subject = 'I don’t have specific figures for a response'
    return subject, text_one, subject
if option == 'unable-to-return-by-deadline':
    subject = 'I’m unable to return the data by the deadline'
    return subject, text_one, subject
```

insacuri (changes requested):
A few bits to start on.  I want to test it out before I make any more comments

frontstage/views/surveys/help/surveys_help.py
insacuri: Also, thinking about it.  If outputs 1 and 3 are the same, why doesn't this function just return 2 things instead of 3?

akssinhaamit: This was still not clear with the this story but the next one will refactor it properly please ignore it for the time being.

frontstage/views/surveys/help/surveys_help.py
akssinhaamit: 👍

insacuri (approved):
LGTM 👍

PR #692, opened 2021-03-01
Bug for blank messages validation

akssinhaamit:
Currently when messages are blank frontstage errors out with 500. Fixing the bug with a proper validation.


insacuri (changes requested):
I tested this an the error above the box is: 'This page has1error' when I think it needs to be 'Message is required'

insacuri (approved):

PR #694, opened 2021-03-01
user account creation password change

akssinhaamit:
[Trello](https://trello.com/c/ZUkRoqTZ/368-s29-user-account-creation-change-my-password)

### DIFF
--- a/frontstage/templates/account/account-change-password.html
+++ b/frontstage/templates/account/account-change-password.html
@@ -81,14 +81,14 @@
         {% endif %}
         {{ form.csrf_token }}
             {{
-                onsInput({
+                onsPassword({
                     "id": "password",
                     "name": "password",
                     "type": "text",
                     "label": {
                         "text": 'Enter your current password'
                     },
-                    "value": password,
+                    "showPasswordText": "Show password",
                     "error": errorPassword,
                 })
             }}
--- a/frontstage/templates/account/account-change-password.html
+++ b/frontstage/templates/account/account-change-password.html
@@ -1,0 +1,162 @@
+{% extends "layouts/_twothirds.html" %}
+{% from "components/breadcrumb/_macro.njk" import onsBreadcrumb %}
+{% from "components/fieldset/_macro.njk" import onsFieldset %}
+{% from "components/input/_macro.njk" import onsInput %}
+{% from "components/button/_macro.njk" import onsButton %}
+{% from "components/lists/_macro.njk" import onsList %}
+{% from "components/password/_macro.njk" import onsPassword %}
+
+{% set page_title = "Change password" %}
+{% set breadcrumbsData = [
+  {
+    "text": "Surveys",
+    "url": "/surveys/todo"
+  },
+  {
+    "text": "Account",
+    "url": "/my-account"
+  }
+] %}
+
+{% block main %}
+{{
+    onsBreadcrumb({
+        "ariaLabel": "Breadcrumbs",
+        "id": "breadcrumbs",
+        "itemsList": breadcrumbsData
+    })
+}}
+{% if errors %}
+        {% set errorData = [] %}
+        {% if errors|length > 1 %}
+            {% set errorTitle = 'There are ' ~ errors|length ~ ' errors on this page' %}
+            {% for error in errors %}
+                {% do errorData.append(
+                    {
+                        "text": errors[error][0],
+                        "index": true,
+                        "url": "#" ~ error ~ '_error',
+                        "classes": "js-inpagelink"
+                    }
+                ) %}
+            {% endfor %}
+        {% else %}
+            {% set errorTitle = 'There is 1 error on this page' %}
+            {% for error in errors %}
+                {% do errorData.append(
+                    {
+                        "text": errors[error][0],
+                        "index": true,
+                        "url": "#" ~ error ~ '_error',
+                        "classes": "js-inpagelink"
+                    }
+                ) %}
+            {% endfor %}
+        {% endif %}
+
+        {% call
+            onsPanel({
+                "type": "error",
+                "classes": "u-mb-s",
+                "title":  errorTitle
+            })
+        %}
+            <p>This <strong>must be corrected</strong> to continue.</p>
+
+            {% from "components/lists/_macro.njk" import onsList %}
+            {{
+                onsList({
+                    "classes": "list--bare",
+                    "itemsList": errorData
+                })
+            }}
+        {% endcall %}
+    {% endif %}
+<h1 class="u-fs-xl">Change your password</h1>
+<form action="" method="post">
+        {% if errorType == 'failed' %}
+            {% set errorPassword = { "text": "Incorrect current password", "id": 'password_error' } %}
+        {% elif errors.password %}
+                {% set errorPassword = { "text": errors['password'][0],  "id": 'password_error' } %}
+        {% endif %}
+        {{ form.csrf_token }}
+            {{
+                onsInput({
+                    "id": "password",
+                    "name": "password",
+                    "type": "text",
+                    "label": {
+                        "text": 'Enter your current password'
+                    },
+                    "value": password,
+                    "error": errorPassword,
+                })
+            }}
+            <div class="field">
+                <h2 class="u-mt-l">Create a new password</h2>
+                <div class="message">
+                    <p>Your password must have:</p>
+                    <ul>
+                        <li>at least 8 characters </li>
+                        <li>at least 1 uppercase letter</li>
+                        <li>at least 1 symbol (eg: ?!£%) </li>
+                        <li>at least 1 number</li>
+                    </ul>
+                </div>
+                <div>
+                {%- call onsFieldset({
+                        "legend": "New password",
+                        "legendClasses": "u-vh"
+                    }) -%}
+                {% if errors.new_password %}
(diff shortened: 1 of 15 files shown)


insacuri (changes requested):
Tested it and 2 of the 3 fields have the password in plaintext

frontstage/templates/account/account-change-password.html
```diff
@@ -0,0 +1,162 @@
...
+<h1 class="u-fs-xl">Change your password</h1>
+<form action="" method="post">
+        {% if errorType == 'failed' %}
+            {% set errorPassword = { "text": "Incorrect current password", "id": 'password_error' } %}
+        {% elif errors.password %}
+                {% set errorPassword = { "text": errors['password'][0],  "id": 'password_error' } %}
+        {% endif %}
+        {{ form.csrf_token }}
+            {{
+                onsInput({
```
insacuri, line 84: This needs to be onsPassword

frontstage/templates/account/account-change-password.html
```diff
@@ -0,0 +1,162 @@
...
+                            "text": "New Password"
+                        },
+                        "showPasswordText": "Show password",
+                        "error": newPasswordError
+                    })
+                }}
+                </div>
+            </div>
+            {{
+                onsInput({
```
insacuri, line 129: This needs to be onsPassword.

It's not in the design, but I think this retype password field is missing a 'show password' checkbox.  The registration screen has 2 so theres no reason why this shouldn't have 2.  I've asked Jeremy about this.

frontstage/templates/account/account-contact-detail-change.html
```diff
@@ -149,7 +149,7 @@ <h1 class="u-fs-xl">Change your contact details</h1>
             "submitType": "timer"
             })
         }}
-        <a href="{{ url_for('account_bp.get_account', tag='todo') }}"
+        <a href="{{ url_for('account_bp.get_account') }}"
       role="button" class="btn btn--secondary" id="btn-upload-ci-cancel"><span class="btn__inner">Cancel</span></a>
```
insacuri, line 153: I know you didn't write this (I think it was me who did sorry...), but can you change the id of this button to something more sensible.

frontstage/views/account/account.py
```diff
@@ -31,8 +38,42 @@ def update_account(session):
...
+        try:
+            auth_controller.sign_in(username, password)
+            bound_logger.info("Attempting to change password via party service")
+            party_controller.change_password(username, new_password)
+            bound_logger.info("password changed via party service")
+            flash('Your password has been changed.')
+            return redirect(url_for('surveys_bp.get_survey_list', tag='todo'))
+        except AuthError as exc:
+            error_message = exc.auth_error
+            if BAD_AUTH_ERROR in error_message:
```
insacuri, line 69: Can you change this variable to `BAD_CREDENTIALS_ERROR` because that's what it is, and calling it this would make it clearer

frontstage/views/account/account.py
```diff
@@ -31,8 +38,42 @@ def update_account(session):
...
+        username = respondent_details['emailAddress']
+        password = request.form.get('password')
+        new_password = request.form.get('new_password')
+        if new_password == password:
+            return render_template('account/account-change-password.html', form=form,
+                                   errors={"new_password": ["Your new password is the same as your old password"]})
+        bound_logger = logger.bind(email=obfuscate_email(username))
+        bound_logger.info("Attempting to find user in auth service")
+        try:
+            auth_controller.sign_in(username, password)
```
insacuri, line 61: Holy moly this line needs a comment!  Something like 
`We call the sign in function as it's the only way to verify that the password provided is correct`

That wasn't clear when I first read it, and a comment would save someone from having to wonder why we're calling a sign_in function at a glance.

tests/integration/mocked_services.py
```diff
@@ -144,3 +144,4 @@
 url_notify_party_and_respondent_account_locked = f'{app.config["PARTY_URL"]}/party-api/v1/respondents/edit-account-' \
                                                  f'status/{party["id"]}'
 url_banner_api = f"{app.config['BANNER_SERVICE_URL']}/banner"
+url_respondent_party_by_id = f'{app.config["PARTY_URL"]}/party-api/v1/respondents/id/{party["id"]}'
```
insacuri, line 147: This didn't need to be added.  In the same file there is `url_get_respondent_party` which is identical.  Use this instead in your tests and remove this new one that you've added.

tests/integration/views/account/test_password_change.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertTrue('at least 1 number'.encode() in response.data)
+        self.assertTrue('New Password'.encode() in response.data)
+        self.assertTrue('Re-type your new password'.encode() in response.data)
+
+    @requests_mock.mock()
+    @patch('frontstage.controllers.party_controller.get_respondent_party_by_id')
+    def test_account_password_change_success(self, mock_request, get_respondent_party_by_id):
+        mock_request.get(url_banner_api, status_code=404)
+        mock_request.get(url_respondent_party_by_id, status_code=404)
+        mock_request.post(url_auth_token, status_code=200)
```
insacuri, line 54: This isn't right.  The token url in the auth service returns a 204 on success.  I know that other examples have it as 200, but they're wrong also.  You don't need to change those ones (but you can if you'd like), but this one needs to be changed.

tests/integration/views/account/test_password_change.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertTrue('at least 1 symbol (eg: ?!£%)'.encode() in response.data)
+        self.assertTrue('at least 1 number'.encode() in response.data)
+        self.assertTrue('New Password'.encode() in response.data)
+        self.assertTrue('Re-type your new password'.encode() in response.data)
+
+    @requests_mock.mock()
+    @patch('frontstage.controllers.party_controller.get_respondent_party_by_id')
+    def test_account_password_change_success(self, mock_request, get_respondent_party_by_id):
+        mock_request.get(url_banner_api, status_code=404)
+        mock_request.get(url_respondent_party_by_id, status_code=404)
```
insacuri, line 53: This doesn't make much sense.  A successful password change test would have a 404 happen if you got the party?  Even if you don't use the data, the mocking should still be set up correctly to simulate what would happen in the real world.

tests/integration/views/account/test_password_change.py
```diff
@@ -0,0 +1,130 @@
...
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue('Your password has been changed.'.encode() in response.data)
+        self.assertTrue('using your enrolment code'.encode() in response.data)
+
+    @requests_mock.mock()
+    @patch('frontstage.controllers.party_controller.get_respondent_party_by_id')
+    def test_account_password_change_form_validation_errors_does_not_meet_requirement(self, mock_request,
+                                                                                      get_respondent_party_by_id):
+        mock_request.get(url_banner_api, status_code=404)
+        mock_request.get(url_respondent_party_by_id, status_code=404)
```
insacuri, line 70: Same as above,  for this (and all the other tests) change the auth_token to 204 status (from 200), and change this to a 200 with data (from 404), even if you're not going to use the data.

insacuri (changes requested):

frontstage/templates/account/account-change-password.html
akssinhaamit: I did it because of the prototype. Will change it :-(

frontstage/templates/account/account-contact-detail-change.html
akssinhaamit: looks like I did write it.. will change it

frontstage/views/account/account.py
akssinhaamit: ok will change it, but I don't think `BAD_AUTH_ERROR` is bad or changing will make it clear :-)

frontstage/views/account/account.py
akssinhaamit: ahh.. I thought it was standard to validate a password via signin. Ok will add the comment.

tests/integration/mocked_services.py
akssinhaamit: Actually I don't need this. you are right

tests/integration/views/account/test_password_change.py
akssinhaamit: Actually don't require this. 👍

tests/integration/views/account/test_password_change.py
akssinhaamit: 👍

tests/integration/views/account/test_password_change.py
akssinhaamit: ok

frontstage/views/account/account.py
insacuri: It could be our naming convention now that I think about it.  If `sign_in` (which it doesn't actually do) was called `verify_credentials` (which is what it's actually doing) then a comment wouldn't be needed.  `sign_in` kinda implies a check + action to make you be signed in. All this is, is a quick thumbs up/down on whether or not the credentials are correct

insacuri:
Just need to change that input field into a password field and it's good to go!

insacuri (approved):

PR #698, opened 2021-03-04
Bug fix for password change

akssinhaamit:
Missed special character in regex. Thinking about it.. there could be any special character hence removing regex and using the existing way of validating password.


insacuri (approved):
Not tested it, but I assume you have.  Looks fine from a code perspective ^^

PR #703, opened 2021-03-12
Design system 32.0.1

jeremygillies:
# What and why?
Updated design system to more recent version (32.0.1)

# How to test?
Visually review all pages for inconsistencies - should match design system https://ons-design-system.netlify.app/

# Trello
https://trello.com/c/M5iLdksw/535-s29a-upgrading-design-system-on-frontstage-to-3201


insacuri (commented):
Looks good from a code perspective.  In testing I noticed there is a big whitespace gap between the bottom of the blue banner and the breadcrumbs.  When I tinkered and removed the `page__main` class from the main tag, it seemed to remove the big gap.

If it's not an issue, I'll approve it, but I thought I'd comment instead of explicitly approving/denying it.

insacuri (approved):
We chatted and the gap is going to be addressed on another card

akssinhaamit (approved):
The template changes looks ok to me. Most of the changes are .njk files which will be a part of the design system so expecting it to be correct. Approving the PR.

PR #704, opened 2021-03-16
Share survey story one

akssinhaamit:
# What and why?
New feature 
# How to test?
Smoke test and business approval

# Trello
[Trello](https://trello.com/c/jZ8tuxCz/483-s30-user-account-creation-share-surveys-setup-of-sharing-link-from-ras-story-1)


frontstage/controllers/party_controller.py
```diff
@@ -510,3 +511,39 @@ def notify_party_and_respondent_account_locked(respondent_id, email_address, sta
...
+    bound_logger.info('Getting enrolment data for the party')
+    enrolment_data = list(get_respondent_enrolments(party_id))
+    business_ids = set()
+    for enrolment in enrolment_data:
+        business_ids.add(enrolment['business_id'])
+    bound_logger.info('Getting businesses against business ids')
+    return get_business_by_business_id(list(business_ids))
+
+
+def get_business_by_business_id(business_ids):
```
insacuri, line 527: There is a function that does this already (albeit with a slightly different url) `get_party_by_business_id` is this new one needed?

frontstage/controllers/party_controller.py
```diff
@@ -510,3 +511,39 @@ def notify_party_and_respondent_account_locked(respondent_id, email_address, sta
         raise ApiError(logger, response)
 
     bound_logger.info('Successfully notified respondent and party service that account is locked')
+
+
+def get_list_of_business_for_party(party_id):
```
insacuri, line 516: Docstring here would do a lot of work.  
What party_id is it?  We still live in a world where a party_id can be a business or a respondent.

frontstage/controllers/party_controller.py
```diff
@@ -510,3 +511,39 @@ def notify_party_and_respondent_account_locked(respondent_id, email_address, sta
...
+    url = f'{app.config["PARTY_URL"]}/party-api/v1/businesses'
+    response = requests.get(url, params=params, auth=app.config['BASIC_AUTH'])
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        raise ApiError(logger, response)
+    return response.json()
+
+
+def get_surveys_listed_against_party_and_business_id(business_id, party_id):
```
insacuri, line 539: Doctstring here too.  Having `:rtype: list` will help future devs know what this function will return

frontstage/controllers/party_controller.py
akssinhaamit: done!

frontstage/controllers/party_controller.py
akssinhaamit: done

frontstage/controllers/party_controller.py
akssinhaamit: done

frontstage/controllers/party_controller.py
```diff
@@ -514,19 +514,31 @@ def notify_party_and_respondent_account_locked(respondent_id, email_address, sta
...
+    return get_business_by_id(list(business_ids))
 
 
-def get_business_by_business_id(business_ids):
+def get_business_by_id(business_id):
+    """
+    returns list of businesses associated with a business id or a list of business ids
+    :param business_id: This takes a single business id or a list of business ids
+    :return: business
+    :rtype: json
```
insacuri, line 538: json isn't a valid return type.  It'd be a dict

frontstage/views/account/account_survey_share.py
```diff
@@ -0,0 +1,131 @@
...
+    ConfirmEmailChangeForm
+from frontstage.views.account import account_bp
+
+logger = wrap_logger(logging.getLogger(__name__))
+
+
+@account_bp.route('/share-surveys', methods=['GET'])
+@jwt_authorization(request)
+def share_survey_overview(session):
+    flask_session['business_selected'] = None
```
insacuri, line 24: It seems like these are being set to none too often.  Is it being done to make sure you start the process with a clean slate?  If so, it'd be better to delete the key to keep the session clean instead of setting it to none (and leaving an empty key there).

Also the naming is fine but I think we can do better to make it easier for future devs.  In the session how you have it,  it's basically a dict that looks like:
```python
{
    'business_selected': 'some data..',
    'surveys_selected': ['data'],
    'receipient_email_address': 'an email address',
    ....other session data
}```

it might be better to organise it like:
```python
{
    'share_survey': {
        'business_selected': 'some data..',
        'surveys_selected': ['data'],
        'receipient_email_address': 'an email address',
    }
    ....other session data
}```

this way it's crystal clear what the data is for, and the possibility of name clashes in the future get reduced to nearly zero.

File: frontstage/controllers/party_controller.py [hunk 6]
```diff
@@ -552,15 +551,12 @@ def get_surveys_listed_against_party_and_business_id(business_id, party_id):
...
     :return: list of surveys
     :rtype: list
     """
-    enrolment_data = list(get_respondent_enrolments(party_id))
-    survey_ids = set()
-    for enrolment in enrolment_data:
-        if enrolment['business_id'] == business_id:
-            survey_ids.add(enrolment['survey_id'])
+    enrolment_data = get_respondent_enrolments(party_id)
+    survey_ids = {enrolment['survey_id'] for enrolment in enrolment_data if enrolment['business_id'] == business_id}
```
akssinhaamit, line 559: @insacuri - What happens in duplicate business?

frontstage/views/account/account_survey_share.py
akssinhaamit: having share_survey key as a consolidated group is a good idea. Can do that. Deleting a session key will need to make sure that the key exists in the session. Which will have additional logic everywhere, to avoid that its better to make a value of None rather deleting it.

frontstage/controllers/party_controller.py
insacuri: Functionally this is the same as what was there before. Using curly braces makes it a set comprehension.  Like a set, if you add the same thing twice, then duplicates won't be added.

frontstage/views/account/account_survey_share.py
akssinhaamit: said that, it seems that having a dict as the flask session is not as efficient as it looks and requires more for maintaining it. A key value pair for the session key seems like the right approach. I will provide a prefix as discussed to make sure its not too generic.

insacuri (approved):

PR #707, opened 2021-03-22
Fix integration test that missing mocked urls

insacuri:
# What and why?

One of the password change tests had a warning that there was a missing mocked.  This was around the survey list screen.  Tried to mock the urls but there were too many.  Instead this replicates (roughly, I'm returning a list rather then a generator) what the `get_survey_list_details_for_party` function returns to bypass the need to mock a bunch of stuff

# How to test?

# Trello

### DIFF
--- a/tests/integration/views/account/test_password_change.py
+++ b/tests/integration/views/account/test_password_change.py
@@ -31,11 +31,23 @@ class TestSurveyList(unittest.TestCase):
 
     @requests_mock.mock()
     @patch('frontstage.controllers.party_controller.get_respondent_party_by_id')
-    def test_account_password_change_success(self, mock_request, get_respondent_party_by_id):
+    @patch('frontstage.controllers.party_controller.get_survey_list_details_for_party')
+    def test_account_password_change_success(self, mock_request, get_survey_list, get_respondent_party_by_id):
+        survey_list = [{"case_id": "6f698975-0a36-45ff-ba66-7a575e414023",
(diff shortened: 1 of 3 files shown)


tests/integration/views/account/test_password_change.py
```diff
@@ -31,11 +31,23 @@ def tearDown(self):
 
     @requests_mock.mock()
     @patch('frontstage.controllers.party_controller.get_respondent_party_by_id')
-    def test_account_password_change_success(self, mock_request, get_respondent_party_by_id):
+    @patch('frontstage.controllers.party_controller.get_survey_list_details_for_party')
```
akssinhaamit, line 34: why do we need survey_list for changing password?

akssinhaamit (changes requested):

tests/integration/views/account/test_password_change.py
insacuri: Have you ran the test?  It raises a warning because of urls that aren't mocked.  The reason is because after you change a password, you're redirected to the todo page.
The test makes sure you're on the todo page, but the test warns you that you can't get all the info for it.  Which is kind of surprising, normally it outright fails if a mock hasn't been done, possibly because of the way it uses threads to build its own cache? 

Either way, this doesn't change functionality, it just suppresses the error by patching what get_survey_list_details_for_party (the function that does all the heavy lifting for the todo page) would give you

akssinhaamit (approved):

PR #711, opened 2021-04-01
Updated DS templates to fix csp_nonce issue

boxadesign:
# What and why?

This is a fix to the templates so that the csp_nonce function can run.

There will be a design system release to provide a solution to the different implementations.

# How to test?

# Trello

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -91,21 +91,21 @@
         },
         "cryptography": {
             "hashes": [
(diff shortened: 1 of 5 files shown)


akssinhaamit (approved):

akssinhaamit:
This issue has been fixed but another console error is visible now only on sample summary page
```
18.js:1 Uncaught (in promise) TypeError: Cannot read property 'classList' of undefined
    at new o (18.js:1)
    at main.js:6
    at Array.forEach (<anonymous>)
    at main.js:6
    at Generator.next (<anonymous>)
    at o (main.js:1)
    at i (main.js:1)
    ```

PR #736, opened 2021-05-24
added party id to eq payload

anwilkie:
# What and why?
The party ID needed to be added to the EQ payload to facilitate the "help" link.

# Trello
[Card](https://trello.com/c/cENWYFNf)


tests/integration/test_generate_eq.py
```diff
@@ -220,3 +220,22 @@ def test_generate_eq_url_non_mandatory_event_date_is_returned(self):
 
         response = EqPayload()._find_event_date_by_tag('employment', collex_events_dates, '123', False)
         self.assertEqual(response, '2018-04-03')
+
+    @requests_mock.mock()
+    def test_generate_eq_payload_has_party_id(self, mock_request):
+
+        mock_request.get(url_get_case, json=case)
```
insacuri, line 227: This whole block of mock requests can be replaced with just these 4:

```python
mock_request.get(url_get_collection_exercise, json=collection_exercise)
mock_request.get(url_get_collection_exercise_events, json=collection_exercise_events)
mock_request.get(url_get_business_party, json=business_party)
mock_request.get(url_get_ci, json=collection_instrument_eq)
```
You shouldn't just copy a bunch of them from another test >_<  If you look at what the `create_payload` function is doing, you can see it's only doing calls to collection exercise, party and collection instrument, so you don't need the case mocks as they'll never be called.

insacuri (changes requested):

insacuri (approved):

warrenbailey (changes requested):
This is really good work @anwilkie however we're now taking a different approach to this so unfortunately it is no longer needed. Please discuss with @insacuri and then close this PR.

insacuri:
Closing as it's not needed anymore

PR #741, opened 2021-06-17
Help url stuff

anwilkie:
# What and why?

The help pages needed to be exposed to the EQ payload by using a `survey_ref` and `ru_ref` in the URLs. These variables are, in turn, used to retrieve the business ID and survey short name.

# How to test?

Log into frontstage, click on "help with this survey", and navigate around the various options. They should all work, and you should see a survey ref (i.e. `139` for QBS) and an ru ref (i.e. `49900000001`) in the help page URLs.

# Trello

[Card](https://trello.com/c/cENWYFNf)


warrenbailey (commented):
I think the issue you have for all of these urls  is that you've changed the get_send_help_message_page function to accept survey survey ref and ru ref, but when it's being called via url_for its still trying to use the short name and business ids as well as the new parameters, so it's failing to find the function and just saying on the same page.

Either you have to update all the url_for calls in the templates and remove short_name and business_id or add an additional url_for function for getting the help pages by survey ref and ru ref and leave everything else as is.

mfcrocker:
Generally looks good, but your breadcrumbs aren't working and are returning our good friend the 500 screen. It's likely you need to update the URL for them

mfcrocker (changes requested):
Generally LGTM, but from testing your breadcrumbs are returning 500s in places. You probably need to update the URL there.

mfcrocker (approved):
LGTM

_infra/helm/frontstage/Chart.yaml
```diff
@@ -18,5 +18,5 @@ version: 1.2.1
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.2.99
 
+appVersion: 2.2.100
```
mfcrocker, line 22:
```suggestion
appVersion: 2.3.0
```

This isn't your fault or anything, I just think if we're getting to patch number 100 it's probably time to roll that sucker over 🤣

mfcrocker (approved):

PR #742, opened 2021-06-17
user-account creation something else changes for frontstage

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/XQ2gnUbP/369-s37-user-account-creation-something-else

### DIFF
--- a/config.py
+++ b/config.py
@@ -72,7 +72,8 @@ class Config(object):
     ACCOUNT_EMAIL_CHANGE_ENABLED = bool(strtobool(os.getenv('ACCOUNT_EMAIL_CHANGE_ENABLED', "False")))
     SHARE_SURVEY_ENABLED = bool(strtobool(os.getenv('SHARE_SURVEY_ENABLED', "True")))
     MAX_SHARED_SURVEY = int(os.getenv('MAX_SHARED_SURVEY', '50'))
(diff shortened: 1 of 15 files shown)


config.py
```diff
@@ -72,6 +72,7 @@ class Config(object):
     ACCOUNT_EMAIL_CHANGE_ENABLED = bool(strtobool(os.getenv('ACCOUNT_EMAIL_CHANGE_ENABLED', "False")))
     SHARE_SURVEY_ENABLED = bool(strtobool(os.getenv('SHARE_SURVEY_ENABLED', "True")))
     MAX_SHARED_SURVEY = int(os.getenv('MAX_SHARED_SURVEY', '50'))
+    TECHNICAL_MESSAGE_ENABLED = bool(strtobool(os.getenv('SHARE_SURVEY_ENABLED', "True")))
```
mfcrocker, line 75:
```suggestion
    TECHNICAL_MESSAGE_ENABLED = bool(strtobool(os.getenv('TECHNICAL_MESSAGE_ENABLED', "True")))
```

mfcrocker (changes requested):
One slight bug

config.py
creativeappdevamit: Thanks, it would have been a disaster 👍 , will change it

creativeappdevamit: fixed

mfcrocker (approved):

PR #748, opened 2021-07-14
Validation update for duplicate transfer share

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/N7aTNPw7/759-user-account-creation-transfer-and-sharing-surveys-validation-update

### DIFF
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -647,7 +647,7 @@ def register_pending_shares(payload):
     register new entries to party for pending shares
     :param payload: pending shares entries dict
     :return: success if post completed
-    :rtype: dict
+    :rtype: response object
     """
     logger.info("Attempting register pending shares")
     url = f'{app.config["PARTY_URL"]}/party-api/v1/pending-surveys'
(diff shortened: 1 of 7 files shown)


frontstage/controllers/party_controller.py
```diff
@@ -659,7 +659,7 @@ def register_pending_shares(payload):
             logger.info("share survey has already been shared, hence ignoring this request.")
         else:
             raise ApiError(logger, response)
-    return response.json()
+    return response
```
warrenbailey, line 662: the :rtype: in the doc string still states dict when it's now a response object

Though i'm not sure you even need to return the response object at all as you don't seem to use it, you could just return a boolean indicating if its successful or not, and maybe an error message if its not.

Mind you what you have will work so I don't mind either way.

warrenbailey (changes requested):

frontstage/controllers/party_controller.py
creativeappdevamit: using response object to check the status code, will modify the doc string

warrenbailey (approved):

PR #752, opened 2021-07-22
created redirect to maintenance page

anwilkie:
# What and why?
The purpose of this PR is to demonstrate a prototype for the maintenance page redirect.

# How to test?
Change the UNDER_MAINTENANCE variable in config.py to True and then access frontstage to check that it redirects to the maintenance page template.

# Trello
[Card](https://trello.com/c/0lGBGkDQ)


frontstage/__init__.py
```diff
@@ -63,3 +64,10 @@ def inject_availability_message():
 
 # Bind routes to app
 import frontstage.views  # NOQA
+
+
+@app.before_request
+def redirect_to_maintenance_page():
+    under_maintenance = True
```
insacuri, line 71: So this PR, if merged right now, would just endlessly render that one template and disable all the other functionality.... seriously?
This clearly needs more work doing to it to make it configurable and a test to show the result with it switched on

insacuri (changes requested):

config.py
```diff
@@ -73,6 +73,7 @@ class Config(object):
     SHARE_SURVEY_ENABLED = bool(strtobool(os.getenv("SHARE_SURVEY_ENABLED", "True")))
     MAX_SHARED_SURVEY = int(os.getenv("MAX_SHARED_SURVEY", "50"))
     TECHNICAL_MESSAGE_ENABLED = bool(strtobool(os.getenv("TECHNICAL_MESSAGE_ENABLED", "True")))
+    UNDER_MAINTENANCE = os.getenv("UNDER_MAINTENANCE", False)
```
insacuri, line 76: Can change this to be more like the one above? make the default value a string and do the strtobool and bool stuff to get it into the right type.

It's a bit fiddly, but `os.getenv`  (https://docs.python.org/3/library/os.html#os.getenv) should have the return value and deafult value be a string so we should abide by this.

insacuri (changes requested):
One small change but looks good

insacuri (approved):

PR #756, opened 2021-08-12
Fixed issue with not retaining user information after a single

creativeappdevamit:
Fixed issue with not retaining user information after a single validation failure on account registration

# What and why?

# How to test?

# Trello
https://trello.com/c/soD3v81b/821-retain-data-on-account-sign-up-screen-when-error-has-been-made

### DIFF
--- a/frontstage/templates/register/register-pending-survey-enter-your-details.html
+++ b/frontstage/templates/register/register-pending-survey-enter-your-details.html
@@ -56,7 +56,11 @@
         {{ form.batch_no }}
         {{ form.is_transfer }}
         <h2>Your name</h2>
-
+        {% if errors %}
+            {% set firstName = form.first_name.data %}
+            {% set lastName = form.last_name.data %}
+            {% set phoneNumber = form.phone_number.data %}
+        {% endif %}
         {%- call onsFieldset({
             "legend": "Your name",
             "legendClasses": "u-vh"
@@ -72,7 +76,8 @@
                     "label": {
                         "text": "First name"
                     },
-                    "error": errorFirstName
+                    "error": errorFirstName,
+                    "value": firstName
                 })
             }}
             {% if errors.last_name %}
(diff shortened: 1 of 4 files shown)


frontstage/templates/register/register-pending-survey-enter-your-details.html
```diff
@@ -56,7 +56,11 @@ <h1 class="u-fs-xl">Create an account</h1>
         {{ form.batch_no }}
         {{ form.is_transfer }}
         <h2>Your name</h2>
-
+        {% if errors %}
```
insacuri, line 59: Does this need to only happen if there is an error?  What happens if you just get the data from an empty form?

frontstage/templates/register/register.enter-your-details.html
```diff
@@ -54,13 +54,19 @@ <h1 class="u-fs-xl">Enter your account details</h1>
         {{ form.enrolment_code }}
 
         <h2>Your name</h2>
-
+        {% if errors %}
+            {% set firstName = form.first_name.data %}
```
insacuri, line 58: Same with this one, Does this need to only happen if there is an error? What happens if you just get the data from an empty form?

frontstage/templates/register/register-pending-survey-enter-your-details.html
creativeappdevamit: We need to retain the data on the UI after there is an error. So we will land up on the same view only in case of error.

frontstage/templates/register/register.enter-your-details.html
creativeappdevamit: We need to retain the data on the UI after there is an error. So we will land up on the same view only in case of error. Data from empty form will not have this attribute set and hence will be shown blank.

insacuri (approved):

PR #766, opened 2021-09-24
Survey not listed on todo tab and refactor

creativeappdevamit:
# What and why?
* S43a User Self Serve - Where is my survey "to do" page link
* S44 Incorrect subject in frontstage help page
* Incorrect bread crumb name
# How to test?

# Trello
https://trello.com/c/cjGG7F01/930-s43a-user-self-serve-where-is-my-survey-to-do-page-link
https://trello.com/c/ctakC5Zu/941-s44-incorrect-subject-in-frontstage-help-page
https://trello.com/c/CoewPFZG/957-incorrect-bread-crumb-name

### DIFF
--- a/frontstage/templates/secure-messages/help/secure-message-send-messages-view.html
+++ b/frontstage/templates/secure-messages/help/secure-message-send-messages-view.html
@@ -28,7 +28,7 @@
         "id": "b-item-8"
     },) %}
 {% endif %}
-{% if breadcrumb_title_two is defined and breadcrumb_title_two != None%}
+{% if breadcrumb_title_two is defined and breadcrumb_title_two != None %}
 {% do breadcrumbsData.append({
         "text": breadcrumb_title_two,
         "url": "/surveys/help/"~option~"/"~sub_option,
--- a/frontstage/views/surveys/help/surveys_help.py
+++ b/frontstage/views/surveys/help/surveys_help.py
@@ -1,7 +1,9 @@
 import json
 import logging
 
-from flask import abort, flash, render_template, request, url_for
+from flask import abort, flash, render_template, request
+from flask import session as flask_session
+from flask import url_for
 from markupsafe import Markup
 from structlog import wrap_logger
 from werkzeug.utils import redirect
@@ -59,8 +61,10 @@ subject_text_mapping = {
     "how-safe-is-my-data": "How safe is my data?",
     "my-survey-is-not-listed": "My survey is not listed",
     "something-else": "Something else",
-    "help-completing-this-survey": "Help answering a survey question",
+    "answer-survey-question": "Help answering a survey question",
     "info-about-the-ons": info_about_the_ons,
+    "completing-this-survey-something-else": help_completing_this_survey_title,
+    "info-ons-something-else": info_about_the_ons,
 }
 breadcrumb_text_mapping = {
     "do-not-have-specific-figures": [help_completing_this_survey_title, "I don’t have specific figures for a response"],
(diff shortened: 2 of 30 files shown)


frontstage/templates/secure-messages/help/secure-message-send-messages-view.html
```diff
@@ -14,24 +14,24 @@
...
+{% if breadcrumb_title_one is defined and  breadcrumb_title_one != None %}
 {% do breadcrumbsData.append({
         "text": breadcrumb_title_one,
-        "url": "/surveys/help/"~survey_ref~"/"~ru_ref~"/"~option,
+        "url": "/surveys/help/"~option,
         "id": "b-item-8"
     },) %}
 {% endif %}
-{% if breadcrumb_title_two is defined %}
+{% if breadcrumb_title_two is defined and breadcrumb_title_two != None%}
```
insacuri, line 31: Need a space between None and %}

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -393,6 +290,58 @@ def send_help_message(session, survey_ref, ru_ref):
         return redirect(url_for("surveys_bp.get_survey_list", tag="todo"))
 
 
+def abort_help_if_session_not_set():
+    """
+    Aborts the process if flask session attributes are not set
+    """
+    if flask_session["help_survey_ref"] is None or flask_session["help_ru_ref"] is None:
```
insacuri, line 297: I don't think this will work as you expect.  If you  try and get a key using the `dict[key]` notation then it'll raise a `KeyError` if it's not found.

These both should be  changed to:
```python
if flask_session.get("help_survey_ref") is None or flask_session.get("help_ru_ref") is None:
```
as `.get(key)` will return None if it doesn't find it.

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -393,6 +290,58 @@ def send_help_message(session, survey_ref, ru_ref):
...
+        abort(404)
```
insacuri, line 298: Can you add a log.info line saying `both keys not in session, aborting` or something like that?

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -393,6 +290,58 @@ def send_help_message(session, survey_ref, ru_ref):
         return redirect(url_for("surveys_bp.get_survey_list", tag="todo"))
 
 
+def abort_help_if_session_not_set():
+    """
+    Aborts the process if flask session attributes are not set
+    """
```
insacuri, line 296: Can you add to this:
```python
:raises HTTPException: Raised when help_survey_ref and help_ru_ref are both not set in the session
```

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -393,6 +290,58 @@ def send_help_message(session, survey_ref, ru_ref):
...
+    if flask_session["help_survey_ref"] is None or flask_session["help_ru_ref"] is None:
+        abort(404)
+
+
+def get_selected_survey_business_details():
```
insacuri, line 301: Can you change this to:
```python
def get_selected_survey_business_details() -> tuple[str, str, str, str, str]:
```
Also, a function returning 5 values is...uncommon.  It's probably fine, but maybe it should  be a dict when it's that many values?  I don't know what the best practice is around this

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -296,89 +235,47 @@ def get_help_option_sub_option_select(session, survey_ref, ru_ref, option, sub_o
...
 @jwt_authorization(request)
-def get_send_help_message_page(session, survey_ref, ru_ref, option, sub_option):
+def get_send_help_message_page(session, option, sub_option):
     """Gets the send message page once the option and sub option is selected"""
-    short_name, business_id = get_short_name_and_business_id(survey_ref, ru_ref)
+    abort_help_if_session_not_set()
+    business_id, ru_ref, short_name, survey, survey_ref = get_selected_survey_business_details()
     subject = subject_text_mapping.get(sub_option)
-    text = breadcrumb_text_mapping.get(sub_option)
+    text = breadcrumb_text_mapping.get(sub_option, None)
```
insacuri, line 249: Can you change this to `breadcrumbs_text` or something like that?

insacuri (changes requested):
A few comments to get started.  The review isn't completed yet

frontstage/views/surveys/help/surveys_help.py
insacuri: Yeah, just checked it (by going to `http://localhost:8082/surveys/help/help-completing-this-survey` directly) and you get the `An error has occurred` page (which I think you get on a 500 error) rather then the 'Not found' you were probably expecting

frontstage/templates/secure-messages/help/secure-message-send-messages-view.html
creativeappdevamit: 👍

frontstage/views/surveys/help/surveys_help.py
creativeappdevamit: 👍

frontstage/views/surveys/help/surveys_help.py
creativeappdevamit: I might be right.. but I also don't know whats the best practice . I have updated the return type.

frontstage/views/surveys/help/surveys_help.py
creativeappdevamit: ok

insacuri (approved):

PR #769, opened 2021-11-05
Change Helm chart to take config type

mfcrocker:
# What and why?
Prod is currently running on Dev config - we need to be able to set this by environment

# How to test?

# Trello
https://trello.com/c/tbb3yi0u/1022-ensure-ras-frontstage-is-correct-in-prod-startup

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -26,76 +26,81 @@
         },
         "certifi": {
             "hashes": [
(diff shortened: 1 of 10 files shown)


insacuri (approved):
Looks good from a code standpoint.  I've not tested it but if you have then I'm happy

PR #779, opened 2021-11-24
Support 'Response Expires At'

mfcrocker:
# What and why?
There's a new optional field in the EQ payload called response expires at. For EQ v3 we need to provide this field, equal to the collection exercise end date in ISO_8601 format

# How to test?

# Trello
https://trello.com/c/mQbYzg2o/1129-s48-eq-v3-support-response-expires-at


insacuri (approved):
I've not tested it, but the change looks good from a code standpoint :)

PR #781, opened 2021-12-02
Remove various feature flags

insacuri:
# What and why?

2 of the features are permanently switched on now so we don't need the feature flags anymore

# How to test?

# Trello


insacuri:
Need to make a PR to remove them separately as the technical message enabled PR is switched off in prod for some reason.

PR #785, opened 2021-12-03
Implement redis cache for survey results

insacuri:
# What and why?

Currently when the survey list is created, for the todo and history page, there is a cache of data so we don't need to keep asking the service for the same information.

The current implementation of the cache stores all the data in a big dictionary for the duration of the request.  The cache of data isn't shared between any of the other workers or replicas, so there is only a small time/effort saving.

This PR is the first step of many, we'll store the results of a search into redis with a key in the form of `frontstage:{category}:{uuid}`.  The category here would be survey, collectionexercise, etc.  The advantages of this is that all replicas and all workers can query this one central bit of information which should vastly reduce the number of calls frontstage makes for the same information.
This key will have a reasonably short expiry to keep the data fresh, and the intention is to potentially have different expiry for bits of information that go stale quicker (i.e., case data needs to be incredibly current, whereas survey data almost never changes, if ever)

# How to test?

- create account
- enrol in a few surveys for a few different businesses
- load up the todo and history screens.  It should work (and maybe be marginally faster?)  Survey might be one of the lowest impact caches so it's unlikely to see much difference unless you time how long it takes to generate in the old way, then time how much time is spent filling the cache and using in the new way over the course of multiple page requests.

# Trello

https://trello.com/c/EdyZfP5w


anwilkie (approved):
Tested out this PR according to the instructions at the top, and the To Do and History pages load quite quickly after enrolling a user (faster than I remember them usually taking). The acceptance tests pass, too, so I'm giving this an approval.

PR #791, opened 2021-12-14
Reduce duplicate backing service calls during EQ launch

insacuri:
# What and why?

# How to test?

# Trello

### DIFF
--- a/config.py
+++ b/config.py
@@ -69,7 +69,6 @@ class Config(object):
     PUBSUB_TOPIC = os.getenv("PUBSUB_TOPIC", "ras-rm-notify-test")
     CANARY_GENERATE_ERRORS = bool(strtobool(os.getenv("CANARY_GENERATE_ERRORS", "False")))
     MAX_SHARED_SURVEY = int(os.getenv("MAX_SHARED_SURVEY", "50"))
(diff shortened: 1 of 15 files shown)


config.py
```diff
@@ -88,7 +88,7 @@ class DevelopmentConfig(Config):
 
     ACCOUNT_SERVICE_URL = os.getenv("ACCOUNT_SERVICE_URL", "http://localhost:8082/surveys/todo")
     ACCOUNT_SERVICE_LOG_OUT_URL = os.getenv("ACCOUNT_SERVICE_LOG_OUT_URL", "http://localhost:8082/sign-in/logout")
-    EQ_URL = os.getenv("EQ_URL", "https://localhost:8086/session?token=")
```
creativeappdevamit, line 91: why change to http?

creativeappdevamit (commented):
Looks ok just one question

config.py
insacuri: When I ran frontstage locally it'd fail to get to mock-eq as it doesn't have a certificate so it kept failing.  Changing it to http fixes it, and all the others in this file are http which means that's probably the right move too

creativeappdevamit (approved):

PR #792, opened 2021-12-15
Refactor 'is_respondent_enrolled' function

insacuri:
# What and why?

The `is_respondent_enrolled` function had a weird return type where it could be either a boolean or a dict.  This is bad practice so this PR modifies that function and everything that uses it to only care about the value returned

# How to test?

# Trello

### DIFF
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -544,15 +544,13 @@ def display_button(status, ci_type):
     return not (ci_type == "EQ" and status in CLOSED_STATE)
 
 
-def is_respondent_enrolled(party_id, business_party_id, survey_short_name, return_survey=False):
-    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+def is_respondent_enrolled(party_id: str, business_party_id: str, survey: dict) -> bool:
     respondent = get_respondent_party_by_id(party_id)
     enrolments = get_respondent_enrolments(respondent)
     for enrolment in enrolments:
         if enrolment["business_id"] == business_party_id and enrolment["survey_id"] == survey["id"]:
-            if return_survey:
-                return {"survey": survey}
             return True
+    return False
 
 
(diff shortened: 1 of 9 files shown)


mfcrocker (approved):
Thanks for this, I hate that this is even a thing you can do in Python :D

frontstage/controllers/party_controller.py
```diff
@@ -544,15 +544,13 @@ def display_button(status, ci_type):
     return not (ci_type == "EQ" and status in CLOSED_STATE)
 
 
-def is_respondent_enrolled(party_id, business_party_id, survey_short_name, return_survey=False):
-    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+def is_respondent_enrolled(party_id: str, business_party_id: str, survey: dict) -> bool:
```
creativeappdevamit, line 547: why cant we raise an exception when it is not enrolled.

frontstage/views/surveys/download_survey.py
```diff
@@ -25,7 +26,8 @@ def download_survey(session):
 
     # Check if respondent has permission to download for this case
     case = case_controller.get_case_by_case_id(case_id)
-    party_controller.is_respondent_enrolled(party_id, business_party_id, survey_short_name)
+    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+    party_controller.is_respondent_enrolled(party_id, business_party_id, survey)
```
creativeappdevamit, line 30: same as above

frontstage/views/surveys/upload_survey.py
```diff
@@ -38,7 +39,8 @@ def upload_survey(session):
         )
 
     # Check if respondent has permission to upload for this case
-    party_controller.is_respondent_enrolled(party_id, business_party_id, survey_short_name)
+    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+    party_controller.is_respondent_enrolled(party_id, business_party_id, survey)
```
creativeappdevamit, line 43: raising an exception?

PR #796, opened 2022-01-12
Use updated sdc crypto

insacuri:
# What and why?

# How to test?

# Trello


config.py
```diff
@@ -35,6 +35,8 @@ class Config(object):
     ACCOUNT_SERVICE_URL = os.getenv("ACCOUNT_SERVICE_URL")
     ACCOUNT_SERVICE_LOG_OUT_URL = os.getenv("ACCOUNT_SERVICE_LOG_OUT_URL")
     EQ_URL = os.getenv("EQ_URL")
+    EQ_V3_URL = EQ_URL
+    # EQ_V3_URL = os.getenv("EQ_V3_URL") TODO uncomment and make this work for real
```
mfcrocker, line 39: We should be doing this as a feature flag rather than relying on us remembering to uncomment things

insacuri:
Closing as this approach isn't right.  We need to use the service field for the encryption keys and for our signing keys, we can use the same one between both eq v2 and v3

PR #800, opened 2022-01-18
Implement business party caching

arroyoAle:
# What and why?
Changing where the business party cache is stored, in the common redis cache rather than in instance cached_data. This is to match what has been done with survey and collection instrument
# How to test?

# Trello

### DIFF
--- a/frontstage/controllers/party_controller.py
+++ b/frontstage/controllers/party_controller.py
@@ -435,7 +435,7 @@ def get_survey_list_details_for_party(respondent: dict, tag: str, business_party
 
     # This is a dictionary that will store all the data that is going to be cached instead of making multiple calls
     # inside the for loop for get_respondent_enrolments.
-    cache_data = {"businesses": dict(), "collexes": dict(), "cases": dict()}
+    cache_data = {"collexes": dict(), "cases": dict()}
     redis_cache = RedisCache()
 
     # Populate the cache with all non-instrument data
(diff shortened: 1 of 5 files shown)


frontstage/controllers/party_controller.py
```diff
@@ -136,6 +135,13 @@ def get_party_by_business_id(party_id, party_url, party_auth, collection_exercis
     return response.json()
 
 
+"""
+This import had to be moved as redis_cache calls the get_party_by_business_id method which needs to be
+initialised before being used
+"""
+from frontstage.common.redis_cache import RedisCache  # NOQA: E402
```
anwilkie, line 142: I'm not entirely sure if this is how we should be handling imports. It seems a little awkward. Is there really not a more elegant solution of resolving this sort of circular import? Because the import is outside of a function, it will likely only be performed once, so it shouldn't affect the performance of the service, but it still bugs me a little. Did you try considering different ways of resolving this issue, or was this the first thing you came up with?

insacuri (approved):

frontstage/controllers/party_controller.py
insacuri: It's a tough one.  I agree with you that doing it like this is not ideal, this caching stuff is still kind-of in a proof of concept phase as I think it'll improve the performance but I can't be certain til it's in.
If this ends up improving things then I think we should have an exercise to refactor this so it's a bit tidier as I'm not 100% happy with how I originally implemented it.

I'm approving the PR as it works and I think this is a justifiably small bit of tech debt that we'll probably end up cleaning reasonably soon if it works; and most likely discarding/refactoring if it doesn't.  You were definitely right to pick up on this though!

PR #803, opened 2022-01-19
fixing timestamp format issue for response_expires_at

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/o2MsO0Gr/1300-fix-eq-v3-payload-issues


insacuri (approved):
Not tested it, but from a code standpoint it looks good!

PR #806, opened 2022-01-21
changed logging statements and added active iac check

anwilkie:
# What and why?
There have been errors and bugs relating to enrolment codes in frontstage recently. In prod on mobile devices, the same enrolment codes have been used more than once, and Safari browsers sometimes error when entering an enrolment code. The cause of these issues is unknown at the moment, but these amended log statements might shed light on the issue. In addition, prior to now, the code for checking the IAC when adding a new survey would explicitly check for the active state, yet for some reason, this check was missing when creating a new account. This PR makes the code more symmetrical between the two processes. I'm not sure if this will fix the bug with enrolment codes being able to be used more than once, but having an extra check in place won't hurt.

# Trello
[Card](https://trello.com/c/sNchCVKZ)

### DIFF
--- a/frontstage/views/surveys/add_survey.py
+++ b/frontstage/views/surveys/add_survey.py
@@ -36,11 +36,13 @@ def add_survey(_):
                 return render_template("surveys/surveys-add.html", form=form, data=template_data)
         except ApiError as exc:
             if exc.status_code == 400:
-                logger.info("Enrolment code already used when attempting to add survey", status_code=exc.status_code, enrolment_code=enrolment_code)
+                logger.info("Enrolment code already used when attempting to add survey", status_code=exc.status_code,
+                            enrolment_code=enrolment_code)
                 template_data = {"error": {"type": "failed"}}
                 return render_template("surveys/surveys-add.html", form=form, data=template_data)
             else:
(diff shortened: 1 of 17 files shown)


JacobHarrisonONS (approved):

frontstage/views/surveys/add_survey.py
```diff
@@ -27,25 +27,33 @@ def add_survey(_):
...
+                    "Failed to submit enrolment code when attempting to add survey",
+                    status_code=exc.status_code,
+                    enrolment_code=enrolment_code,
+                )
                 raise
 
         cryptographer = Cryptographer()
         encrypted_enrolment_code = cryptographer.encrypt(enrolment_code.encode()).decode()
-        logger.info("Successful enrolment code submitted")
+        logger.info("Successful enrolment code submitted when attempting to add survey", enrolment_code=enrolment_code)
```
creativeappdevamit, line 56: I don't think its successfully submitted
Can you add a log line to suggest that all validation has been completed and now we will decrypt 
another on that the decoding was done sucessfully
and once we actually save the info in db then we suggest it has been sumitted.

creativeappdevamit (changes requested):
Looks ok to me. Can we add some more logging along side cryptograher?

creativeappdevamit (approved):
Fix merge conflict

PR #807, opened 2022-01-24
Implement collection exercise caching

arroyoAle:
# What and why?
Changing where the collection exercise cache is stored, in the common redis cache rather than in instance cached_data. This is to match what has been done with survey and collection instrument and business party, also the population of enrolments has been changed since it used the entire collex dictionary

# How to test?

# Trello

### DIFF
--- a/frontstage/common/redis_cache.py
+++ b/frontstage/common/redis_cache.py
@@ -90,14 +90,14 @@ class RedisCache:
 
         return json.loads(result.decode("utf-8"))
 
-    def get_collection_exercise(self, key):
+    def get_collection_exercises_by_survey(self, key):
         """
         Gets the collection-exercise from redis or the collection-exercise service
 
         :param key: Key in redis (for this example will be a frontstage:collection-exercise:id)
         :return: Result from either the cache or collection-exercise service
         """
-        redis_key = f"frontstage:collection-exercise:{key}"
+        redis_key = f"frontstage:collection-exercise-by-survey-id:{key}"
         try:
             result = redis.get(redis_key)
         except RedisError:
--- a/frontstage/common/redis_cache.py
+++ b/frontstage/common/redis_cache.py
@@ -11,6 +11,7 @@ from frontstage.controllers.collection_instrument_controller import (
 )
 from frontstage.controllers.party_controller import get_party_by_business_id
 from frontstage.controllers.survey_controller import get_survey
+from frontstage.controllers.collection_exercise_controller import get_live_collection_exercises_for_survey
 
 logger = wrap_logger(logging.getLogger(__name__))
 
@@ -18,7 +19,8 @@ logger = wrap_logger(logging.getLogger(__name__))
 class RedisCache:
     SURVEY_CATEGORY_EXPIRY = 600  # 10 mins
     COLLECTION_INSTRUMENT_CATEGORY_EXPIRY = 600  # 10 mins
-    BUSINES_PARTY_CATEGORY_EXPIRY = 600
+    BUSINES_PARTY_CATEGORY_EXPIRY = 600  # 10 mins
+    COLLECTION_EXERCISE_CATEGORY_EXPIRY = 600  # 10 mins
 
     def get_survey(self, key):
         """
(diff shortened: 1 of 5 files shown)


frontstage/controllers/party_controller.py
```diff
@@ -438,21 +430,25 @@ def get_survey_list_details_for_party(respondent: dict, tag: str, business_party
...
+    cache_data = {"cases": dict()}
     redis_cache = RedisCache()
 
     # Populate the cache with all non-instrument data
     caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag)
 
-    enrolments = get_respondent_enrolments_for_started_collex(enrolment_data, cache_data["collexes"])
+    #  Populate the enrolments by creating a dictionary using the redis_cache
+    enrolments = get_respondent_enrolments_for_started_collex(
+        enrolment_data, dict((s_id, redis_cache.get_collection_exercise(s_id)) for s_id in surveys_ids)
```
insacuri, line 441: Change s_id to `survey_id`.  Easier to read at a glance and letters aren't at a premium ;)

frontstage/common/redis_cache.py
```diff
@@ -18,7 +21,8 @@
 class RedisCache:
     SURVEY_CATEGORY_EXPIRY = 600  # 10 mins
     COLLECTION_INSTRUMENT_CATEGORY_EXPIRY = 600  # 10 mins
-    BUSINES_PARTY_CATEGORY_EXPIRY = 600
+    BUSINES_PARTY_CATEGORY_EXPIRY = 600  # 10 mins
```
insacuri, line 24: Can you change the name of this to be the correct spelling of business?  I missed it in the last PR when it was added

frontstage/controllers/party_controller.py
```diff
@@ -356,13 +355,6 @@ def caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag):
     # in the cache_data.
     threads = []
 
-    for survey_id in surveys_ids:
```
insacuri, line 359: I think with this, the surveys_ids variable isn't used.  can you remove it from this function and the thing that calls it?

frontstage/controllers/party_controller.py
```diff
@@ -438,21 +430,28 @@ def get_survey_list_details_for_party(respondent: dict, tag: str, business_party
...
     redis_cache = RedisCache()
 
     # Populate the cache with all non-instrument data
     caching_data_for_survey_list(cache_data, surveys_ids, business_ids, tag)
 
-    enrolments = get_respondent_enrolments_for_started_collex(enrolment_data, cache_data["collexes"])
+    #  Populate the enrolments by creating a dictionary using the redis_cache
+    enrolments = get_respondent_enrolments_for_started_collex(
+        enrolment_data,
+        dict((survey_id, redis_cache.get_collection_exercises_by_survey(survey_id)) for survey_id in surveys_ids),
```
insacuri, line 442: It's pretty minor and a bit subjective, but can you split this line into its own variable above just to make it easier to read.  Also, you can use a dictionary comprehension instead so the line above should look like:
```python
collection_exercises = {survey_id: redis_cache.get_collection_exercises_by_survey(survey_id) for survey_id in surveys_ids}
```

frontstage/controllers/party_controller.py
```diff
@@ -438,21 +430,28 @@ def get_survey_list_details_for_party(respondent: dict, tag: str, business_party
...
         business_party = redis_cache.get_business_party(enrolment["business_id"])
         survey = redis_cache.get_survey(enrolment["survey_id"])
 
         # Note: If it ever becomes possible to get only live-but-not-ended collection exercises from the
         # collection exercise service, the filter_ended_collection_exercises function will no longer
         # be needed as we can request what we want instead of having to filter what we get.
-        live_collection_exercises = filter_ended_collection_exercises(cache_data["collexes"][survey["id"]])
+
+        live_collection_exercises = filter_ended_collection_exercises(
+            redis_cache.get_collection_exercises_by_survey(survey["id"])
```
insacuri, line 453: If you do the suggestion above, you can change this to 
```python
live_collection_exercises = filter_ended_collection_exercises(collection_exercises[survey["id"]])
```
this will prevent a call to redis.  While redis is fast, getting a local variable is much faster than getting it from redis

insacuri (approved):

PR #808, opened 2022-01-24
add check for dns

JacobHarrisonONS:
# What and why?
add flag to check if there is a dns for EQ 

# How to test?

# Trello
https://trello.com/c/1u49hEzI/1021-write-update-acceptance-test-to-utilise-a-completed-eq

### DIFF
--- a/_infra/helm/frontstage/templates/deployment.yaml
+++ b/_infra/helm/frontstage/templates/deployment.yaml
@@ -120,7 +120,7 @@ spec:
             {{- if .Values.dns.enabled }}
             value: "{{ .Values.eq.url }}"
             {{- else }}
-            value: "http://$(MOCK_EQ_SERVICE_HOST):$(MOCK_EQ_SERVICE_PORT)/session?token="
+            value: "http://mock-eq.{{ .Values.namespace }}:8086/session?token="
             {{- end }}
           - name: EQ_V3_URL
             {{- if .Values.dns.enabled }}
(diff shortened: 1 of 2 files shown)


_infra/helm/frontstage/templates/deployment.yaml
```diff
@@ -117,9 +117,19 @@ spec:
                 name: {{ .Chart.Name }}
                 key: email-salt
           - name: EQ_URL
+            {{- if .Values.dns.enabled }}
             value: "{{ .Values.eq.url }}"
+            {{- else }}
+            value: "http://mock-eq.{{ .Values.namespace }}.svc.cluster.local:{{ .Values.dns.wellKnownPort }}/session?token="
```
creativeappdevamit, line 123: doesn't seems right

_infra/helm/frontstage/templates/deployment.yaml
```diff
@@ -117,9 +117,19 @@ spec:
...
+            {{- end }}
           - name: EQ_V3_URL
+            {{- if .Values.dns.enabled }}
             value: "{{ .Values.eq.v3url }}"
+            {{- else }}
+            value: "http://mock-eq.{{ .Values.namespace }}.svc.cluster.local:{{ .Values.dns.wellKnownPort }}/v3/session?token="
```
creativeappdevamit, line 129: same as above something like `it should be something like `"http://$(MOCK_EQ_HOST):$(MOCK_EQ_SERVICE_PORT)"``

creativeappdevamit (changes requested):

PR #810, opened 2022-01-25
Improve efficiency and reduce complexity of generating survey list page

insacuri:
# What and why?

The current algorithm is something like:
- Get all the cases for the businesses the respondent is part of
- Get all survey ids for everything the respondent is enrolled in
- Get all collection exercises for those surveys
- Filter all the non-live collection exercises out (so you only have live ones)
- Create a list of cases that are only for live exercises
-  For each case
  - Do stuff to generate a row in the survey list table

This PR would change it to:
- Get all the cases for the businesses the respondent is part of 
- For each case
  - See if the case is for a survey the respondent is part of.  Go to the next case otherwise
  - Get the collection exercise for the case.
  - See if it's live,  go to the next case otherwise.
  - Do stuff to generate a row in the survey list table

This should work because cases are pre-filtered by status depending on the page it's for.  So the history page will only have a list of completed cases, and the todo page will have a list of NOTSTARTED cases.  We have to do the check for live collection exercises because a NOTSTARTED case can be for a collection exercise we've set up but not made live yet, and we don't want people filling those out!

I think this is simpler?  Regardless, we get all the cases for the businesses the respondent is part of.  The meat of the work is figuring out which cases need to be in the list.  The old way included a bit of faff but the new way, I think, is easier to understand.

# How to test?

# Trello

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -15,9 +15,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 
-version: 2.3.56
+version: 2.3.57
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 
-appVersion: 2.3.56
+appVersion: 2.3.57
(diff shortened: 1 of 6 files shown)


insacuri:
Closing because the idea didn't work and was pretty slow!

PR #812, opened 2022-01-27
add api documentation

JacobHarrisonONS:
# What and why?
Document api calls for frontstage

# How to test?

# Trello
https://trello.com/c/hONtDmUr/1312-task-document-all-the-endpoints-for-frontstage


creativeappdevamit (changes requested):
Kindly add which one requires login and which one doesn't

creativeappdevamit (changes requested):
Also we would like to know the frontstage endpoints and not the controllers endpoints which they are talking to.

Please look into the views for frontstage and document the endpoint exposed as a part of the frontstage

creativeappdevamit (commented):
Looks ok to me.. If we missed something we have a document to update

creativeappdevamit (approved):

PR #816, opened 2022-02-09
test upgrade

mybigopenbox:
# What and why?
Get an indication of what is involved
# How to test?
actual attempt
# Trello
https://trello.com/c/CDUtVJsb/1351-spike-upgrade-design-system-version-in-ras-frontstage-max-2-days-only


pricem14pc:
Redundant

PR #820, opened 2022-02-25
Adding response_id for EQ v3

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/2YkV9EQ7/1388-add-responseid-while-launching-eq-v3

### DIFF
--- a/tests/integration/test_generate_eq.py
+++ b/tests/integration/test_generate_eq.py
@@ -70,6 +70,7 @@ class TestGenerateEqURL(unittest.TestCase):
                     party_id=respondent_party["id"],
                     business_party_id=business_party["id"],
                     survey=survey_eq,
+                    version=None,
                 )
         self.assertEqual(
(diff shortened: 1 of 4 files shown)


tests/integration/test_generate_eq.py
```diff
@@ -94,12 +95,61 @@ def test_generate_eq_url_no_eq_id(self, mock_request):
...
+        with app.app_context():
+            payload = EqPayload().create_payload(
+                case,
+                collection_exercise,
+                party_id=respondent_party["id"],
+                business_party_id=business_party["id"],
+                survey=survey_eq,
+                version="v3",
+            )
+        self.assertTrue("response_id" in payload)
```
insacuri, line 127: Can you add a line that looks like:
```python
self.assertEquals(payload['response_id'], <whatever the response_id is meant to look like>)
```

just so people looking at the unit tests can have an idea of what it looks like.  This helps with the whole 'tests as documentation' thing

insacuri (approved):
One comment that isn't worth blocking the PR for, but would be good to get done.

PR #821, opened 2022-03-04
Bumping DS version to remove loadash.js vulnerability

creativeappdevamit:
# What and why?

# How to test?

# Trello
https://trello.com/c/XoqOBRbC/1421-out-of-date-loadashjs-in-use


insacuri (approved):
Deployed and tested it locally.  Seems to be okay from what I could see.

PR #822, opened 2022-03-08
Improve process to update Design System version

rmccar:
# What and why?
This PR updates, improves and documents the process to update the Design System macros in ras-frontstage.

- a `.design-system-version` file has been created where the Design System version can now be set
- the Design System version has been parameterised in the `load-templates` script and that script has been renamed to `load-design-system-templates` to better explain what it does
- the Design System templates have been added to the `.gitignore` file and been removed from the ras-frontstage repo
- the `load-design-system-templates` script has been added to run when the `start` script runs

# How to test?
- remove the Design System macros from your local copy of ras-frontstage (delete the `frontstage/templates/components` and `frontstage/templates/layout` folders)
- run `make start` and see that the Design System macros are loaded and the app runs correctly

# Trello


anwilkie (approved):
Switched to this branch and ran `make start` to confirm that design system stuff is loaded upon booting up the service.

Makefile
```diff
@@ -7,11 +11,11 @@ build-docker:
...
 	docker build -f _infra/docker/Dockerfile .
 
-start:
-	pipenv run python run.py
+load-design-system-templates:
+	pipenv run ./scripts/load_templates.sh $(DESIGN_SYSTEM_VERSION)
 
-UNIT_TESTS=tests/unit
-INTEGRATION_TESTS=tests/integration
+start: load-design-system-templates
```
creativeappdevamit, line 17: do we need to load templates if its the same version?

creativeappdevamit (changes requested):
I think we need the components and templates as we want the frontstage to be self sustain as it runs in the container. Please let me know if I am missing something.

.gitignore
```diff
@@ -100,4 +100,8 @@ ENV/
 .build
 
 # pyenv
-.python-version
\ No newline at end of file
+.python-version
+
+# Design System templates
+frontstage/templates/components
```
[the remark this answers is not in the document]
rmccar: These are now ignored because I don't think the DS templates should be checked into the RASRM repo because they aren't really part of the frontstage app codebase and should be pulled in each time

Makefile
rmccar: If the templates are no longer checked into the repo like I mentioned above, these will need to be loaded in each time and doing it this way will mean that a manual copy/run of the load script isn't needed to be done. You can just run `make start`. Also when changing DS versions, ensuring clearing out the templates and pulling them in again will automatically make sure the latest templates are being used and stop any problems such as if templates are renamed in the DS for example, when there could be the potential for both the old and new versions of a template to exist in the repo. I think this is all beneficial and doesn't slow down the `make start` command at all.

creativeappdevamit: I get that... for development purposes, it will work but as I said we deploy frontstage as a container images, I don't think curl commands can run while deployments.

creativeappdevamit (approved):

PR #830, opened 2022-03-15
Revert "Removing next parameter from url"

insacuri:
Reverts ONSdigital/ras-frontstage#829


mfcrocker:
I'm pretty much certain this isn't being caused by this PR - Response Ops is at fault

PR #834, opened 2022-03-23
remove beta key from service yaml

JacobHarrisonONS:
# What and why?
In ras-frontstage we're using a beta key in a helm chart where there is one without the beta. prefix. I have removed the prefix

# How to test?

# Trello
https://trello.com/c/Ca4f1uqJ/1153-update-service-in-ras-frontstage-to-not-use-beta-key-in-helm-chart


creativeappdevamit (commented):
https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features#same_backendconfig_for_all_service_ports

creativeappdevamit (approved):

PR #835, opened 2022-03-24
setting samesite cookie to strict

anwilkie:
# What and why?
During the pen test, the `authorization` cookie generated when a user logs in was discovered to not have the `samesite` setting set. This PR sets the cookie to `strict` (which is also used for the `authorization` cookie in response ops).

# How to test?
Log in to front stage with Google Chrome, click on the three dots in the top-right corner of your window, click on "more tools", then "developer tools", then in the developer tools menu, click "application" in the top bar, followed by "cookies" in the left menu. Ensure that the `authorization` cookie's `samesite` variable is set to `strict`.

# Trello
[Card](https://trello.com/c/lrVwM0jo)

### DIFF
--- a/frontstage/views/sign_in/sign_in.py
+++ b/frontstage/views/sign_in/sign_in.py
@@ -104,6 +104,7 @@ def login():  # noqa: C901
             expires=redis_session.get_expires_in(),
             secure=secure,
             httponly=secure,
+            samesite="strict",
         )
         count = conversation_controller.get_message_count_from_api(redis_session)
         redis_session.set_unread_message_total(count)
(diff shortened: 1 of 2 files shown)


mfcrocker (approved):

frontstage/views/sign_in/sign_in.py
```diff
@@ -104,6 +104,7 @@ def login():  # noqa: C901
             expires=redis_session.get_expires_in(),
             secure=secure,
             httponly=secure,
+            samesite="strict",
```
creativeappdevamit, line 107: why not secure?

anwilkie: You can only set `samesite` to `None`, `Lax`, or `Strict`. There is no `secure` option. Detailed here: https://blog.heroku.com/chrome-changes-samesite-cookie#how-is-chrome-protecting-users-against-csrf-attacks

PR #846, opened 2022-07-18
added ingress policy

anwilkie:
# What and why?
Frontstage needed to restrict security encryption to forbid users from using TLS version 1.0 or 1.1 (1.2 is required, in other words). Please see [this terraform PR ](https://github.com/ONSdigital/ras-rm-terraform/pull/227) for more info.

# How to test?
This will be tested out in preprod.

# Trello
[Card](https://trello.com/c/dUS1ubPX)


_infra/helm/frontstage/values.yaml
```diff
@@ -24,6 +24,7 @@ ingress:
   surveysHost: surveys.example.com
   certNameSurveys: surveys-cert
   timeoutSec: 30
+  sslPolicy: frontstage-ingress-ssl-policy
```
arroyoAle, line 27: Do we want to set this in the config? Or was this enough as they're all going to have the same name?

anwilkie: This policy will have the same name in all environments (including prod and preprod). Dev doesn't have an SSL policy (since there's no ingress set up), and even if it did, the SSL policy wouldn't be namespace specific since there'd be only one policy set up for the ras-rm-dev environment as a whole.

arroyoAle (approved):

PR #848, opened 2022-07-28
Improve error handling around account creation

insacuri:
# What and why?
The errors around account creation failures weren't easy to diagnose, so this PR adds a little more logging and more accurate status codes to help figure out future issues.

# How to test?

- Deploy this and the party PR with the same branch name (https://github.com/ONSdigital/ras-party/pull/371) to your environment. 
- Run the acceptance tests to seed data
- Try to create an account that already exists ([example@example.com](mailto:example@example.com)) and look at both the frontstage and party logs. The logging should make it more clear what is going on.
- Try and create an account to make sure nothing else has broken!

# Trello

### DIFF
--- a/frontstage/views/register/enter_account_details.py
+++ b/frontstage/views/register/enter_account_details.py
@@ -33,7 +33,6 @@ def register_enter_your_details():
 
     if request.method == "POST" and form.validate():
         email_address = form.email_address.data
-        logger.info("Attempting to create account", email=obfuscate_email(email_address))
 
         registration_data = {
             "emailAddress": email_address,
@@ -60,14 +59,12 @@ def register_enter_your_details():
                 flash("Something went wrong, please try again or contact us", "error")
                 return render_template("register/register.enter-your-details.html", form=form, errors=form.errors)
             elif exc.status_code == 409:
-                logger.info("Email already used", email=obfuscate_email(email_address), enrolment_code=enrolment_code)
                 error = {"email_address": ["This email has already been used to register an account"]}
                 return render_template("register/register.enter-your-details.html", form=form, errors=error)
             else:
                 logger.error("Failed to create account", status=exc.status_code, error=exc.message)
                 raise exc
 
-        logger.info("Successfully created account", email=obfuscate_email(email_address))
         return render_template("register/register.almost-done.html", email=email_address)
 
     else:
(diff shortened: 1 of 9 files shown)


frontstage/views/register/enter_account_details.py
```diff
@@ -31,8 +32,9 @@ def register_enter_your_details():
     iac_controller.validate_enrolment_code(enrolment_code)
 
     if request.method == "POST" and form.validate():
-        logger.info("Attempting to create account")
         email_address = form.email_address.data
+        logger.info("Attempting to create account", email=obfuscate_email(email_address))
```
arroyoAle, line 36: Do we need this log line to happen twice? One with the iac and one without? (Line 77 of party_controller.py)

frontstage/views/register/enter_account_details.py
```diff
@@ -46,14 +48,26 @@ def register_enter_your_details():
...
+                logger.info(
+                    "Party returned an error",
+                    email=obfuscate_email(email_address),
+                    enrolment_code=enrolment_code,
+                    error=exc.message,
+                )
+                flash("Something went wrong, please try again or contact us", "error")
+                return render_template("register/register.enter-your-details.html", form=form, errors=form.errors)
+            elif exc.status_code == 409:
+                logger.info("Email already used", email=obfuscate_email(email_address), enrolment_code=enrolment_code)
```
arroyoAle, line 63: This is really similar to the log on 87 of party controller

frontstage/views/register/enter_account_details.py
```diff
@@ -46,14 +48,26 @@ def register_enter_your_details():
...
                 error = {"email_address": ["This email has already been used to register an account"]}
                 return render_template("register/register.enter-your-details.html", form=form, errors=error)
             else:
-                logger.error("Failed to create account", status=exc.status_code)
+                logger.error("Failed to create account", status=exc.status_code, error=exc.message)
                 raise exc
 
-        logger.info("Successfully created account")
+        logger.info("Successfully created account", email=obfuscate_email(email_address))
```
arroyoAle, line 70: This is also very similar, again with/without iac

arroyoAle (changes requested):

frontstage/views/register/enter_account_details.py
insacuri: Good point, removed

frontstage/views/register/enter_account_details.py
insacuri: Good point, removed

frontstage/views/register/enter_account_details.py
insacuri: Good point, removed

PR #852, opened 2022-08-05
re-added frontendconfig

anwilkie:
# What and why?
Upon doing further research into this SSL policy problem, I realised that the issue wasn't with my config for frontstage, but with the spinnaker service account permissions. As such, I have re-added the frontendconfig resource from my previous PR with the same config.

# Trello
[Card](https://trello.com/c/dUS1ubPX)


SteveScorfield (approved):
Looks fine

PR #853, opened 2022-08-08
Expired email tokens will cause 500 on password reminder

SteveScorfield:
# What and why?
When a password verification token was out of date, it is deleted from the DB but the password reset counter is not reset to 0 (it will always be greater than 0). When the user tries to reset their password, the code evaluates an 'if' statement that checks if the reset count is greater than 0 (which it will be) it will then execute the try block within that searches for the deleted token within a dictionary. As it is has been deleted, it is not in the dictionary and so an error is thrown. 

To fix the issue, I have added an additional check to see if token is included as part of the dictionary. If it is not, then a new token will be generated (the reset count will not be affected). The user will then be able to receive the password reset email or will receive the 'tried too many times' message.

# How to test?
Find a user where the password verification token is out of date (so deleted) and the password reset count is greater than 0. Using the 'Forgot password' section, enter the email address of the account and click 'Send reset link'. This will then redirect to the 'Check your email section'.

# Trello
https://trello.com/c/PITGUDj6/1896-expired-email-tokens-will-cause-500-on-password-reminder

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -15,9 +15,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 
-version: 2.4.34
+version: 2.4.35
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
 
-appVersion: 2.4.34
+appVersion: 2.4.35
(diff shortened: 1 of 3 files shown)


SteveScorfield:
[Expired email tokens will cause 500 on password reminder](https://trello.com/c/PITGUDj6/1896-expired-email-tokens-will-cause-500-on-password-reminder)

JacobHarrisonONS (approved):

JacobHarrisonONS (approved):

PR #864, opened 2022-10-04
Fix package vulnerabilities, remove unused templates and rename others

insacuri:
# What and why?

This PR does 3 things.
- Updates packages to overcome security vulnerabilities
- Renames a number of templates that had an odd naming convention.  A few of them were called `register.<other stuff>.html`.  Normally we only have 1 dot in a file for the extension so this puts it back.
- Removes a few templates that weren't used

# How to test?

- Deploy app to environment
- Run acceptance tests
- Log into an existing user and add a new survey (acceptance test would've made some codes in rops for you to use)
- Log out and create a new user with another code.  It should work just fine.

# Trello


anwilkie (approved):
Tested with the instructions in the description and enrolling users worked okay.

PR #865, opened 2022-10-06
Testing djlint

matthew-robinson-ons:
# What and why?
Testing Djlinter on html files 
# How to test?
# Trello


matthew-robinson-ons:
[Test adding html template linter to ras-frontstage](https://trello.com/c/d1zillBY/2126-test-adding-html-template-linter-to-ras-frontstage)

frontstage/templates/sign-in/sign-in.account-not-verified.html
```diff
@@ -8,7 +8,7 @@ <h1 class="ons-u-fs-xl">Verify your email</h1>
 
     <p>When you created your account we sent you an email to verify your email address.</p>
     <p>Please follow the link the email to confirm your email address and sign in.</p>
-    <br>
+    <br />
```
insacuri, line 11: Need to remove the extra space (for this and all the others that have random extra space before the /).  Should be <br/>

Makefile
```diff
@@ -23,12 +23,14 @@ lint:
...
 	pipenv run isort .
 	pipenv run black --line-length 120 .
+	pipenv run djlint .
 	pipenv run flake8
 
 lint-check: load-design-system-templates
 	pipenv check
 	pipenv run isort . --check-only
 	pipenv run black --line-length 120 --check .
+	pipenv run djlint .
```
insacuri, line 33: I think you need to add the `--check` flag here right? so that it's `pipenv run djlint . --check`.  Otherwise I think it'll format it for you and then carry on, instead of failing the build which is what we want

insacuri (changes requested):
A few small bits but looks good

Makefile
matthew-robinson-ons: `--check` seems to be used as the formatter in DJlint
Screen Shot 2022-10-11 at 15 05 27 pm
When run the build fails as the code needs to be reformatted to match the suggested output (which does not look good). 
When run without `--check` it only checks the linting errors and fails if those are present. Is the formatter required?

insacuri: yeah sorry ^^ I missed that check was only for the formatter.  Also linters by definition check themselves >_<

frontstage/templates/secure-messages/conversation-view.html
```diff
@@ -15,7 +15,7 @@
         {% call onsPanel({
             "classes": "ons-u-mb-s"
         }) %}
-            <p class="ons-u-fs-r ons-u-mb-xs">This conversation has now been closed. <br>
+<p class="ons-u-fs-r ons-u-mb-xs">This conversation has now been closed. </p>
```
insacuri, line 18: Umm, this one might be wrong still?  Adding the closing p is good, but you removed the <br>.
Also this needs to be indented back to how it was as it looks weird with it being all the way to the left

insacuri (approved):

PR #874, opened 2022-11-08
Remove pipenv check

LJBabbage:
# What and why?
Currently we have 2 pipenv checks that are failing the build, one concerns py and the other one is a temporary fix which ignores 51499: wheel
# How to test?
Test everything still works
# Trello

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,12 +20,14 @@ docker-test: REDIS_PORT=6379
 docker-test: test
 
 lint:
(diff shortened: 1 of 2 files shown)


pricem14pc (approved):
Rule being ignored as per the [output here](https://github.com/ONSdigital/ras-frontstage/actions/runs/3421646174/jobs/5698070087#step:7:386)

PR #881, opened 2022-12-13
Remove eq version code

insacuri:
# What and why?

Now that no exercises are using EQ v2 anymore, we can start removing all mentions of the old platform.  This PR ignores the version of EQ that the collection exercise states (which should be v3 anyway, but it's in theory still possible for v2 ones to be made) and just creates a v3 payload and goes to v3 regardless.

# How to test?

- Create exercise (manually or via acceptance tests)
- Set exercise to v2 (via UI or database)
- Have it go live and enrol for it
- Launch it, it should take you to the v3 endpoint in mock-eq.  You can also double check here that the eqVersion for the exercise is v2 to make sure.

# Trello

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -34,11 +34,11 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:0d9c601124e5a6ba9712dbc60d9c53c21e34f5f641fe83002317394311bdce14",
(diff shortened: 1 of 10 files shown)


pricem14pc (approved):
All v3 and v2 EQs are correctly resolving to the`EQ_V3_URL` and launching the Mock EQ v3.
EQ Launch tests passing
Unrelated updates to tests behaving as expected and passing

PR #888, opened 2023-02-13
Mock redis in frontstage tests

matthew-robinson-ons:
# What and why?
The unit tests in frontstage were only able to be run with an instance of redis also running, otherwise multiple tests would fail. This Pr addresses that by introducing a redis mock (fakeredis) to be used in testing. 
# How to test?
Run frontstage unit tests with and without a running instance of redis, all tests should pass both times. 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-639

### DIFF
--- a/frontstage/__init__.py
+++ b/frontstage/__init__.py
@@ -1,3 +1,4 @@
+import fakeredis as fakeredis
 import redis
 from flask import render_template
 from flask_talisman import Talisman
@@ -48,6 +49,10 @@ talisman = Talisman(
     frame_options="DENY",
 )
 redis = redis.Redis(host=app.config["REDIS_HOST"], port=app.config["REDIS_PORT"], db=app.config["REDIS_DB"])
+if app.config["TESTING"]:
+    redis = fakeredis.FakeRedis(
+        host=app.config["REDIS_HOST"], port=app.config["FAKE_REDIS_PORT"], db=app.config["REDIS_DB"]
+    )
 
 app.jinja_env.undefined = ChainableUndefined
 app.jinja_env.add_extension("jinja2.ext.do")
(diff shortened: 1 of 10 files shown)


frontstage/__init__.py
```diff
@@ -1,3 +1,4 @@
+import fakeredis as fakeredis
```
insacuri, line 1: Why the `as fakeredis` at the end?

matthew-robinson-ons: Was autocompleted when i added fakeredis will remove it before merging.

frontstage/__init__.py
```diff
@@ -48,6 +49,10 @@
     frame_options="DENY",
 )
 redis = redis.Redis(host=app.config["REDIS_HOST"], port=app.config["REDIS_PORT"], db=app.config["REDIS_DB"])
+if app.config["TESTING"]:
```
insacuri, line 52: This should be written as:
```python
if app.config["TESTING"]:
    <make a fake redis>
else:
    <make a real one>
```

Also, can you put a comment in each path to say `logger.info('Using real redis')` and `logger.info('Using fake redis')` so there's less ways for a mistake to happen

insacuri (changes requested):

insacuri:
Tested it both locally and in k8s and it seems happy.  Ran the acceptance tests and it's good too.
Adding the log line to say if it's using the real or fake one inspires waaay more confidence that a mistake isn't being made.  Also, the fact that the `TESTING` config item is hardcoded as False everywhere except the TestingConfig means it should be incredibly obvious if the wrong config is used.

insacuri (approved):

PR #893, opened 2023-04-21
Add session expire-at endpoint

LJBabbage:
# What and why?
This PR adds the expire-at end point to allow the timeout modal to work. This is more a stepping stone PR as timeout is in 2 stories and without delivering both but knowing what the 2nd part needs it has been developed with that in mind
# How to test?
Deploy this PR to your environment and then hit the session/expires-at endpoint (N.B the time should be displayed as UTC) It shouldn't change if you refresh the page, but as you wander around the rest of frontstage it will
# Trello


frontstage/views/session.py
```diff
@@ -0,0 +1,14 @@
...
+
+from frontstage.common.authorisation import jwt_authorization
+
+session_bp = Blueprint("session", __name__)
+
+
+@session_bp.route("/expires-at", methods=["GET"])
+@jwt_authorization(request, refresh_session=False)
+def session_expires_at(session):
+    expires_at = session.get_expires_in()
```
LJBabbage, line 13: This is could do with fixing at some stage, expires_in is not right, however I can't just change this due to the affect it would have on live jwt session on a deploy to PROD

frontstage/common/authorisation.py
```diff
@@ -33,7 +33,7 @@ def validate(token):
     return True
 
 
-def jwt_authorization(request):
+def jwt_authorization(request, refresh_session=True):
```
LJBabbage, line 36: A bit of inside trading here, I know from past experience that the timeout modal does not want expire-at to update the session when hit

frontstage/views/session.py
```diff
@@ -0,0 +1,14 @@
...
+    return jsonify(expires_at=datetime.fromtimestamp(expires_at, tz=timezone.utc).isoformat())
```
LJBabbage, line 14: We should be dealing with UTC

PR #896, opened 2023-05-05
RAS-739: Add modal to base template

SteveScorfield:
# What and why?
There is a requirement that people logged into frontstage need to be notified that they will be logged out 60 seconds before their session times out. They can either stay on the page and their timeout counter will be refreshed or they will be redirected to the login screen.

These code changes will achieve this.

# How to test?
1. Sign into frontstage with a user account
2. Once redirected to the todo survey list keep the page loaded and do not navigate anywhere whilst logged in with the above user account
3. After 59 minutes have elapsed, a message will appear on the screen with a counter
4. Test both the following:
- Click the 'Continue survey' button and you should be on the same screen that you you originally opened
-  Click the 'Continue survey' button once the timer has counted down. This will redirect you to the login page

# Jira
https://jira.ons.gov.uk/browse/RAS-739

### DIFF
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -80,8 +80,10 @@ class Session(object):
         redis.setex(self.session_key, 3600, self.encoded_jwt_token)
         self.persisted = True
 
-    def get_formatted_expires_in(self):
-        return datetime.fromtimestamp(self.get_expires_in(), tz=timezone.utc).isoformat()
+    def get_formatted_expires_in(self, refresh=None):
+        if refresh:
+            self.refresh_session()
+        return datetime.fromtimestamp((self.get_expires_in() - 3400), tz=timezone.utc).isoformat()
 
 
 def _get_new_timestamp(ttl=3600):
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -81,7 +81,7 @@ class Session(object):
         self.persisted = True
 
     def get_formatted_expires_in(self):
-        return datetime.fromtimestamp((self.get_expires_in() - 3500), tz=timezone.utc).isoformat()
+        return datetime.fromtimestamp(self.get_expires_in(), tz=timezone.utc).isoformat()
 
 
 def _get_new_timestamp(ttl=3600):
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -80,9 +80,7 @@ class Session(object):
         redis.setex(self.session_key, 3600, self.encoded_jwt_token)
         self.persisted = True
 
-    def get_formatted_expires_in(self, refresh=None):
-        if refresh:
-            self.refresh_session()
+    def get_formatted_expires_in(self):
         return datetime.fromtimestamp(self.get_expires_in(), tz=timezone.utc).isoformat()
 
 
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -83,7 +83,7 @@ class Session(object):
     def get_formatted_expires_in(self, refresh=None):
         if refresh:
             self.refresh_session()
-        return datetime.fromtimestamp((self.get_expires_in() - 3400), tz=timezone.utc).isoformat()
+        return datetime.fromtimestamp(self.get_expires_in(), tz=timezone.utc).isoformat()
 
 
 def _get_new_timestamp(ttl=3600):
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -81,7 +81,7 @@ class Session(object):
         self.persisted = True
 
     def get_formatted_expires_in(self):
-        return datetime.fromtimestamp(self.get_expires_in(), tz=timezone.utc).isoformat()
+        return datetime.fromtimestamp((self.get_expires_in() - 3500), tz=timezone.utc).isoformat()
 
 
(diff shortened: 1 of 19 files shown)


frontstage/views/surveys/add_survey.py
```diff
@@ -16,23 +16,41 @@
...
         logger.info("Enrolment code submitted when attempting to add survey", enrolment_code=enrolment_code)
         # Validate the enrolment code
         try:
             iac = iac_controller.get_iac_from_enrolment(enrolment_code)
             if iac is None:
-                logger.info("Enrolment code not found when attempting to add survey", enrolment_code=enrolment_code)
+                logger.info(
+                    "Enrolment code not found when attempting to add survey",
+                    enrolment_code=enrolment_code,
+                    expires_at=expires_at,
```
LJBabbage, line 34: Not sure how useful adding the expires_at in the logs would be, by virtue of the jwt_authorization decorator it will always be valid

LJBabbage (changes requested):
Unfortunately I don't think this will work as desired. On my tests the patch request will not extend the session, and as such mean the respondent will still have to do something within the timeout. (i.e if they press the button on the pop-up with 10 seconds to go, they will then have to do something on the screen behind inside 10 seconds) It's worth checking as this might be correct, but differs from the  implementation in eQ which does extend it. 

Unless I missed it, we also need tests to check patch behaviour.

frontstage/views/surveys/add_survey.py
SteveScorfield: This has been removed. Not sure why I put these in!

tests/integration/views/test_session.py
```diff
@@ -52,3 +52,24 @@ def test_expires_at_future_time(self, mock_redis):
         # Then the expiration time is in 40 minutes as calling expires-at doesn't extend the session
         self.assertEqual(expected_expires_at, parsed_json["expires_at"])
         self.assertEqual((future_time + timedelta(minutes=40)).isoformat(), expected_expires_at)
+
+    @freeze_time(TIME_TO_FREEZE)
+    @patch("redis.StrictRedis.get")
+    def test_patch_expires_at_current_time(self, mock_redis):
```
LJBabbage, line 58: This seems a bit complicated, I think all we need to do is check that the patch extends the timeout. Set a time using freeze, walk it forward 20 minutes (or whatever) call patch and make sure the session timeout is 60 minutes further along.

frontstage/views/session.py
```diff
@@ -1,14 +1,31 @@
...
+from flask import Blueprint, jsonify, make_response, redirect, request, session, url_for
 
 from frontstage.common.authorisation import jwt_authorization
+from frontstage.common.session import Session
 
 session_bp = Blueprint("session", __name__)
 
 
-@session_bp.route("/expires-at", methods=["GET"])
+@session_bp.route("/expires-at", methods=["GET", "PATCH"])
```
LJBabbage, line 9: In this instance I think it is easier to split the methods out i.e have one for PATCH and one for GET. For the one that is looking after patch just use @jwt_authorization(request) rather than 
@jwt_authorization(request, refresh_session=False) and it will do the session extension for you

frontstage/common/session.py
```diff
@@ -80,6 +80,11 @@ def save(self):
         redis.setex(self.session_key, 3600, self.encoded_jwt_token)
         self.persisted = True
 
+    def get_formatted_expires_in(self, refresh=None):
+        if refresh:
+            self.refresh_session()
```
LJBabbage, line 85: We shouldn't need to do this, @jwt_authorization(request) extends the session https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/common/authorisation.py#L62 so the views you have updated below will actually do it twice. As mentioned in another comment if you had 2 methods, one for GET and one for PATCH which uses the 2 refresh states in jwt_authorization, (default and false) everything should work without doing much else

frontstage/views/session.py
```diff
@@ -1,14 +1,31 @@
...
-    return jsonify(expires_at=datetime.fromtimestamp(expires_at, tz=timezone.utc).isoformat())
+def session_expires_at(expires_at):
+    refresh = None
+    if request.method == "PATCH":
+        refresh = True
+    return jsonify(expires_at=expires_at.get_formatted_expires_in(refresh))
+
+
+@session_bp.route("/session-expired", methods=["GET"])
+def session_expired():
```
LJBabbage, line 19: Isn't this the same as a logout and as such should be delegated? Unless I missed it there should also be a test

LJBabbage (changes requested):

frontstage/views/session.py
SteveScorfield: As above

tests/integration/views/test_session.py
SteveScorfield: I've made it simpler now.

frontstage/views/session.py
SteveScorfield: I've updated the code based on the recommendations.

frontstage/views/session.py
SteveScorfield: This has been removed and the logout endpoint is now used.

LJBabbage (approved):

PR #899, opened 2023-05-19
RAS-765 Update eq payload to V2

LJBabbage:
# What and why?
This PR updates the payload sent to eQ to v2. It also adds much needed payload tests, which were lacking and should really have been there before
# How to test?
The best way is to expose the payload and then validate it against https://github.com/ONSdigital/ons-schema-definitions#json-schema-validation. 

N.B There is some strange config differences when using response_expires with different timezones being used. I will raise another card for this, but considering we didn't test anything before, this is a large step in the right direction

# Trello


JacobHarrisonONS (approved):
Validated payload and it was all good. tested completing EQ in dev and worked fine but felt a little slow but assuming dev is having a bad day.

PR #903, opened 2023-06-01
update pipenv

JacobHarrisonONS:
# What and why?
Removes security vulnerabilities by updating pipenv packages. Also introduces two ignore codes for djlint which preventing linting errors introduced after it was updated which will need to be addressed in another card. 
# How to test?
Run the tests and check the add survey function works correctly as the redirect had to be updated.
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-783

### DIFF
--- a/frontstage/views/surveys/add_survey_submit.py
+++ b/frontstage/views/surveys/add_survey_submit.py
@@ -89,13 +89,16 @@ def add_survey_submit(session):
     return redirect(
         url_for(
             "surveys_bp.get_survey_list",
-            _anchor=(business_party_id, added_survey_id),
             _external=True,
             business_party_id=business_party_id,
             survey_id=added_survey_id,
             tag="todo",
             already_enrolled=already_enrolled,
         )
+        + "#"
+        + business_party_id
+        + "#"
+        + added_survey_id
     )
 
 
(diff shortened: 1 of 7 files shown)


frontstage/views/surveys/add_survey_submit.py
```diff
@@ -89,13 +89,16 @@ def add_survey_submit(session):
...
         url_for(
             "surveys_bp.get_survey_list",
-            _anchor=(business_party_id, added_survey_id),
             _external=True,
             business_party_id=business_party_id,
             survey_id=added_survey_id,
             tag="todo",
             already_enrolled=already_enrolled,
         )
+        + "#"
```
LJBabbage, line 98: This doesn't make any sense to me, those anchors don't appear on the page and there are 2 which won't work? The only anchor that is relevant is the **new-survey** one and that naturally doesn't need business or add survey id. I think we either remove them or use new-survey and probably f string it to sort out the weird format

LJBabbage (changes requested):

frontstage/views/surveys/add_survey_submit.py
matthew-robinson-ons: Have looked into if the anchor is required and doesn't look like it is used currently. Have removed the code from this PR.

LJBabbage (approved):

PR #906, opened 2023-06-29
update design system version

anwilkie:
# What and why?
Updating design system version to 62.2.4.

# How to test?
Deploy to environment, run acceptances tests, and navigate around frontstage to make sure everything looks okay.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-816)


LJBabbage (changes requested):
Please see this card https://trello.com/c/nS3unaCw/2638-fix-breaking-changes-for-ds-6300. The change log in DS might need a 2nd look as well https://github.com/ONSdigital/design-system/releases . This one https://github.com/ONSdigital/design-system/pull/2682 with the aria-label we do use https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/templates/surveys/surveys.html#L51 There might be more, with the tech card and this I stopped reviewing

LJBabbage:
As we have a card to address 63.0.0, this card should only be up to [62.2.4](https://github.com/ONSdigital/design-system/releases/tag/62.2.4)

matthew-robinson-ons (approved):
Checked breaking changes and breadcrumbs all looked good. Rest of frontage working as expected.

PR #907, opened 2023-06-29
RAS-796 Rewrite jwt_authorization in frontstage

LJBabbage:
# What and why?
This PR updates the jwt_authorization decorator, the code has been tighten up and logging improve, coverage is now 100% and the decorator should now be more efficient. I have removed the optional validation as this makes no sense. I have removed redundancy in tests and updated the JWTs

# How to test?
The tests are a good pattern to match, login normally, then try without an auth cookie, try with a expired jwt etc. A general smoke test of major routes isn't a bad idea either, although our tests do offer coverage
# Jira


config.py
```diff
@@ -21,7 +21,6 @@ class Config(object):
     SECURITY_USER_PASSWORD = os.getenv("SECURITY_USER_PASSWORD")
     BASIC_AUTH = (SECURITY_USER_NAME, SECURITY_USER_PASSWORD)
     JWT_SECRET = os.getenv("JWT_SECRET")
-    VALIDATE_JWT = os.environ.get("VALIDATE_JWT", True)
```
LJBabbage, line 24: I can't see a valid reason why we ever not want to this this. Firstly VALIDATE_JWT is now a bad name for this as all this is actually doing is checking the expire_in date in the payload. However there is no situation we don't want to do this, in fact it is fairly crucial we do. If we accidentally changed how expire_in change, we want to know about it, in all environments and tests

frontstage/common/authorisation.py
```diff
@@ -2,69 +2,59 @@
...
 
-
-def validate(token):
-    logger.debug("Validating token")
-
-    now = datetime.now().timestamp()
-    expires_at = token.get("expires_in")
-    if expires_at:
-        if now >= expires_at:
-            flash("To help protect your information we have signed you out.", "info")
```
LJBabbage, line 25: This shouldn't be in a decorator, moved to the right place

frontstage/common/authorisation.py
```diff
@@ -2,69 +2,59 @@
...
-                    jwt = decode(encoded_jwt, jwt_secret, algorithms="HS256")
-                    logger.debug("Token decoded successfully")
-                except JWTError:
-                    logger.warning("Unable to decode token")
-                    raise JWTValidationError
-            else:
-                logger.warning("No authorization token provided")
-                flash("To help protect your information we have signed you out.", "info")
-                session["next"] = request.url
-                return redirect(url_for("sign_in_bp.login"))
```
LJBabbage, line 57: redirect here is also not quite right

frontstage/error_handlers.py
```diff
@@ -65,8 +74,8 @@ def connection_error(error):
 
 @app.errorhandler(JWTValidationError)
 def jwt_validation_error(error):
-    logger.error("JWT validation error", url=request.url, status_code=403)
-    return render_template("errors/403-error.html"), 403
+    logger.error(error=error.message, status_code=500)
```
LJBabbage, line 77: This shouldn't be 403, we generate that JWT and hold it, if it is wrong we have a problem in how we generate it. With that in mind no amount of retries will fix this and this should be a 500

tests/integration/test_jwt_authorization.py
```diff
@@ -1,87 +1,121 @@
...
 from frontstage.common.session import Session
 from frontstage.exceptions.exceptions import JWTValidationError
 
 valid_jwt = (
-    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXJ0eV9pZCI6ImY5NTZlOGFlLTZ"
-    "lMGYtNDQxNC1iMGNmLWEwN2MxYWEzZTM3YiIsImV4cGlyZXNfYXQiOiIxMDAxMjM0NTY"
-    "3ODkiLCJyb2xlIjoicmVzcG9uZGVudCIsInVucmVhZF9tZXNzYWdlX2NvdW50Ijp7InZh"
-    "bHVlIjowLCJyZWZyZXNoX2luIjozMjUyNzY3NDAwMC4wfSwiZXhwaXJlc19pbiI6MzI1M"
-    "jc2NzQwMDAuMH0.m94R50EPIKTJmE6gf6PvCmCq8ZpYwwV8PHSqsJh5fnI"
+    "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJwYXJ0eV9pZCI6InRlc3QiLCJyb2xlIjoicmVzcG9uZGVudCIsInVucmVhZF9"
```
LJBabbage, line 19: This had only payloads, updated

tests/integration/test_jwt_authorization.py
```diff
@@ -1,87 +1,121 @@
...
         @jwt_authorization(request)
-        def test_function(session):
+        def test_function(_):
             pass
 
         test_function()
 
     def test_jwt_authorization_success(self):
-        self.session.encoded_jwt_token = valid_jwt
-        self.session.session_key = str(uuid4())
```
LJBabbage, line 53: Not needed

tests/integration/test_jwt_authorization.py
```diff
@@ -1,87 +1,121 @@
...
 )
 
 
 class TestJWTAuthorization(unittest.TestCase):
     def setUp(self):
         self.app = app.test_client()
-        self.app.testing = True
-        self.session = Session.from_party_id("test")
-
-    def tearDown(self):
```
LJBabbage, line 40: Not needed

SteveScorfield (approved):
Looks good to me

PR #908, opened 2023-07-04
RAS-797 Update authorization cookie to session

LJBabbage:
# What and why?
This PR changes the authorization cookie to session specific. How we are currently using the cookie means that a respondent would be logged out after 60 minutes regardless of the expire value in the JWT payload. This is incorrect as it should be using that value in the JWT, we have been changing that value to extend a timeout but because of the hard coded cookie it hasn't been working.

N.B. I did all the heavy lifting, separating out authorization cookie from the JWT payload check and all the required tests in this PR https://github.com/ONSdigital/ras-frontstage/pull/907 so this is simply changing the cookie.

# How to test?
Check the cookie is now session specific and the timeout is being controlled by the JWT expire_in value. The easiest way to do this if you don't want to wait hours is change the timeout here https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/common/session.py#L87 . I changed it to 10 seconds then logged in, click around for 20 seconds and made sure it didn't log me out, then left it for 10+ seconds and make sure it did

image

# Jira


SteveScorfield (approved):
Looks good to me

PR #911, opened 2023-07-14
RAS-739 Add modal to base template

LJBabbage:
# What and why?
This PR adds the timeout modal.
# How to test?
It's best to test this locally, for 2 reasons, csrf is turned off in dev and the timeout is set to 60 minutes, turning on csrf in dev isn't an issue, but the timeout is hardcoded which is a pain.

Now that you have it running locally, go into the code and change this to something sensible https://github.com/ONSdigital/ras-frontstage/blob/main/frontstage/common/session.py#L87 like 65, this means the modal will pop up after 5 seconds making it a lot easier to test. Basically you now what to interact with the app and the modal, recommended tests

1. Navigated around the app clicking with-inside the 5 seconds(if you set it to 65), make sure the timeout doesn't appear
2. Let the modal pop up then press the continue and make sure the session is extended (http://localhost:8082/session/expires-at is useful), repeat however many times you like
3. Let the modal run down and check it logs you out
4. Check the modal works without focus, i.e log in to frontstage then go to another page and make sure it is logged out after you come back after 70 seconds
5. Anything else you can think off

# Jira

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.68
+version: 2.4.69
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.4.68
+appVersion: 2.4.69
--- a/frontstage/common/authorisation.py
+++ b/frontstage/common/authorisation.py
@@ -9,7 +9,7 @@ from werkzeug.exceptions import Unauthorized
 
 from frontstage import app
 from frontstage.common.session import Session
-from frontstage.exceptions.exceptions import JWTValidationError
+from frontstage.exceptions.exceptions import JWTTimeoutError, JWTValidationError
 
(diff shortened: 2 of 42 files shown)


anwilkie (approved):
Tested locally by altering the value of the timeout. It works as expected, although there is an interesting piece of behaviour where typing in a text box (such as a secure message box) doesn't constitute an action that resets the modal timeout, so the user can be interrupted by the pop-up in the middle of typing a message. Given how long it would take the modal to timeout naturally, it's unlikely to be much of an issue, so I don't see a need/reason to fix it.

LJBabbage:
> Tested locally by altering the value of the timeout. It works as expected, although there is an interesting piece of behaviour where typing in a text box (such as a secure message box) doesn't constitute an action that resets the modal timeout, so the user can be interrupted by the pop-up in the middle of typing a message. Given how long it would take the modal to timeout naturally, it's unlikely to be much of an issue, so I don't see a need/reason to fix it.

The session gets extended on the exercising of the jwt_authorization decorator, that only happens when a request (POST, GET) is made, because of that typing in a box won't extend it. In some ways this will actually help stop someone losing a secure message they have spent an hour typing, as instead of waiting till they press the submit and then giving them a timeout, they would get a warning first, which allows them to extend

PR #915, opened 2023-08-01
Remove timeout modal

LJBabbage:
# What and why?
It looks like there might be a race condition (it could be something else, I haven't fully debugged) with the modal timeout and the authorization cookie, the modal is calling to the expire-at when created and finding there isn't a cookie sometimes and raising a 401. This is not a fix, purely a removal at present, another PR will follow this
# How to test?
Just make sure you can login and navigate around and the modal code has gone from the html
# Jira


anwilkie (approved):
Ran acceptance tests, and they passed. Went on frontstage, navigated around okay, and made sure that the html code that had been removed wasn't there anymore.

PR #919, opened 2023-08-07
RAS-857 Remove GA4 <noscript>, change analytics cookie key

LJBabbage:
# What and why?
Remove non javascript code and update 
# How to test?
This isn't really possible to test in dev, bar checking that the non javascript code doesn't appear anymore in the html. The reason you can't really test any further is because we don't have a tag manager id for dev only preprod

N.B you will need to give the env GOOGLE_TAG_MANAGER_ID a value in config or export it otherwise nothing will appear
# Jira


SteveScorfield (approved):
Looks good

PR #922, opened 2023-08-23
Fix password reset with email change

arroyoAle:
# What and why?
This fixes the case where a user changes email while having requested a forgot password link by deleting the password reset token in the database when the party email is changed
# How to test?
- deploy to environment
- follow the `forgot password` path for an account
- check the token is in the database
- change the user email
- check the token has been removed from the database
# Jira
https://jira.ons.gov.uk/browse/RAS-840

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -16,14 +16,6 @@
         ]
     },
     "default": {
-        "async-timeout": {
-            "hashes": [
-                "sha256:4640d96be84d82d02ed59ea2b7105a0f7b33abe8703703cd0ab0bf87c427522f",
-                "sha256:7405140ff1230c310e51dc27b3145b9092d659ce68ff733fb0cefe3ee42be028"
-            ],
-            "markers": "python_full_version <= '3.11.2'",
-            "version": "==4.0.3"
-        },
         "blinker": {
             "hashes": [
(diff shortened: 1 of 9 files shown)


LJBabbage (changes requested):
I noticed this in draft this morning so circled back around when it was approved as it really shouldn't be approved in that state and also realised that the code needs more work and the orchestration is in the wrong place. Although I wouldn't usually review in draft as it might not be finished, because it now has an approval and  the fact it should be in another service I am inclined to comment

Orchestration shouldn't be in frontstage it should be in party as that is where the data is changed, and we shouldn't be jumping back and forward from another service (and with mocks in tests) to do that. 

The request from the respondent is also not updating the email at the point the code has been added, it just creates a value in the pending_email_address column, and only migrates to the email on the hitting of another endpoint which the user gets from the email. As such the token is actually still valid for a password change until they do that. Wiping it on the original requests for change is not quite right and could lead to strange behaviour if the password request email is click. 

We need to think about and cater for both emails clicked in either order (email change and password change) which can only be done in party when the actual data is updated. I also noticed with this that the email reset count is still 1 after an email change, that feels strange and potentially problematic without a token being there. 

We should also be explicit with new tests, this was raised as a bug and because of that tests should focus on those scenarios. Updating current tests so they don't break does not really reflect the user journey that caused this.

arroyoAle:
Changes have been moved to the party service as that is where they should be handled

PR #926, opened 2023-09-05
fix ons-icon in html files

anwilkie:
# What and why?
I missed something in the last design system upgrade where the icons that appear near links (such as on the About the ONS info page) appeared enormous. This turned out to be because the design system v65 renamed the `ons-svg-icon` to `ons-icon` (this wasn't mentioned in the breaking changes!).

Also, the design system upgrade messed with the appearance of the tab component due to the fact that the way we use it is not in accordance with the design system's ethos. The way the tabs are currently created is by a hacky HYML code block, which isn't how it's meant to be done, but it allows us to have the appropriate functionality. To continue supporting our implementation of tabs, the .css file has had some additional config in order to maintain its current appearance. Also, make sure that the tab component (seen when looking at surveys to-do/history/messages) appears correctly, with a line going across the page underneath the tabs.

# How to test?
Deploy, run acceptance tests, go to frontstage and navigate to one of the help pages and make sure that the icons near links are not way larger than they should be.

# Jira
[Card](https://jira.ons.gov.uk/browse/RAS-863)

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.78
+version: 2.4.79
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.4.78
+appVersion: 2.4.79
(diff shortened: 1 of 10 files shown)


matthew-robinson-ons:
Icons are appearing as normal however I don't seem to be able to complete a mock EQ I receive `An error has occurred
Sorry, something has gone wrong.` on the page. The history tab also doesn't look quite right to me but i couldn't confirm that as needed the survey to successfully complete.

matthew-robinson-ons (approved):
Code works as expected. There appears to be a separate issue with mock eQ though it isn't affected by this code so not blocking this PR.

arroyoAle (approved):
Tabs and icons look right and nothing else looks out of place from navigating the site

PR #928, opened 2023-09-07
Fix password reset with email change

arroyoAle:
# What and why?
Adds error handling for when a password reset token has been generated but the email has been changed since (invalidating the token and deleting it from the party database) and the user tries to use the old reset token

# How to test?
- deploy to environment
- follow the forgot password path for an account
- change the user email and verify the change through the link provided
- try using the password reset link (should show the `Your link is invalid or has already been used` page)
# Jira
[RAS-840](https://jira.ons.gov.uk/browse/RAS-840)


tests/integration/test_passwords.py
```diff
@@ -216,6 +216,22 @@ def test_reset_password_post_success(self, mock_request):
         self.assertEqual(response.status_code, 200)
         self.assertTrue("Your password has been changed".encode() in response.data)
 
+    @requests_mock.mock()
+    def test_reset_password_respondent_not_found(self, mock_request):
+        mock_request.get(url_banner_api, status_code=404)
+        mock_request.put(url_password_change, status_code=200)
```
LJBabbage, line 222: Do we need to mock url_password_change? Might also look a bit neater if the mocks were all together

frontstage/views/passwords/reset_password.py
```diff
@@ -43,6 +43,9 @@ def get_reset_password(token, form_errors=None):
     except (BadSignature, BadData):
         logger.warning("Invalid token sent to frontstage password reset", token=token, exc_info=True)
         return render_template("passwords/password-expired.html", token=token)
+    except TypeError:
```
LJBabbage, line 46: We really shouldn't be doing this, this is catching a bug, rather than fixing it. The 404 should be raised and handled correctly

LJBabbage (changes requested):
I think we might need a pragmatic approach with this one. This isn't the way we should be handling the problem, however I can see get_respondent_by_email is used in multiple different places (also all wrong and causing other 500s) meaning it would expand this PR. I will raise a card to fix it properly. We  don't need to update the lock file so this should be removed from the update

tests/integration/test_passwords.py
arroyoAle: Removed and moved order of statements

LJBabbage (approved):

PR #929, opened 2023-09-07
RAS-872: Extend EQ launch claims to include sds_dataset_id if required for the form_type

SteveScorfield:
# What and why?
The SDS ID is required to be passed to eQ when an SDS is present and the form type(s) in question match that supplied by SDS. This PR achieves this by processing the the supplementary dataset within the CE, checking that the form type(s) matches the ones being processed as part of the CE and then passes the SDS ID to eQ.

# How to test?
1. Deploy this PR and the CE SVC PR (https://github.com/ONSdigital/rm-collection-exercise-service/pull/318)  to your local environment
2. Add a CE in rOps and then add an SDS that matches the CE - this can be done by either directly adding it to the DB or adding it to the pubsub subscription 'supplementary-data-service-subscription-[YOUR-ENV]'.
3. Grab the enrolment code for CE from rOps and link it to the correct account in Frontstage by using the 'add a new survey to your account'
4. Once the survey shows in the survey 'To do' list, click on it and it should take you to the eQ mock

# Jira
https://jira.ons.gov.uk/browse/RAS-872


frontstage/common/eq_payload.py
```diff
@@ -87,6 +88,11 @@ def create_payload(self, case, ce, party_id: str, business_party_id: str, survey
         if employment_date := self._find_event_date_by_tag("employment", ce_events, ce_id, False):
             payload["survey_metadata"]["data"]["employment_date"] = employment_date
 
+        if ce["supplementaryDatasetEntity"]:
+            supplementary_dataset = json.loads(ce["supplementaryDatasetEntity"]["supplementaryDatasetJson"])
+            if form_type in supplementary_dataset["form_types"]:
```
JacobHarrisonONS, line 93: version control this please to not be enabled in prod

JacobHarrisonONS (changes requested):

JacobHarrisonONS (approved):

frontstage/common/eq_payload.py
SteveScorfield: Spoke with Mark and he believes we should be fine as we've essentially feature flagged the queue that we pick up the SD from.

PR #938, opened 2023-10-10
update code to render GTM and GA snippet based on if its present in config

JacobHarrisonONS:
# What and why?
Due to the potential security risk, we should not currently include the gtm.html snippet in the page to avoid sending any GTM managed metrics to Google Tag Manager in prod and preprod.

# How to test?
This will need to be tested within preprod. 
After the deployment I will use Tag Assistant 
Tag Assistant lets you quickly see which of your site's trackers are working as expected. And if there are issues.

I expect tag assistant to inform me that it was unable to find any data due to the snippet not being present.

# Jira
https://jira.ons.gov.uk/browse/RAS-905

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.83
+version: 2.4.84
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
-appVersion: 2.4.83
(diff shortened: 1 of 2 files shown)


arroyoAle (approved):
Acceptance tests run and frontstage works, pages look the same

pricem14pc (approved):
Successfully tested in a local environment and behaving as expected

The inclusion of `GOOGLE_ANALYTICS_MEASUREMENT_ID` and `GOOGLE_TAG_MANAGER_ID` as environment variable independently includes the corresponding JS snippet in the rendered page.

Also including each key with a `null` value has the same effect as not including the key at all.

Including the key=value results in requests to the expected service. Otherwise the JS snippet is not rendered and the request is not made.

PR #942, opened 2023-11-08
RAS-800: Profile Python Frontstage app in GCP

SteveScorfield:
What and why?
We don't want to profile the Frontstage Python app in prod. This PR achieves this.

How to test?

Jira
https://jira.ons.gov.uk/browse/RAS-800


arroyoAle (changes requested):
Profiler and frontstage work, when I set the value to False manually, the profiler does not start. The only thing is the commented code

frontstage/create_app.py
```diff
@@ -39,6 +39,9 @@ def create_app_object():
     if not app.config["DEBUG"]:
         app.wsgi_app = GCPLoadBalancer(app.wsgi_app)
 
+    # if os.environ.get("APP_SETTINGS") == "TestingConfig":
+    #     app.config["PROFILER_ACTIVE"] = False
+
```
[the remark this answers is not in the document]
SteveScorfield: It should have been removed. Oversight by myself.

arroyoAle (approved):
Commented code removed and profiler works as expected

Pipfile.lock
SteveScorfield, line 1: This needed recreating to be up to date with latest pipfile. Will prevent the pipfile being out of date and having to be resolved by Git.

frontstage/views/sign_in/sign_in.py
```diff
@@ -38,7 +37,7 @@ def home():
 
 @sign_in_bp.route("/", methods=["GET", "POST"])
 def login():  # noqa: C901
-    if os.environ.get("APP_SETTINGS") != "TestingConfig":
+    if app.config["PROFILER_ACTIVE"]:
         try:
             googlecloudprofiler.start(verbose=3, service="frontstage")
```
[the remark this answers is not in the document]
SteveScorfield: You are absolutely correct. I have moved this to create_app.

_infra/helm/frontstage/values.yaml
```diff
@@ -96,4 +96,7 @@ gnu:
   fingerprint: ED1B7A3EADF95687
 
 password:
-  attemptTimeout: 86400
\ No newline at end of file
+  attemptTimeout: 86400
+
+frontstageProfiler:
+  enabled: true
```
pricem14pc, line 102: This should default to our production configuration of false. It will then be overridden as required.

pricem14pc (changes requested):

pricem14pc:
For consistency across all the dependent repos (inc. config) it would be preferable to call env var `PROFILER_ENABLED` rather than "ACTIVE"

_infra/helm/frontstage/templates/deployment.yaml
```diff
@@ -235,5 +235,7 @@ spec:
             value: "{{ .Values.gnu.recipient }}"
           - name: PASSWORD_RESET_ATTEMPTS_TIMEOUT
             value: "{{ .Values.password.attemptTimeout }}"
+          - name: PROFILER_ENABLED
+            value: "{{- if .Values.frontstageProfiler.enabled -}}True{{- else -}}False{{- end -}}"
```
pricem14pc, line 239: [frontstageProfiler:
  enabled: true](https://github.com/ONSdigital/ras-rm-config/pull/236#pullrequestreview-1739976749)

_infra/helm/frontstage/values.yaml
```diff
@@ -96,4 +96,7 @@ gnu:
   fingerprint: ED1B7A3EADF95687
 
 password:
-  attemptTimeout: 86400
\ No newline at end of file
+  attemptTimeout: 86400
+
+frontstageProfiler:
```
pricem14pc, line 101: As per [this]https://github.com/ONSdigital/ras-rm-config/pull/236#pullrequestreview-1739976749 comment. The context of this is frontstage so no need to prefix it.

pricem14pc (changes requested):

_infra/helm/frontstage/values.yaml
SteveScorfield: Makes sense. This is updated.

_infra/helm/frontstage/templates/deployment.yaml
SteveScorfield: Makes sense. This is updated.

config.py
```diff
@@ -117,6 +119,8 @@ class DevelopmentConfig(Config):
     WTF_CSRF_TIME_LIMIT = int(os.getenv("WTF_CSRF_TIME_LIMIT", "3200"))
     SECURE_APP = bool(strtobool(os.getenv("SECURE_APP", "False")))
 
+    PROFILER_ENABLED = True
```
pricem14pc, line 122: I think these should also default to False even in dev mode and be overridable
```
PROFILER_ENABLED = bool(strtobool(os.getenv("PROFILER_ENABLED", "False")))
```

config.py
```diff
@@ -140,3 +144,4 @@ class TestingConfig(DevelopmentConfig):
     SECURE_APP = bool(strtobool(os.getenv("SECURE_APP", "False")))
     ACCESS_CONTROL_ALLOW_ORIGIN = os.getenv("ACCESS_CONTROL_ALLOW_ORIGIN", "http://localhost")
     UNDER_MAINTENANCE = bool(strtobool(os.getenv("UNDER_MAINTENANCE", "False")))
+    PROFILER_ENABLED = False
```
pricem14pc, line 147: I think these should also default to False even in dev mode and be overridable
```
PROFILER_ENABLED = bool(strtobool(os.getenv("PROFILER_ENABLED", "False")))

pricem14pc (changes requested):

pricem14pc (approved):
Now defaults to not start the profiler in all envs/Flask configs, and overrides to enable the profiler with env config.

PR #943, opened 2023-11-13
Have updated the pipfile lock

SteveScorfield:
# What and why?
Regenerating the pipfile.lock was not carried out. This PR fixes this issue.

# How to test?

# Jira


pricem14pc (changes requested):
I think this is wrong as it _should_ still contain the cloud profiler that is on main. The Pipfile.lock should be the lock file from that build.

Pipfile.lock
```diff
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "48ab78ae907484c4229edf470a47699d6b9f4eedbe5acc0d89f94b9443b1a5d4"
+            "sha256": "f7a6b4a562031e4545aa5f48e4df9f654b31b15fff9feba4eee8700777b9d494"
```
pricem14pc, line 4: This is a different sha265 to what we have on main `d31989`

pricem14pc: I've raised this PR to cover only the lock file update https://github.com/ONSdigital/ras-frontstage/pull/944

pricem14pc:
https://github.com/ONSdigital/ras-frontstage/pull/944

PR #945, opened 2023-11-16
Update to Design System 66

mfcrocker:
# What and why?
This updates to the latest Design System

# How to test?
Smoke test

# Jira
RAS-927


arroyoAle (approved):
I've clicked around and ran the tests, everything looks right and works

matthew-robinson-ons (approved):
Also found no issues after running the tests and checking the pages

PR #946, opened 2023-11-27
Add Talisman session cookie secure

LJBabbage:
# What and why?
Talisman by default has session cookie secure to True https://github.com/GoogleCloudPlatform/flask-talisman/blob/master/flask_talisman/talisman.py#L78, however we need to be able to control this for the performance environment and dev where it shouldn't be on. This will allow non https sites to still access the cookie which holds the value for csrf. Normal issues with our set-up and that the yaml will not build out till merge
# How to test?
Run the app with data and make sure you can log in
# Jira

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.4.88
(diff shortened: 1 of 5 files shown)


arroyoAle (commented):
Changes look fine and work in dev under acceptance tests and can login to frontstage

arroyoAle (approved):

SteveScorfield (approved):
Tested in my env and acceptance tests ran okay. PR approved.

PR #948, opened 2023-11-29
Update session cookie secure for dev

LJBabbage:
# What and why?
Updates the session cookie secure to false. This will make no difference as the env is in debug which overrides it, but realistically it should be false
# How to test?
There is no need to test, but if you really want to run the dev environment and login to frontstage
# Jira


mfcrocker (approved):

SteveScorfield (approved):
Tested and looks good. PR reviewed and approved.

PR #949, opened 2023-11-29
RAS-938 Upgrade Frontstage to Flask 3

arroyoAle:
# What and why?
Upgrading flask to version 3.0.0, the only changes that seem to affect frontstage are to the markup import which is used to display the success boxes when a message is successfully sent (and other things)

# How to test?
Deploy to your environment and run the acceptance tests, send a message from frontstage and check that the confirmation message is correctly flashed

# Jira
https://jira.ons.gov.uk/browse/RAS-938

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,7 +65,7 @@ jobs:
           make test
       - name: Authenticate with Google Cloud
         id: auth
-        uses: google-github-actions/auth@v0
+        uses: google-github-actions/auth@v2
         with:
           credentials_json: ${{ secrets.GCR_KEY }}
       - name: Setup Google Cloud SDK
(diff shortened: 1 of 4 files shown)


LJBabbage (commented):
Any idea why this was pinned in the first place? I know I did it in rm-reporting because it was using scaffold, seems odd that it was done if there were no changes needed

SteveScorfield (approved):
Tested and reviewed.

arroyoAle:
> Any idea why this was pinned in the first place? I know I did it in rm-reporting because it was using scaffold, seems odd that it was done if there were no changes needed. Might be worth looking at blame in github and seeing if there is a comment on the PR

It was pinned in Steve's PR (https://github.com/ONSdigital/ras-frontstage/pull/939/files) but he had already updated the markup imports when it was pinned. Not sure exactly why it was pinned, seems to be test related issues that were solved by another fix and not unpinned

LJBabbage (approved):
Looks fine, you will inadvertently fix the crytography dependabot alert as well https://github.com/ONSdigital/ras-frontstage/security/dependabot/28

PR #953, opened 2024-01-04
Update CodeQL to v2

matthew-robinson-ons:
# What and why?
This PR updates the deprecated version of CodeQL, also removes an warning message when setting up the python dependencies. 
# How to test?
Check output of the codeQL run against this PR. 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-907


LJBabbage (approved):
Looks like there is a v3 being released very soon, down to the depreciation of Node 16. Probably worth getting a card up on the board for the future as I don't fully understand how this might affect us, if at all.

PR #955, opened 2024-01-09
Resolve test deprecations

matthew-robinson-ons:
# What and why?
This PR addresses the deprecation warnings seen when running the tests locally. 

Setuptools uses a deprecated call to pkg_resources which isn't managed by us. After updating setuptools to the most recent version and no change to the deprecation warnings, this part has been split out into a separate card to investigate a new approach. 

Pytest doesn't like non-test classes starting with Test so this has been renamed. 
# How to test?
run the tests locally, should no longer see a warning message for PytestCollectionWarning.
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&view=detail&selectedIssue=RAS-925


pyproject.toml
```diff
@@ -3,3 +3,9 @@ profile = "black"
 
 [tool.djlint]
 ignore="H021"
+
```
matthew-robinson-ons, line 6: Not sure if this is the direction we want to go with this, will create a card to revisit this if we do implement the filtering.

LJBabbage: I don't think we should be doing this. After some reading it seems there is a migration path to use something else, with that in mind I think we need to explore that first, rather than just ignoring in the code. I am not against that being put on a separate story as it could be chunky.

matthew-robinson-ons: Will look to remove and split out this part of the PR into a separate card.

LJBabbage (approved):

arroyoAle (approved):
Happy with our warning being dealt with and new card made to deal with the setup tools ones

PR #962, opened 2024-03-05
Update to Design System v70

mfcrocker:
# What and why?
We keep up to date with design system releases.

# How to test?
General smoke test paying attention to any janky CSS. Pay special attention to the Secure Message thread list

# Jira
RAS-1093


LJBabbage (changes requested):
This one needs to go back to design system, they have their piping in the wrong order and they haven't documented the component

frontstage/templates/secure-messages/conversation-list.html
```diff
@@ -36,7 +36,8 @@
                         "title": "Closed",
                         "url": url_for('secure_message_bp.view_conversation_list', is_closed='true')
                     }
-                ]
+                ],
+                "headingLevel": 2
```
LJBabbage, line 40: You shouldn't need to provide this anymore

LJBabbage (changes requested):
It looks like you have missed this change https://github.com/ONSdigital/design-system/pull/3047

anwilkie (changes requested):
Screenshot 2024-03-20 at 09 51 18

Tests run okay, but when I was looking around frontstage, I noticed that the formatting of the account details page has gotten messed up, and the `Name:`, `Email address:`, and `Phone number:` fields should be bolded, but aren't.

anwilkie (approved):
The issues I highlighted have been resolved. Looks good to me.

PR #963, opened 2024-03-12
Attempt at replacing python-jose

mfcrocker:
# What and why?
This was a spike, but I may be able to just replace 1:1

# How to test?
General smoke test, especially around areas where we communicate with other services

# Jira
RAS-1092


Pipfile
```diff
@@ -32,6 +31,9 @@ python-gnupg = "*"
 werkzeug = "*"
 fakeredis = "*"
 email-validator = "*"
+pyjwt = "*"
+isort = "*"
+black = "*"
```
LJBabbage, line 36: These are dev packages and shouldn't be included here

LJBabbage (changes requested):

Pipfile
```diff
@@ -32,8 +32,6 @@ werkzeug = "*"
 fakeredis = "*"
 email-validator = "*"
 pyjwt = "*"
-isort = "*"
```
LJBabbage: It looks like you might have forgotten to regenerate the lock file. Sorry for slow re-review, an ill daughter and some time off has delayed things

LJBabbage (approved):

PR #968, opened 2024-04-19
Upgrade requests-mock to bump idna to 3.7

pricem14pc:
# What and why?

Bump requests-mock dev package to bump idna to 3.7


arroyoAle (approved):
Tested and works like normal

matthew-robinson-ons (approved):

PR #969, opened 2024-04-23
RAS-1135 Redis causing acceptance test issues in Frontstage

LJBabbage:
# What and why?
Fixes the acceptance test which fails due to slow response from redis
# How to test?

# Jira


matthew-robinson-ons (approved):
Ran tests and few times and no issues

PR #971, opened 2024-04-26
RAS-1150 Secure messages going to team mailbox instead of to user on ABS

LJBabbage:
# What and why?
The changes in SM recently introduced a bug when answering a thread, the msg_to was not being passed and therefore was always going back to group, this leads to 2 errors
"When an internal user sends a secure message on ABS and the respondent returns a message, this is going to the ABS team and not in the 'my messages' tab of the person who is dealing with the message. "

"Incoming SDCs are showing as Initials and missing the previous message thread, appears to be affecting all survey inboxes"

This PR adds in the msg_to and corrects some other minor issues spotted when going over it. It also restores get_msg_to and the tests that went with it

# How to test?
Create a new collection exercise and make it live (or use acceptance tests) now in ROPs create a user and send a message to an enrolled respondent. Go to frontstage and make sure it arrives, reply to it and confirm it goes back to the user that sent it (i.e your new user) and that the history of the message is there. 

# Jira

### DIFF
--- a/frontstage/views/secure_messaging/message_get.py
+++ b/frontstage/views/secure_messaging/message_get.py
@@ -7,7 +7,7 @@ from structlog import wrap_logger
 
 from frontstage import app
 from frontstage.common.authorisation import jwt_authorization
-from frontstage.common.message_helper import refine
+from frontstage.common.message_helper import from_internal, refine
 from frontstage.controllers.conversation_controller import (
     InvalidSecureMessagingForm,
     get_conversation,
@@ -52,8 +52,11 @@ def view_conversation(session, thread_id):
         subject = refined_conversation[0].get("subject")
         survey_id = refined_conversation[0].get("survey_id")
         business_id = refined_conversation[0].get("ru_ref")
+        msg_to = get_msg_to(refined_conversation)
         try:
-            send_message(request.form, party_id, subject, category, survey_id, business_id)
+            send_message(
+                request.form, party_id, subject, category, msg_to=msg_to, survey_id=survey_id, business_id=business_id
+            )
             thread_url = url_for("secure_message_bp.view_conversation", thread_id=thread_id) + "#latest-message"
             flash(Markup(f"Message sent. <a href={thread_url}>View Message</a>"))
             return redirect(url_for("secure_message_bp.view_conversation_list"))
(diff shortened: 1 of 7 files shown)


frontstage/views/secure_messaging/message_get.py
```diff
@@ -111,3 +114,12 @@ def view_conversation_list(session):
         is_closed=strtobool(is_closed),
         unread_message_count=unread_message_count,
     )
+
+
+def get_msg_to(conversation):
```
LJBabbage, line 119: This has been restored rather than written

frontstage/views/surveys/help/surveys_help.py
```diff
@@ -274,7 +274,9 @@ def send_help_message(session, option: str, sub_option: str):
         category = "TECHNICAL" if option == "info-about-the-ons" else "SURVEY"
         subject = sub_option_dict["subject"]
         try:
-            msg_id = send_message(request.form, party_id, subject, category, survey_id, business_id)
+            msg_id = send_message(
```
LJBabbage, line 277: Doh, forget to put in ce_id

frontstage/views/surveys/help/technical_help.py
```diff
@@ -59,7 +59,7 @@ def send_help_technical_message(session):
     party_id = session.get_party_id()
 
     try:
-        msg_id = send_message(request.form, party_id, "dd", subject)
+        msg_id = send_message(request.form, party_id, subject, "TECHNICAL")
```
LJBabbage, line 62: Didn't spot I left it like this

tests/integration/test_message_get.py
```diff
@@ -75,3 +76,19 @@ def test_view_conversation_message_is_required(self, mock_request, send_message)
         response = self.app.post("secure-message/threads/9e3465c0-9172-4974-a7d1-3a01592d1594", follow_redirects=True)
 
         self.assertTrue("Message is required".encode() in response.data)
+
+    def test_get_msg_to_returns_group_if_no_internal_user(self):
```
LJBabbage, line 80: Restored

matthew-robinson-ons (approved):
Tested and the messages are now being sent back to the intended user and appear in the my messages tab

arroyoAle (approved):
Happy with this, message threads appear correctly and are messages appear in the correct `My messages` for the correct user

PR #976, opened 2024-05-20
RAS-1181 Remove get_business-by-ru-ref from frontstage

LJBabbage:
# What and why?
get_business_by_ru_ref is not used in frontstage, but is calling an endpoint which is used by rops that need rationalising. There are also unnecessary mocks which can be remove
# How to test?
Make sure the function is not used 
# Jira


pricem14pc (approved):
Eyeballed and searched the code, deployed and run all 4 suites of acceptance tests, and carried out a manual test. Agreed this doesn't seem to be used.

matthew-robinson-ons (approved):
Can't find any uses and all working as expected after removal.

PR #979, opened 2024-06-03
Remove EQ version

mfcrocker:
# What and why?
EQ v2 has been dead for some time, this removes the concept of a particular version. It also temporarily mutes the Jinja vulnerability as the fix hasn't yet been released

# How to test?
General smoke test around surveys

# Jira
RAS-1144

### DIFF
--- a/frontstage/common/eq_payload.py
+++ b/frontstage/common/eq_payload.py
@@ -56,11 +56,13 @@ class EqPayload(object):
         ru_ref = f"{party['sampleUnitRef'] + party['checkletter']}"
         int_time = int(time.time())
 
+        # Note that the version mentioned here is the RM->EQ communication version, not the EQ version
         payload = {
             "exp": int_time + 300,
             "iat": int_time,
             "jti": str(uuid.uuid4()),
             "tx_id": tx_id,
+            "version": "v2",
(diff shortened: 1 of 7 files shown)


frontstage/common/eq_payload.py
```diff
@@ -61,7 +61,6 @@ def create_payload(self, case, ce, party_id: str, business_party_id: str, survey
             "iat": int_time,
             "jti": str(uuid.uuid4()),
             "tx_id": tx_id,
-            "version": "v2",
```
LJBabbage, line 64: We don't want to remove the payload version, this will break eQ

LJBabbage (changes requested):

tests/integration/test_generate_eq.py
```diff
@@ -45,7 +45,6 @@
 PAYLOAD = {
     "exp": 1672574700,
     "iat": 1672574400,
-    "version": "v2",
```
LJBabbage, line 48: Needs restoring as well

LJBabbage (changes requested):

arroyoAle (approved):
Tests run fine and there are no actual functional changes needed

PR #996, opened 2024-09-30
RAS-1305 Introduce non root application user

arroyoAle:
# What and why?
Adds an app specific user so we don't have it run on the root user - no changes to the image as the one we are currently using has the least number of vulnerabilities
# How to test?
- deploy this PR to your env
- run tests and make sure frontstage still works
- `kubectl exec -it <pod-name>` into the pod and run `ps aux` to make sure that the new `appuser` is running the app instead of the root user
# Jira
[RAS-1305](https://jira.ons.gov.uk/browse/RAS-1305)


_infra/docker/Dockerfile
```diff
@@ -1,13 +1,16 @@
 # Dockerfile specifically for kubernetes
 FROM python:3.11-slim
 
-RUN apt update && apt install -y build-essential curl gunicorn
+RUN apt update && apt install -y build-essential curl gunicorn procps
```
arroyoAle, line 4: This is the only way I've found to install `ps` now that we are the new user - it will be removed after testing

anwilkie (approved):
Ran the command and confirmed that `appuser` is running the app. Tests passed, too. Approved.

pricem14pc (approved):
Don't forget to revert the `procps` package install

PR #999, opened 2024-10-15
Help journey sign in

matthew-robinson-ons:
# What and why?
This PR updates the sign in page as part of the update to the survey help journey. 
Some of the spacing is a little strange in the figma with overlapping text fields on error messages so i have stuck with design system components which means the spacing doesn't match exactly to the design docs.  

# How to test?
Run the tests
Check the updated page against the figma designs and ensure everything works and looks appropriate given the design. 
# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-1293


frontstage/templates/sign-in/sign-in.html
```diff
@@ -141,12 +127,13 @@ <h1 class="ons-u-fs-xl">Sign in</h1>
                     "label": {
                         "text": "Email Address"
                     },
+                    "value": email if email else '',
```
matthew-robinson-ons, line 130: Passing through the email so it persists on a failed attempt to sign in

.github/workflows/main.yml
```diff
@@ -16,9 +16,10 @@ on:
 
 env:
   IMAGE: frontstage
-  REGISTRY_HOSTNAME: eu.gcr.io
-  HOST: ${{ secrets.GOOGLE_PROJECT_ID }}
-  RELEASE_HOST: ${{ secrets.RELEASE_PROJECT_ID }}
+  REGISTRY_HOSTNAME: europe-west2-docker.pkg.dev
```
LJBabbage, line 19: It can stay, but generally clean separate cards are better

frontstage/templates/sign-in/sign-in.html
```diff
@@ -9,39 +9,33 @@
...
-                {% do errorData.append(
-                    {
-                        "text": errorType[error][0],
-                        "url": "#username_error",
-                        "classes": "ons-js-inpagelink"
-                    }
-                ) %}
-            {% endfor %}
+                } %}
+        {% elif errorType['username'] and errorType['password'] %}
```
LJBabbage, line 21: This is odd duplication with line 33, if the dict has {'type': {'username': ['Email Address is required'] regardless of the password being there it will set the same errorData

frontstage/templates/sign-in/sign-in.html
```diff
@@ -9,39 +9,33 @@
...
+            {% set errorData = {
+                    "text": "Enter an Email Address",
```
LJBabbage, line 23: Why would we not use the error message set in the form, which is in scope?

frontstage/templates/sign-in/sign-in.html
```diff
@@ -9,39 +9,33 @@
 
 {% block main %}
 
-    {% if errorType %}
-        {% set errorData = [] %}
+{% if errorType %}
+        {% set errorData = {} %}
+        {% set errorTitle = 'There is a problem with your answer' %}
         {% if errorType == 'failed' %}
```
LJBabbage, line 15: This error type "failed" is a bit strange and confusing, If I don't enter a password it won't come back as a fail, but by  definition it has failed

frontstage/templates/sign-in/sign-in.html
```diff
@@ -116,17 +105,14 @@ <h1 class="ons-u-fs-xl">You've activated your account</h1>
...
         {% endif %}
 
         {% if errorType == 'failed' %}
-            {% set errorUN = { "text": "Email Address is required", "id": 'username_error' } %}
-        {% elif errorType %}
-            {% set errorUN = { "text": "Invalid email address", "id": 'username_error' } %}
+            {% set errorUN = { "text": "Sign in details are invalid. Enter an email address and a password", "id": 'username_error' } %}
+            {% set errorPW = { "text": "", "id": 'password_error' } %}
+        {% elif errorType['username'] %}
+            {% set errorUN = { "text": "Enter an email address", "id": 'username_error' } %}
```
LJBabbage, line 115: This is all a bit odd and mixing speech marks

LJBabbage (changes requested):
This page is a bit of a muddle and probably needs sorting out before these changes. From the get it is return this dict {'error': {'type': {}, 'logged_in': 'False'} which is odd. The template_data just extends to include account_activated, which doesn't fell right either.

The data={"error": {"type": "failed"}} is a bit strange and should be more refined, it seems to happen when the respondent is locked; we don't understand the error from auth; or the respondent doesn't exist in party, but because it is so generic, when I look at the frontend I struggle to follow the logic, which then doesn't seem right

I found behaviour was a bit odd, if I fill in a email but leave my password empty, my email disappeared but told me I had an error with my password.  

There seems a bit of duplication, mixing of speech marks, logic which could be in the backend and re-use of what should be constants

.github/workflows/main.yml
matthew-robinson-ons: Changes have been merged into to parent branch so no need to include here now

LJBabbage (approved):
Does what was asked on the card, the spacing is like a broken mouth of teeth.

PR #1003, opened 2024-11-01
RAS-1298: Add a new survey page

SteveScorfield:
# What and why?
As part of the user journey improvement work, there is a requirement to update and improve the add survey section. This PR achieves this.

# How to test?
- Run tests locally
- Deploy this PR to your local env and run acceptance tests
- check that it matches the requirements laid out in the Jira ticket

# Jira
https://jira.ons.gov.uk/browse/RAS-1298


frontstage/templates/surveys/surveys-add.html
```diff
@@ -1,17 +1,40 @@
...
+{% set breadcrumbsData = [
+    {
+        "text": "Surveys",
+        "url": "/surveys/todo",
+        "id": "b-item-1"
+    }
+] %}
 
-{% set page_title = "Add a survey" %}
+{% block breadcrumbs %}
```
SteveScorfield, line 22: The style inherited from the design system for the breadcrumbs that there is padding of 8px around the link text. Karen is happy for this to replace the "ons-u-mb-s" (18px) margin.

frontstage/templates/surveys/surveys-add.html
```diff
@@ -68,11 +104,17 @@ <h1 class="u-fs-xl ons-u-mt-s">Add a survey</h1>
...
                 </div>
         </form>
+        {%
+            call onsDetails({
+                "id": "details-example-with-warning",
+                "title": "Need help with the enrolment code?"
+            })
+        %}
+            <p>If your enrolment code has been lost, does not work or you did not receive one, request a new code.</p>
+            <p>To do this, call +44 300 1234 931 or <a href='/surveys/help/something-else'>send a message</a></p>
```
SteveScorfield, line 116: I have spoken to Karen about where specifically this "send a message" link needs to point to. It will eventually point to the link created as part of the "Send message" journey. I have made a note in both cards and links so that this link is updated once the work has been completed.

frontstage/templates/surveys/surveys-add.html
```diff
@@ -1,17 +1,40 @@
...
+{% from "components/panel/_macro.njk" import onsPanel %}
 {% set errorType = data['error']['type'] %}
+{% set panelVariant = "branded" %}
+{% set panelText = "" %}
+
+{% set page_title = "Add a new survey" %}
+
+{% set breadcrumbsData = [
+    {
+        "text": "Surveys",
```
matthew-robinson-ons, line 16: Breadcrumbs should display "back" not "surveys"

frontstage/templates/surveys/surveys-add.html
matthew-robinson-ons: If you go down this route currently it leads to 500 errors but as the journey is updated in a separate card no issues. Just calling it out for the following card associated with adding a survey.

matthew-robinson-ons (changes requested):
One small change to the breadcrumbs text, otherwise looks good.

frontstage/templates/surveys/surveys-add.html
SteveScorfield: Good spot. Not sure how I missed this! I've updated it now.

matthew-robinson-ons (approved):
All looking good and matches the design docs

LJBabbage (changes requested):
I think we need to take another look at this and refactor /add-survey, it is a bit of mess and this just builds on that. There is all sorts of code and logging duplication, exceptions that can't happen and very odd use of  form data which is purely a error dict {"error": {"type": "failed"}} or {"error": {}}. This then makes this bit really odd {% set errorType = data['error']['type'] %} and then the logic % if errorType == "failed" %}. 

The defining of vars to use once i.e {% set errorMsg = 'Enter a valid enrolment code' %} is also not something I would expect and should be fed by the validation step in the backend. There is also some mixing of matching of speech marks which I am unsure why we are doing.

pricem14pc (changes requested):
This PR has been raised against the `main` branch but it should be re-raised against the `survey-respondent-help-journey` development trunk branch.

SteveScorfield:
> This PR has been raised against the `main` branch but it should be re-raised against the `survey-respondent-help-journey` development trunk branch.

This has now been switched

LJBabbage (approved):

PR #1004, opened 2024-11-12
Upgrade gevent to resolve CVE-73655

pricem14pc:
# What and why?

Upgrading gevent to resolve failing python check CVE

# Jira

https://jira.ons.gov.uk/browse/RAS-1354


pricem14pc:
Deployed and tested inc. acceptance tests.

arroyoAle (approved):
lock looks good, frontstage works

matthew-robinson-ons (approved):
Looks good, all working as expected. There is no bot version bump though so will need to manually increase the chart.yaml before merge.

PR #1006, opened 2024-11-20
RAS-1374 Clean up unused files in frontstage

LJBabbage:
# What and why?
This PR removes unused JSON files from our test suit
# How to test?
Because these are all test files, if the test file pass, all is good
# Jira


matthew-robinson-ons (approved):
Thats a lot of files. Tests all pass without them

pricem14pc (approved):

PR #1010, opened 2024-11-26
RAS-1382 Write test for get_survey_list_details_for_party in frontstage

LJBabbage:
# What and why?

# How to test?

# Jira


pricem14pc (approved):
Tests passing locally and on PR. I've eyeballed these but it's almost impossible to give a high level of confidence without essentially re-visiting and re-working them from the ground up. Therefore approving with the understanding these may need tweaking/extending as the work in this area progresses.

matthew-robinson-ons (approved):
Tests passing, eyeballed code and no issues.

PR #1012, opened 2024-11-27
RAS-1380 Integrate party new enrolment and respondent in frontstage

LJBabbage:
# What and why?
This PR integrates the new respondent and enrolment endpoints

# How to test?
The new respondent and enrolment endpoints are used in 15 places. To cover them all you need to
1. Login to an account with an enrolment
2. Go to my account and change your name
3. Go back to my account and change your email, and accept the change
4. Share access to a ce to a new respondent, and accept by the respondent
5. Share access to a ce to respondent that already exists and accept by the respondent
6. Transfer a survey (i.e I am no longer required to complete any surveys in my accounts) and accept by the respondent
7.  Open an survey for an eQ
8. Delete your account

Care must be taken when go from one account to another as the session will be wrong, it's best to logout in each instance. 

**N.B I have checked each scenario, it might be a case of allowing the business testers to do the manual checks and focus on the code**

# Jira


frontstage/views/surveys/surveys_list.py
arroyoAle, line 1: Is [this](https://github.com/ONSdigital/ras-frontstage/pull/1012/files#diff-b2508b942c20641cabf9095efe7b125ad0bb047cea89fdaf42c5eb61ebfc3332L85-L97) still needed? There's nothing left that calls it and the bit that used it has been removed as we're getting only the enabled enrolments by default (it's also hurting the test coverage because its not getting run at all)

arroyoAle (changes requested):
All looks fine and works fine, just one comment around a function thats not used anymore

SteveScorfield (approved):
Tested as agreed with Lewis. Works fine and code looks good. Approved.

PR #1015, opened 2024-12-12
Upgrade python version to 3.12

matthew-robinson-ons:
# What and why?
Upgrades python to 3.12 as part of ongoing tech upgrades. 

# How to test?
Remove your existing pipenv with 3.11
Install python 3.12 (I used pyenv) either globally or a new local version.
Create a new environment with pipenv install 
Check your python interpreter has updated or add a new one based off the created pipenv if it isn't using 3.12
Run the tests and app locally 
Deploy this PR, run acceptance tests and check frontstage is working correctly. 

# Jira
https://jira.ons.gov.uk/secure/RapidBoard.jspa?rapidView=1493&projectKey=RAS&view=detail&selectedIssue=RAS-1379

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,10 +49,10 @@ jobs:
         with:
           fetch-depth: '0'
           token: ${{ secrets.BOT_TOKEN }}
-      - name: Set up Python 3.11
+      - name: Set up Python 3.12
         uses: actions/setup-python@v4
         with:
-          python-version: '3.11'
+          python-version: '3.12'
       - name: Build
         run: |
           python -m pip install --upgrade pip
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11-slim
+FROM python:3.12-slim
 
 RUN apt update && apt install -y build-essential curl
 RUN pip install pipenv
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ verify_ssl = true
 url = "https://pypi.python.org/simple"
 
 [requires]
-python_version = "3.11"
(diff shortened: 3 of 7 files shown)


anwilkie (changes requested):
When I try and run tests locally, I get the following error: `ModuleNotFoundError: No module named 'distutils'`. When I looked up the origin of this error, I discovered that the distutils module is no longer compatible with python 3.12 (and was deprecated back in 3.10). This library is called in a few places on frontstage. [This needs to be replaced with setuptools instead.](https://peps.python.org/pep-0632/)

matthew-robinson-ons:
> When I try and run tests locally, I get the following error: `ModuleNotFoundError: No module named 'distutils'`. I looked up the origin of this error and discovered that the `distutils` module is no longer compatible with python 3.12 (and was deprecated back in 3.10). This library is called in a few places on frontstage. [This needs to be replaced with setuptools instead.](https://peps.python.org/pep-0632/)

If you run `pip install setuptools` it will fix this error. A new pipenv doesn't install setuptools anymore in 3.12, i had a local install and we include setuptools in our lock file which is why it passed on the pr build.

We only use distutils to import strtobool which can't be migrated to setuptools https://peps.python.org/pep-0632/#migration-advice:

> For these functions, and any others not mentioned here, you will need to reimplement the functionality yourself. The legacy documentation can be found at https://docs.python.org/3.9/distutils/apiref.html
distutils.dir_util.create_tree
distutils.util.change_root
distutils.util.strtobool

Will create a card to implement our own version of strtobool and replace the deprecation in the future, I tested a way that works but as we use strtobool in our config it is not a straightforward replacement.

anwilkie (approved):
Re-tested, and it looks good to me.

PR #1016, opened 2024-12-17
RAS-1400 Extend party enrolment to include business and survey details

LJBabbage:
# What and why?
This PR updates the survey_list build process, instead of using redis and calling to party/survey multiple times it is provided on the first call. It also adds a call to a new endpoint specially for checking a respondent is enrolled 
# How to test?
Deploy this card as well https://github.com/ONSdigital/ras-party/pull/449. Add some enrolment data to your account, acceptance tests could be used, but you might want to add another IAC as well. If it logs in you are good. Also check you can access a eq, download and upload a seft
# Jira

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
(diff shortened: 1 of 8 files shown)


matthew-robinson-ons (approved):
Functions correctly and code changes look good to me

SteveScorfield (approved):
Changes look fine and tests all okay. Approved

PR #1019, opened 2025-01-06
RAS-1402: Replace deprecated distutils package in Frontstage

SteveScorfield:
# What and why?
Distutils is used by Frontstage to import the 'srttobool' function. Distutils is deprecated and is no longer imported by Python 3.12 and requires a manual approach to get it to build locally. Based on the documentation, it is recommended to reimplement this function independently. This PR achieves this.
 
# How to test?
- Check unit tests run
- Deploy this PR to your env
- Run the acceptance checks and check that they pass
- Check that all messages are retrieved in the message section
- Check you are able to register a new account

# Jira
https://jira.ons.gov.uk/browse/RAS-1402


LJBabbage (changes requested):
Our current process is really lumpy, in nearly every place we are taking in a string, converting it with strtobool to 0 or 1 then using python bool to make it a bool. I can't get my brain to understand why and why the function is called strtobool, ok you can argue 1/0 is bool, but it's usual denoted version is true/false. That and the function being over implemented for our needs as we don't use y, yes, t, on, etc. I say we bin off their implementation and have our own proper strtobool which does just what we want. i.e 
```
if value == "true"
    return True
```
The one in message_get.py (the one that leaves it as an int) won't matter if it is a proper bool (True/False)

LJBabbage (approved):
All good, and routes identified tested

matthew-robinson-ons (approved):
Tested and works as expected. Wasn't sure about having in the root directory but as the function is in the config and application will need to be there to cover both uses.

SteveScorfield:
This has been reviewed and can now be merged ready for a later deployment to prod of testing.

PR #1021, opened 2025-01-09
Fix issue of sending blank surveyID to SMV2

SteveScorfield:
# What and why?
It was was found during the testing of https://jira.ons.gov.uk/browse/RAS-1402 that a message could not be sent from Frontstage when the secure message V2 was switched on. This wasn't noticed until preprod as this was unexpected. It was nothing to do with the code change as part of the reference Jira card it was just a more in depth test was conducted which highlighted the defect. 

The problem being was that SM V1 stores empty survey_ids as empty strings whereas SM V2 uses a 'None' value. So this will break SM V2 which is implemented to except optional UUIDs.

This change should fix the issue. 

# How to test?
- Send a new 'Something else' message in Frontstage via the 'My Account' screen
- Go back to the sent message screen and try to send another message in the same thread
- This should send correctly and you should be sent back to the message screen. 

# Jira
No ticket but related to https://jira.ons.gov.uk/browse/RAS-1402

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.37
+version: 2.5.38
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application.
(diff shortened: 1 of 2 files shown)


LJBabbage (approved):

matthew-robinson-ons (approved):
Messages sending with no survey associated successfully.

PR #1027, opened 2025-01-16
RAS-1198 Contact us (after log in)

LJBabbage:
# What and why?
This PRs makes the contact_us page dynamic on whether a respondent is authorised or not
# How to test?
Go to the contact us page, when logged in and logged out
# Jira

### DIFF
--- a/frontstage/templates/contact-us.html
+++ b/frontstage/templates/contact-us.html
@@ -25,10 +25,15 @@
 {% block main %}
     <h1 class="ons-u-fs-2xl">Contact us</h1>
 
-    <h2>
-        Email
-    </h2>
-        <p><a href="mailto:surveys@ons.gov.uk">surveys@ons.gov.uk</a></p>
+    {% if authorization %}
+        <h2>Message us</h2>
+        <p>We will get back to you within 5 working days</p>
+        <p><a id="contact_us" href="{{ url_for('contact_us_bp.send_message') }}">Send message</a>
+     </p>
+    {% else %}
+         <h2>Email</h2>
+         <p><a href="mailto:surveys@ons.gov.uk">surveys@ons.gov.uk</a></p>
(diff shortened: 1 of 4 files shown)


frontstage/views/contact_us.py
```diff
@@ -22,7 +26,16 @@
...
 def contact_us():
-    return render_template("contact-us.html")
+    authorization = False
+    if session_key := request.cookies.get("authorization"):
+        redis_session = Session.from_session_key(session_key)
+        try:
+            validate_jwt(redis_session, session_key)
+            authorization = True
+        except (Unauthorized, DecodeError, JWTValidationError, JWTTimeoutError):
+            pass
```
LJBabbage, line 36: Thought about putting logging in here, but I think it is pointless considering the context. This is to either to display an email or a hyperlink to the secure message form if authorised. I can see not scenario when we debug this, If it was a decorated endpoint maybe, but realistically we have extended a non @jwt_authorization route to decide if a hyperlink is shown.

frontstage/templates/contact-us.html
matthew-robinson-ons, line 1: When logged in the back button takes me to the sign in page and not to the home page which is a bit confusing. I can't find any doc with the intended behaviour though, i'm presuming it should go to the homepage.

matthew-robinson-ons (changes requested):
Send Message functionality looks good and works as intended just a query around the back breadcrumb.

frontstage/templates/contact-us.html
LJBabbage: Should go back to origin so contact_us, I will correct it

arroyoAle (approved):
Looks good and matches design and functionality works

matthew-robinson-ons (approved):
Functions as intended, breadcrumbs to be addressed on a separate story.

PR #1032, opened 2025-01-24
Get collection exercises by survey ids

arroyoAle:
# What and why?

# How to test?

# Jira

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,16 +19,15 @@ start: load-design-system-templates
 docker-test: REDIS_PORT=6379
 docker-test: test
 
-#remove -i 70612 once jinja2 is upgraded past v3.1.4
 lint:
-	pipenv check -i 70612
+	pipenv check
 	pipenv run isort .
 	pipenv run black --line-length 120 .
 	pipenv run djlint frontstage/ --ignore=H037,H021
 	pipenv run flake8
 
 lint-check: load-design-system-templates
-	pipenv check -i 70612
+	pipenv check
 	pipenv run isort . --check-only
 	pipenv run black --line-length 120 --check .
 	pipenv run djlint frontstage/ --ignore=H037,H021
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -709,11 +709,12 @@
         },
         "jinja2": {
             "hashes": [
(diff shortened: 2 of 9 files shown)


anwilkie (approved):

arroyoAle:
> So I don't believe this will work if there are no enrolments and will throw an exception. I also am not sure why this is still stored in redis, hasn't that been made redundant? I am also wonder what exercise /surveys is also being used for (if any) as it has been extended

Updated to be able to handle no enrolments and removed redis. I think you mean the old endpoint in collection exercise? In which case its still used by rOps, party and the acceptance tests

LJBabbage:
> > So I don't believe this will work if there are no enrolments and will throw an exception. I also am not sure why this is still stored in redis, hasn't that been made redundant? I am also wonder what exercise /surveys is also being used for (if any) as it has been extended
> 
> Updated to be able to handle no enrolments and removed redis. I think you mean the old endpoint in collection exercise? In which case its still used by rOps, party and the acceptance tests

It probably should be on the other PR. You added case events to be returned from 
```@RequestMapping(value = "/surveys", method = RequestMethod.GET, produces = "application/json")``` 
 I wondered what other things that is used for and if there is a performance hit

arroyoAle:
> > > So I don't believe this will work if there are no enrolments and will throw an exception. I also am not sure why this is still stored in redis, hasn't that been made redundant? I am also wonder what exercise /surveys is also being used for (if any) as it has been extended
> > 
> > 
> > Updated to be able to handle no enrolments and removed redis. I think you mean the old endpoint in collection exercise? In which case its still used by rOps, party and the acceptance tests
> 
> It probably should be on the other PR. You added case events to be returned from `@RequestMapping(value = "/surveys", method = RequestMethod.GET, produces = "application/json")` I wondered what other things that is used for and if there is a performance hit

Oh I see, I've also updated the other pr so if no survey ids are passed it returns a 204, and the `/surveys` endpoint we are now using here does not seem to be used by anything else from a github search - https://github.com/search?q=org%3AONSdigital+%2Fcollectionexercises%2Fsurveys&type=code

LJBabbage:
Just waiting for build to be fixed on this one

LJBabbage (approved):

PR #1033, opened 2025-01-30
Useful links component

matthew-robinson-ons:
# What and why?
Introduces the useful links component to the survey home and completed surveys page
# How to test?
Check the component matches the intended design docs and that the navigation is as expected.

**N.B.** 
- The ui is a bit compact as it is not using the updated surveys home and completed pages. 
- The tablist also looks a bit odd as it doesn't use the full grid width but that is being removed on the banner menu heading ticket. 
- The breadcrumbs on the last 2 links don't work and take you to the sign in page and the cancel buttons take you to unexpected pages. We need to revisit breadcrumbs to make them dynamic.

# Jira
[RAS-1285](https://jira.ons.gov.uk/browse/RAS-1285)

### DIFF
--- a/frontstage/templates/partials/useful_links.html
+++ b/frontstage/templates/partials/useful_links.html
@@ -6,33 +6,33 @@
                     "ariaLabel": 'Related content',
                     "rows": [
                         {
-                            "id": 'useful-links',
-                            "title": 'Useful links',
+                            "id": "useful-links",
+                            "title": "Useful links",
                             "itemsList": [
                                 {
-                                    "id": 'add_survey_button',
-                                    "url": url_for('surveys_bp.add_survey'),
-                                    "text": 'Add a new survey',
+                                    "id": "add_survey_button",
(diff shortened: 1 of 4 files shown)


frontstage/templates/partials/useful_links.html
```diff
@@ -0,0 +1,43 @@
...
+                                {
+                                    "url": url_for("account_bp.share_survey_overview"),
+                                    "text": "Share access to surveys",
+                                },
+                                {
+                                    "url": url_for("account_bp.transfer_survey_overview"),
+                                    "text": "Transfer surveys"
+                                },
+                                {
+                                    "url": url_for("contact_us_bp.send_message"),
```
matthew-robinson-ons, line 26: The docs have this as the old get help with this survey link, however have matched the current implementation of that link which takes you to the send message page.

LJBabbage (approved):
We definitely need a wash up exercise as the edges are getting blurred between these stories, but this seems good to merge

PR #1041, opened 2025-02-25
Update contact us breadcrumb navigation

matthew-robinson-ons:
# What and why?
This PR updates the breadcrumb navigation on the contact us page, to reflect the new survey help journey. 
# How to test?
Deploy this PR to your environment 
Check when not signed in that the breadcrumbs take you back to the sign in page. 
Check when signed in that the contact us breadcrumbs either from the footer or from the useful links takes you back to the previous page you were on. 
# Jira
[RAS-1489](https://jira.ons.gov.uk/browse/RAS-1489)

### DIFF
--- a/frontstage/templates/contact-us.html
+++ b/frontstage/templates/contact-us.html
@@ -8,7 +8,7 @@
 {% set breadcrumbsData = [
     {
         "text": "Back",
-        "url": "/sign-in/",
+        "url": "/sign-in/" if not authorization else request.referrer,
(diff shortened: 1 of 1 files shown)


arroyoAle (approved):
Happy with this, send you back to the page you were in before contact us when signed in. Only thing is that when not signed in it always takes you back to the sign in page, which is how it was and the story has no comments on this path, so happy with how it is now

anwilkie (approved):

PR #1044, opened 2025-02-26
Survey respondent help journey: Edit account details

mfcrocker:
# What and why?
Improvements to the edit account details journey

# How to test?
Go to my account and edit account details, check it matches the Figma journey (I've cleared with Alina that we should be maintaining the functionality around sending a confirmation email, but have tidied it up significantly to fit with the new design)

# Jira
RAS-1449


LJBabbage (changes requested):
Looks like a problem when it errors. I just pressed change details, and got this. I got the same when I forced the email to error, i.e removed the @ 

Screenshot 2025-03-04 at 13 41 38

PR #1046, opened 2025-03-05
RAS-1502 Apply Dynamic Navigation 'Share access to my surveys'

LJBabbage:
# What and why?

# How to test?

# Jira


matthew-robinson-ons (approved):
non-dynamic navigation works and able to move between the pages using the cancel and back buttons.

PR #1047, opened 2025-03-12
RAS-1519 Secure message - no enrolments

LJBabbage:
# What and why?
This PR make sure you can still send a secure message without an enrolment
# How to test?
I find it easiest to run the acceptance tests, then go to the db (party) and remove the enrolment for the respondent. Once the respondent has no enrolments, make sure you can send a secure message
# Jira


matthew-robinson-ons (approved):
Tested and works correctly. Was able to send a message with no enrolments and with enrolments. Agree with comments on the jira card that it's strange to select help with this survey without any enrolments but as a fix for now looks good.

anwilkie (approved):

PR #1049, opened 2025-03-19
Improve logging around SEFT upload

mfcrocker:
# What and why?
We currently throw exceptions when the user uploads a SEFT without choosing a file or when their file is too small, causing logging noise. This rectifies that.

(The original card only specified the missing file, but after discussions with @pricem14pc we agreed this should also encompass file too small)

# How to test?
Check we aren't throwing exceptions when you upload a SEFT response, and also ensure that it still works properly when they upload a proper file.

# Jira
RAS-1485


frontstage/controllers/collection_instrument_controller.py
```diff
@@ -113,16 +113,17 @@ def upload_collection_instrument(file, case: dict, business_party: dict, party_i
             gcp_survey_response.upload_seft_survey_response(case, file_contents, file_name, survey_ref)
             ci_post_case_event(case_id, party_id, "SUCCESSFUL_RESPONSE_UPLOAD")
             logger.info("Successfully uploaded collection instrument", case_id=case_id, party_id=party_id)
+            return None
         except FileTooSmallError:
             ci_post_case_event(case_id, party_id, "UNSUCCESSFUL_RESPONSE_UPLOAD")
-            raise CiUploadError(FILE_TOO_SMALL)
+            return FILE_TOO_SMALL
```
SteveScorfield, line 119: Just a question, can a file ever bee too small for the purposes of the SEFT upload?

SteveScorfield (approved):
Tested and works as expected. Code looks good. Approved.

matthew-robinson-ons (approved):
Ran tests locally and spinnaker. All passing. Code changes look good and able to see info messages when file is not attached and in other scenarios such as file name too large etc.

frontstage/controllers/collection_instrument_controller.py
matthew-robinson-ons: We have a method that checks if a file is above a byte but i don't think a xlsx could ever be less than a byte.

```
    def check_if_file_size_too_small(file_size) -> bool:
        return file_size < 1
```

frontstage/views/surveys/upload_survey.py
```diff
@@ -62,10 +62,23 @@ def upload_survey(session):
 
     try:
         # Upload the file to the collection instrument service
-        collection_instrument_controller.upload_collection_instrument(
+        error_text = collection_instrument_controller.upload_collection_instrument(
             upload_file, case, business_party, party_id, survey
         )
+        if error_text is not None:
```
SteveScorfield, line 68: Shouldn't this be "if error_text"?  You wouldn't need the 'is not None' then

tests/unit/controllers/test_collection_instrument_controller.py
```diff
@@ -103,16 +103,6 @@ def test_upload_collection_instrument_success(self, *_):
                 except (ApiError, CiUploadError):
                     self.fail("Unexpected error thrown when uploading collection instrument")
 
-    @patch("frontstage.controllers.case_controller.post_case_event")
```
SteveScorfield, line 106: Why is this being removed?

SteveScorfield (changes requested):
I was just checking my question I raised and realised I had missed the 2 highlighted changes.

tests/unit/controllers/test_collection_instrument_controller.py
matthew-robinson-ons: Yeah test coverage could be improved with the new logic added above

frontstage/views/surveys/upload_survey.py
mfcrocker: There's fundamentally no difference so it's a matter of taste

PR #1051, opened 2025-03-20
Info page back link

matthew-robinson-ons:
# What and why?
This Pr updates the back link on /help/info-ons to take you to back to the todo page. 

It also removes now unaccessible pages:
- /help
- /help/something-else
- /help/help-with-my-password
As we go directly to the /help/info-ons page now. 
# How to test?
Check the breadcrumbs work correctly and navigation from useful links to /help/info-ons and back works as intended. 
# Jira
[RAS-1528](https://jira.ons.gov.uk/browse/RAS-1528)


frontstage/templates/help/help-info-ons.html
```diff
@@ -5,12 +5,8 @@
...
 {% set breadcrumbsData = [
     {
-        "text": "Sign in",
-        "url": "/sign-in"
-    },
-    {
-        "text": "Help",
-        "url": "/help"
+        "text": "Back",
+        "url": "/surveys/todo"
```
SteveScorfield, line 9: Incase you click the link when you are on the completed tab and load the info about ONS, then shouldn't this return back to he completed tab?

SteveScorfield (changes requested):
It all looks good. I've added a question around the link. I'm also wondering if a test should be include to test this.

frontstage/templates/help/help-info-ons.html
matthew-robinson-ons: Ideally yes, but we need a rework of all the back links and navigation to make them dynamic going forward. For mvp the navigation of the back links to todo is the simplest, otherwise we will have to loop back and update all breadcrumbs to make sure they play well together.

matthew-robinson-ons: With the test, i didn't think we tested on breadcrumbs, just on the loading of the resulting page which is covered in the testing for info-ons. I do need to remove some of the views though from those pages so, which i'll do now. Coverage should go up as a result.

matthew-robinson-ons: I am a bit confused as to why all the testing for the help pages are named sign_out_help, it might be worth revisiting all this afterwards? especially after removing some of the pages as some of the naming and tests are a bit strange

LJBabbage (approved):

SteveScorfield (approved):
I'm happy this works correctly. Tested and approved.

PR #1055, opened 2025-03-27
Help sm

LJBabbage:
# What and why?
Updates the other 2 routes in "Information about the Office for National" to show direct link to secure message if logged in
# How to test?
Check the new routes work
# Jira


matthew-robinson-ons (approved):
Tested and works well. Separating the need more info makes sense using it across all pages.

arroyoAle (approved):
Works as expected, I get taken to the secure message page when I click the link in either page

PR #1056, opened 2025-04-03
RAS-1539 Remove secure message count from frontstage

LJBabbage:
# What and why?
This code removes the secure message count being called and stored in redis
# How to test?
Make sure the application works as before
# Jira

### DIFF
--- a/frontstage/common/session.py
+++ b/frontstage/common/session.py
@@ -8,7 +8,6 @@ class Session(object):
     def __init__(self, session_key, encoded_jwt_token):
         self.encoded_jwt_token = encoded_jwt_token
         self.session_key = session_key
-        self.persisted = False
 
     @classmethod
     def from_session_key(cls, session_key):
@@ -18,22 +17,22 @@ class Session(object):
         else:
             encoded_jwt_token = redis.get(session_key)
         session = cls(session_key, encoded_jwt_token)
-        session.persisted = True
+
         return session
 
     @classmethod
     def from_party_id(cls, party_id):
-        """Create a new and unpersisted session object from a party_id, this will encode a JWT
-        and set an expiry time, it will default the unread message count to 0"""
+        """Create a new session object from a party_id, this will encode a JWT
+        and set an expiry time"""
         data_dict = {
             "party_id": party_id,
             "role": "respondent",
-            "unread_message_count": {"value": 0, "refresh_in": _get_new_timestamp(ttl=300)},
             "expires_in": _get_new_timestamp(),
         }
         encoded_jwt_token = jwt.encode(data_dict)
         session_key = str(uuid4())
         session = cls(session_key, encoded_jwt_token)
+        session.set()
         return session
 
     def refresh_session(self):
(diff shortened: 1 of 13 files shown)


frontstage/common/session.py
```diff
@@ -18,49 +17,36 @@ def from_session_key(cls, session_key):
...
         session = cls(session_key, encoded_jwt_token)
-        session.persisted = True
+
         return session
 
     @classmethod
     def from_party_id(cls, party_id):
-        """Create a new and unpersisted session object from a party_id, this will encode a JWT
-        and set an expiry time, it will default the unread message count to 0"""
+        """Create a new session object from a party_id, this will encode a JWT
```
LJBabbage, line 25: Perhaps historically from_party_id was used by something else which needed it not to be persisted and set, in this instance it now is only used on sign_in.py, which instantly persists and set afterwards, basically we don't need to do that anymore

frontstage/common/session.py
```diff
@@ -8,7 +8,6 @@ class Session(object):
     def __init__(self, session_key, encoded_jwt_token):
         self.encoded_jwt_token = encoded_jwt_token
         self.session_key = session_key
-        self.persisted = False
```
LJBabbage, line 11: This doesn't seem to have any use. It's a generic object so isn't a attribute of redis so seems historically, maybe it was useful at some stage

anwilkie (approved):

frontstage/common/session.py
```diff
@@ -73,12 +59,8 @@ def get_decoded_jwt(self):
     def get_expires_in(self):
         return jwt.decode(self.encoded_jwt_token)["expires_in"]
 
-    def is_persisted(self):
-        return self.persisted
-
-    def save(self):
+    def set(self):
```
pricem14pc, line 62: This is probably an acceptable change given the single (Redis) backend implementation. I'm guessing this was intended to be a generic "save" interface that could then have any backend library implementation. But they never got that far.

frontstage/common/session.py
pricem14pc: From what I can see, this attribute was introduced solely in some slightly clunky logic around updating the Redis session with the message count (probably a workaround to a bug). Happy that's no longer required, and can be removed.

frontstage/common/session.py
pricem14pc: Yes, I think this was a workaround to a bug in the count not being handled correctly under some scenarios. Fortunately removing the feature removed the need to know if the session object has been persisted in Redis or not.

pricem14pc (approved):
Tested all features impacted, reviewed code and happy with the session persisted attribute being removed.

frontstage/common/session.py
LJBabbage: It's not ideal and I believe setex is depreciated, however it wasn't my intention to hang around on this bit, make it align marginally better and get a story up to look to migrate to set with EX attribute

frontstage/common/session.py
LJBabbage: My thoughts as well

PR #1076, opened 2025-10-03
Allow handing of base help route on top of 2.5.67

pricem14pc:
# What and why?

Allow handing of base help route

# How to test?

# Jira


SteveScorfield (approved):
Tested and works as expected. Approved.

arroyoAle (approved):

PR #1107, opened 2026-05-07
RAS-1823: RFC: Secure Message retention information- frontstage

SteveScorfield:
# What and why?
As part of the message deletion work, we need to inform the business users that there will be a 30 day retention policy on closed messages. This PR achieves that.

# How to test?
Check unit tests work
1. Deploy this PR to your env
2. Run acceptance tests
3. Log into the UI, navigate to the message list section and the message should be displayed correctly with coreect spacing based on the Figma diagram

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1823

### DIFF
--- a/_infra/helm/frontstage/Chart.yaml
+++ b/_infra/helm/frontstage/Chart.yaml
@@ -14,8 +14,8 @@ type: application
 
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
-version: 2.5.90
+version: 2.5.91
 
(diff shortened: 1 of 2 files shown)


LJBabbage (commented):
The image is slightly ambiguous, this message will appear on both open and closed tabs. The image only shows the closed tab and doesn't detail what should be on open. Is a message about closed messages relevant in this situation, although it is also the landing page which muddies the water

LJBabbage (approved):

arroyoAle (approved):
Happy this displays correctly

PR #1108, opened 2026-05-11
RFC : Secure Messages banner update - frontstage

SteveScorfield:
# What and why?
As part of the secure message deletion work, we need to inform a user when they click on their closed message, that this message will be deleted in 30 days along with the time it was closed. This message needs to disappear if the closed message is reopened. As the original banner has been replaced with this new message, the links to go back to messages has been removed and so a back button is required to enable a smooth user experience.  

# How to test?
1. Deploy this PR to your env
2. Run acceptance tests to generate data
3. In ROPS close the users message
4. Goto Frontstage -> Messages -> Closed messages and click on the closed message
5. A back button should be present along with the new message
6. Go back to ROPS and reopen the message
7. Go back to Frontstage, click on the reopened message and the new message should no longer be present
8. Clock the back link and you will return to the opened message list
9. Run acceptance tests again to ensure nothing is broken

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1905


frontstage/templates/layouts/configs/breadcrumbs.html
arroyoAle, line 1: I can't see this being used, is it needed?

arroyoAle (commented):
Looks good, new panel is shown, back button appears and takes users back to the correct messages page, there is a way to get to the message from elsewhere but the acceptance criteria says it always goes back the the respective messages page. Just one comment around the new template

frontstage/templates/layouts/configs/breadcrumbs.html
SteveScorfield: I forgot to remove this. Removed.

arroyoAle (approved):

matthew-robinson-ons (approved):
All working as per the acceptance criteria

PR #1113, opened 2026-06-01
RAS-1927: Upgrade frontstage to python 3.14

SteveScorfield:
# What and why?
We need to update the version of Python from 3.12 to 3.14. This PR achieves this.

**Note**: The google-cloud-profiler isn't used in any of our environments, so it has been removed.

# How to test?
1. Check acceptance tests work
2. Generic regression tests around the UI

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1927


matthew-robinson-ons (dismissed):
The profiler is currently disabled in all envs but is it not used point in time? I'm not too familiar with it's usage but if it's not used we have references to it in ras-rm-config and service account updates which relate to it which should also be removed so we don't have loose ends. Either on this card or as a mop up afterwards.

matthew-robinson-ons (changes requested):
The profiler is currently disabled in all envs but is it not used point in time? I'm not too familiar with it's usage but if it's not used we have references to it in ras-rm-config and service account updates which relate to it which should also be removed so we don't have loose ends. Either on this card or as a mop up afterwards.

SteveScorfield:
> The profiler is currently disabled in all envs but is it not used point in time? I'm not too familiar with it's usage but if it's not used we have references to it in ras-rm-config and service account updates which relate to it which should also be removed so we don't have loose ends. Either on this card or as a mop up afterwards.

I see what you mean. However, the last time we used the profiler is back in November 2023 for dev and this was turned off in June 2024 as it was accidentally left on causing irrelevant logging. I don't think the effort to fix this based on the amount we use it, is worth it. If we need a profiler later on, then we can investigate a compatible one to use. 

I missed the config and there will be some changes in terraform as well. I'll just add 2 more PRs to the card. It's linked so it makes sense to do it all in one go.

matthew-robinson-ons (approved):

warrenbaileyons:
Are we removing the profiler because it doesn't work with 3.14 or just that it hasn't been used since 2023?

SteveScorfield:
> Are we removing the profiler because it doesn't work with 3.14 or just that it hasn't been used since 2023?

In both cases. It doesn't work with 3.14 and it hasn't been used since 2023. For the effort to replace something we rarely use (or will ever use), I don't think it makes sense to get it working/replace it.

warrenbaileyons:
Tested and works

warrenbaileyons (approved):

PR #1124, opened 2026-08-06
RAS-1861 update GA script

JacobHarrisonONS:
# What and why?
update GA script to track users

# How to test?
add this to config

analytics:
  enabled: true
  measurementID: G-SPB3W8KTCF

enable GA in dev and deploy frontstage to your dev
this should link your dev with preprod GA
log in and do some stuff, the GA dashboard should detect the user

# Jira
https://officefornationalstatistics.atlassian.net/browse/RAS-1855


anwilkie (approved):
Tested out, and was able to see user activity on the GA dashboard after I went on frontstage in dev.

matthew-robinson-ons (approved):
Able to see active users, when providing a tag. 

The history in the dashboard doesn't seem to be storing correctly for me but I don't think that's related to this card and probably a configuration within tag manager itself.

PR #1125, opened 2026-08-07
RAS-2015 AI Security Issue - Password reset POST accepts any valid signed email token without proving it is the current stored reset token

arroyoAle:
# What and why?
Adding new endpoint used in party to pass the token and handle the verification logic there. Also relocking the pipenv since the glcoud storage was above the pinned version
# How to test?
- Deploy your env
- Run the acceptance tests
- Follow the password reset path in frontstage twice, and note down the tokens both times
- Run this curl command to exercise the exploit:
```
BASE_URL="http://localhost:8082"
TOKEN="replace-with-old-reset-token"

curl -sS -i \
  -X POST "$BASE_URL/passwords/reset-password/$TOKEN" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data-urlencode 'password=Gizmo007!Gizmo' \
  --data-urlencode 'password_confirm=Gizmo007!Gizmo'
```
- Log in with the new password
- Deploy this PR and https://github.com/ONSdigital/ras-party/pull/489, and run the curl command again, you can add ` > test.html` to see the full html output
- Make sure that the attempt is blocked and the password is not changed and the valid token is still in the party db
- Kill the auth deployment and test that the db is rolled back when the connection fails and the vaild token is still in the party db
- Redeploy auth
- Test using the up to date token and that the password resets
# Jira
[RAS-2015](https://officefornationalstatistics.atlassian.net/browse/RAS-2015)

[RAS-2015]: https://officefornationalstatistics.atlassian.net/browse/RAS-2015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frontstage/views/passwords/reset_password.py
```diff
@@ -63,6 +63,14 @@ def post_reset_password(token):
     try:
         duration = app.config["EMAIL_TOKEN_EXPIRY"]
         email = verification.decode_email_token(token, duration)
+        respondent = party_controller.get_respondent_by_email(email)
+        party_token = respondent.get("password_verification_token") if respondent else None
+
+        if not party_token or token != party_token:
```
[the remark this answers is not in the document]
arroyoAle: Addressed by https://github.com/ONSdigital/ras-party/pull/489

matthew-robinson-ons (approved):
Looks good for using the new party api

matthew-robinson-ons (approved):

frontstage/controllers/party_controller.py
```diff
@@ -100,6 +100,23 @@ def change_password(email, password):
     bound_logger.info("Successfully changed password through the party service")
 
 
+def reset_password(email, password, token):
```
LJBabbage, line 103: Type hinting and return type would be nice

frontstage/controllers/party_controller.py
```diff
@@ -100,6 +100,23 @@ def change_password(email, password):
...
+    bound_logger = logger.bind(email=obfuscate_email(email))
+    bound_logger.info("Attempting to reset password through the party service")
```
LJBabbage, line 105: For me I always find the attempting, then passed/failed combination over kill. Passed or failed does the job we need

frontstage/controllers/party_controller.py
```diff
@@ -100,6 +100,23 @@ def change_password(email, password):
...
+
+    data = {"email_address": email, "new_password": password, "token": token}
+    url = f"{app.config['PARTY_URL']}/party-api/v1/respondents/reset_password"
+    response = requests.put(url, auth=app.config["BASIC_AUTH"], json=data)
+
+    try:
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        bound_logger.error("Failed to send reset password request to party service")
```
LJBabbage, line 114: That error message needs changing

LJBabbage: I thought we were catching time outs and missing services now a days?

frontstage/views/passwords/reset_password.py
```diff
@@ -63,19 +62,17 @@ def post_reset_password(token):
     try:
         duration = app.config["EMAIL_TOKEN_EXPIRY"]
         email = verification.decode_email_token(token, duration)
-        party_controller.change_password(email, password)
-        party_controller.delete_verification_token(token)
+        party_controller.reset_password(email, password, token)
     except ApiError as exc:
```
LJBabbage, line 66: As said in the other PR, these don't quite align with what can come back

LJBabbage (changes requested):

frontstage/controllers/party_controller.py
arroyoAle: Removed the attempt log

frontstage/controllers/party_controller.py
arroyoAle: Updated the log and added the connection error and time out exceptions

frontstage/views/passwords/reset_password.py
```diff
@@ -58,24 +57,26 @@ def post_reset_password(token):
     if not form.validate():
         return get_reset_password(token, form_errors=form.errors)
 
-    password = request.form.get("password")
+    password = request.form["password"]
 
     try:
```
LJBabbage: This whole try is a bit of a muddle decode_email_token can raise among other things SignatureExpired and BadSignature, but it's in a try that won't catch it

frontstage/views/passwords/reset_password.py
```diff
@@ -58,24 +57,26 @@ def post_reset_password(token):
...
         duration = app.config["EMAIL_TOKEN_EXPIRY"]
         email = verification.decode_email_token(token, duration)
-        party_controller.change_password(email, password)
-        party_controller.delete_verification_token(token)
+        party_controller.reset_password(email, password, token)
     except ApiError as exc:
-        if exc.status_code == 409:
+        if exc.status_code == 400:
+            logger.warning(
```
LJBabbage: 400 can be raised by invalid by a invalid token as well, so this probably should be payload

frontstage/views/passwords/reset_password.py
```diff
@@ -58,24 +57,26 @@ def post_reset_password(token):
...
+                "Invalid password sent to party service", api_url=exc.url, api_status_code=exc.status_code, token=token
```
LJBabbage: I don't really understand why we would log the url,  is there value doing it? And why we wouldn't log the email, as it will be in scope at this stage

LJBabbage (changes requested):
The flow needs some work, we should validate the form then have a try, except for the decode, before having one for the party. If party can't perform the password change we are getting back a 500 which we then just raise and dead end the respondent

frontstage/views/passwords/reset_password.py
```diff
@@ -58,24 +57,26 @@ def post_reset_password(token):
...
+            )
+        elif exc.status_code == 409:
```
LJBabbage: How can the token expire on party but not frontstage, however 409 is a conflict, so the warning is wrong anyway

frontstage/views/passwords/reset_password.py
arroyoAle: Cleaned this up, its now catching the errors that decode can throw and showing the right pages

frontstage/views/passwords/reset_password.py
arroyoAle: I'm pretty sure that only an invalid token sent to party will return a 400, so I'll leave it with this log line

frontstage/views/passwords/reset_password.py
arroyoAle: Swapped the url with the obfuscated email

frontstage/views/passwords/reset_password.py
LJBabbage: The error message says password

frontstage/views/passwords/reset_password.py
```diff
@@ -58,24 +58,43 @@ def post_reset_password(token):
     if not form.validate():
         return get_reset_password(token, form_errors=form.errors)
 
-    password = request.form.get("password")
+    password = request.form["password"]
 
     try:
         duration = app.config["EMAIL_TOKEN_EXPIRY"]
```
LJBabbage: Doesn't need to be in the try

LJBabbage (approved):
Coverage is lacking for exceptions which we should always try to do, but it works as expected

matthew-robinson-ons (approved):
re-review after changes

PR #1126, opened 2026-08-25
Initial changes to code

SteveScorfield:
# What and why?

# How to test?

# Jira

### DIFF
--- a/frontstage/exceptions/exceptions.py
+++ b/frontstage/exceptions/exceptions.py
@@ -88,6 +88,14 @@ class InvalidEqPayLoad(Exception):
         self.message = message
 
 
+class InvalidDetails(Exception):
+    def __init__(self, party_id, case_id, survey_id):
+        super().__init__()
+        self.party_id = party_id
+        self.case_id = case_id
+        self.suvey_id = survey_id
+
+
 class UserDoesNotExist(Exception):
     def __init__(self, message):
         super().__init__()
--- a/frontstage/views/surveys/upload_survey.py
+++ b/frontstage/views/surveys/upload_survey.py
@@ -11,7 +11,7 @@ from frontstage.controllers import (
     party_controller,
     survey_controller,
 )
-from frontstage.exceptions.exceptions import CiUploadError, NoSurveyPermission, InvalidDetails
+from frontstage.exceptions.exceptions import CiUploadError, InvalidDetails, NoSurveyPermission
 from frontstage.views.surveys import surveys_bp
 from frontstage.views.template_helper import render_template
 
--- a/frontstage/views/surveys/upload_survey.py
+++ b/frontstage/views/surveys/upload_survey.py
@@ -11,7 +11,7 @@ from frontstage.controllers import (
     party_controller,
     survey_controller,
 )
-from frontstage.exceptions.exceptions import CiUploadError, NoSurveyPermission
+from frontstage.exceptions.exceptions import CiUploadError, NoSurveyPermission, InvalidDetails
 from frontstage.views.surveys import surveys_bp
 from frontstage.views.template_helper import render_template
 
@@ -40,14 +40,13 @@ def upload_survey(session):
         )
         abort(400)
 
-    # Check if respondent has permission to upload for this case
-    survey = survey_controller.get_survey_by_short_name(survey_short_name)
-    if not party_controller.is_respondent_enrolled(party_id, business_party_id, survey["id"]):
-        raise NoSurveyPermission(party_id, case_id)
-
     case = case_controller.get_case_by_case_id(case_id)
     case_group = case.get("caseGroup")
     collection_exercise_id = case_group.get("collectionExerciseId")
+    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+    survey_id = survey["id"]
+    held_survey_id = case_group.get("surveyId")
+
     business_party = party_controller.get_party_by_business_id(
         case_group["partyId"],
         app.config["PARTY_URL"],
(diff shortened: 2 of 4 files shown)


anwilkie (approved):

frontstage/exceptions/exceptions.py
```diff
@@ -88,6 +88,14 @@ def __init__(self, message):
         self.message = message
 
 
+class InvalidDetails(Exception):
```
LJBabbage: Not sure we need a new exception for just this, it's just a bad request. If for some reason we do need it, it realistically shouldn't be a generic name for a particular set or variables and exception

frontstage/views/surveys/upload_survey.py
```diff
@@ -40,14 +44,13 @@ def upload_survey(session):
...
-    survey = survey_controller.get_survey_by_short_name(survey_short_name)
-    if not party_controller.is_respondent_enrolled(party_id, business_party_id, survey["id"]):
-        raise NoSurveyPermission(party_id, case_id)
-
     case = case_controller.get_case_by_case_id(case_id)
     case_group = case.get("caseGroup")
     collection_exercise_id = case_group.get("collectionExerciseId")
+    survey = survey_controller.get_survey_by_short_name(survey_short_name)
+    survey_id = survey["id"]
+    held_survey_id = case_group.get("surveyId")
```
LJBabbage: Held is a slightly unusual name, and losses context lower down, this is the case_group_survey_id

frontstage/views/surveys/upload_survey.py
```diff
@@ -40,14 +44,13 @@ def upload_survey(session):
...
+
     business_party = party_controller.get_party_by_business_id(
```
LJBabbage: This is very odd logic, use the business id to get the business details and then use the id on line 62 which will be the same.  business_party is used later on, but the call isn't needed at this point, I also question the usage then, I see only the id and the checkletter being used, and with the  checkletter not being in party and only getting pulled in because it is passing a verbose param and going elsewhere is strange

frontstage/views/surveys/upload_survey.py
```diff
@@ -56,6 +59,16 @@ def upload_survey(session):
         verbose=True,
     )
 
+    held_business_party_id = business_party["id"]
+
+    if business_party_id != held_business_party_id or survey_id != held_survey_id:
+        logger.error("Party ID and/or Survey ID do not match", party_id=party_id, survey_id=survey_id)
```
LJBabbage: Programmatically it will never be **and** as python won't evaluate the 2nd condition if the first is false. It probably needs to also be explicit, as if I look in the log I wouldn't know what this meant without finding the error and then determining that the provide combination doesn't match the combination in the case

tests/integration/views/surveys/test_upload_survey.py
```diff
@@ -29,7 +31,9 @@ def setUp(self):
         self.app = app.test_client()
         self.app.set_cookie("authorization", "session_key")
         self.headers = {
-            "Authorization": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicmluZ3JhbUBub3d3aGVyZS5jb20iLCJ1c2VyX3Njb3BlcyI6WyJjaS5yZWFkIiwiY2kud3JpdGUiXX0.se0BJtNksVtk14aqjp7SvnXzRbEKoqXb8Q5U9VVdy54"  # NOQA
+            "Authorization": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoicmluZ3JhbUBub3d3aGVyZS5jb20iLCJ1c2Vy"
```
LJBabbage: Is this just linting?

tests/test_data/case/case_diff_businessId.json
```diff
@@ -0,0 +1,24 @@
+{
```
LJBabbage: We really should try avoid doing this, frontstage is terrible for lots of very specific json files which just adds noise, either give the test just elements it needs, or re-use an existing high level file and in-line change

tests/test_data/party/business_party_mismatch.json
```diff
@@ -0,0 +1,32 @@
+{
```
LJBabbage: Same as above

LJBabbage (changes requested):
Needs a little more work, the logic is a bit muddled (before you started) and the view is overloaded, some of it should be in a controller. Testing is missing coverage on 65,66 (new code) and 70(old) and should not be done via a client. I'm also not a fan of the extra JSON files to use one var

frontstage/views/surveys/upload_survey.py
SteveScorfield: I've removed the need for this.

frontstage/exceptions/exceptions.py
SteveScorfield: I removed this. I was working out what was the best way to handle this. I'll revert back to a log.

frontstage/views/surveys/upload_survey.py
SteveScorfield: During testing, when using a different business party ID to simulate a malicious intent, the business_party object retrieved a different business party ID to the one supplied. Which is what would be expected when the search is using the party_id (saved as part of thesession) which would be the source of truth for this.

frontstage/views/surveys/upload_survey.py
SteveScorfield: There is a possibility that the business_party_id supplied is correct but the survey_id is a malicious change. So either is a possibility.

I've updated the logging to be more informative.

tests/integration/views/surveys/test_upload_survey.py
SteveScorfield: Yep. Lint was complaining about it.

LJBabbage (approved):

